PR #45, opened 2013-07-07
Add an append option and refactor methods to use ApiResource functions

jpiasetz:
Refactored a few methods to use ApiResource. There are a few more place where something like this can be done.


bkrausz:
Given that this has been sitting for years, I'm going to close it out. Feel free to open a new PR if still applicable.

PR #117, opened 2014-12-18
Add support for a per-request options hash in the place of an API key

kyleconroy:
For example, when creating a customer, you can specify an idempotency
key:

```
Stripe_Customer::create(array(
  'description' => 'Some customer',
), array(
  'idempotency_key' => 'foo',
));
```
This allows us to add more per-request attributes in the future.


kyleconroy:
@bkrausz how does this look now?

bkrausz:
lgtm, :sheep:. We'll need to merge this into experimental-v2, but there should be few enough of these that I'm fine doing it all at once another week (unless you'd like to volunteer).

kyleconroy:
Attempting to merge into experimental-v2, will see what happens

PR #119, opened 2015-01-06
Added Bitcoin Receiver & Transaction objects

maxwells:


maxwells:
hey Brian -- just as a heads up, I've updated this PR by removing BitcoinTransaction as a top level API object

lib/Stripe/BitcoinTransaction.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+class Stripe_BitcoinTransaction extends Stripe_ApiResource
```
bkrausz: Is the intent here that bitcoin transactions will properly have their own object, but you won't be able to do anything special with them like save?

maxwells: Yes. They're only createable through Customer action (eg. transferring funds to the receiver). I'm certainly open to having the BitcoinTransaction sublist on BitcoinReceiver just be associative arrays, though

bkrausz: They do seem to be listable and fetchable, so do we want to have `retrieve` and `all` methods?

maxwells: I chatted about it with @raycmorgan and @canderson and we felt it was an unnecessarily granular unit to expose at the bindings level. It represents the set of transactions that fill a bitcoin receiver (customers pushing coins to the receiver) and we couldn't think of much of a valid use case for querying those outside the context of a bitcoin receiver. The proposed changes here does allow them to be pulled as a sublist

```php
$receiver->transactions->all();
```

How do you feel about this approach?

bkrausz: works for me, thanks for clarifying. :sheep:

PR #121, opened 2015-01-09
 Switch to PHPUnit / Remove PHP 5.2 Support

chadicus:
I broke these changes into lots of small commits. 

With the addition of PHPUnit you will be required to have php 5.3.3 or greater for development.

:exclamation:  The final commit in the PR removes PHP 5.2 support.  And ups the minimum supported version of PHP to 5.3.3


bkrausz:
Thanks for doing all this work! I'm taking a look now. It would be great if we could consider the PHP 5.2/5.3 question separately from this. How easy would it be to make PHPUnit work in PHP 5.2? If it's a pain don't bother, we can find another solution, but I wouldn't want this change to be blocked on that discussion unnecessarily.

tests/Stripe/ApiRequestorTest.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-class Stripe_ApiRequestorTest extends UnitTestCase
+class Stripe_ApiRequestorTest extends PHPUnit_Framework_TestCase
```
bkrausz: `Stripe_TestCase`?

tests/Stripe/AuthenticationErrorTest.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-class Stripe_AuthenticationErrorTest extends UnitTestCase
+class Stripe_AuthenticationErrorTest extends PHPUnit_Framework_TestCase
```
bkrausz: `Stripe_TestCase`

tests/Stripe/ObjectTest.php
```diff
@@ -1,12 +1,12 @@
 <?php
 
-class Stripe_ObjectTest extends UnitTestCase
+class Stripe_ObjectTest extends PHPUnit_Framework_TestCase
```
bkrausz: `Stripe_TestCase`

I know the 1:1 translation of all of these would have them left as the base test case, but I see no reason not to inherit everything from `Stripe_TestCase` and give us the consistency there.

tests/Stripe/TokenTest.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+class Stripe_TokenTest extends PHPUnit_Framework_TestCase
```
bkrausz: `Stripe_TestCase`

bkrausz:
Ok, took a look, everything looks great except for a couple of minor nits. I think there are 2 easy paths to shipping this:

1. Make this continue to support 5.2 via similar `pear install` hacks that we currently use, enabling us to decide on the "post 5.2 world" API separately from this change.
2. Move this onto the experimental-v2 branch and simultaneously decide what API we'd like in the v2 cut.

I'm happy to do #2, but it's a question of stabilizing the API as much as possible before we actually cut the change: I'd hate to add in, for example, namespaces, and then a month later also remove support for array access - we should batch as many preferred breaking changes as possible together.

tests/Stripe/ApiRequestorTest.php
chadicus: These tests did not use the original `StripeTestCase` As they did not need any of the additional functionality, i felt it was best to not add that here

chadicus:
@bkrausz The minimum PHP version for PHPUnit is 5.3.3.  The phpunit pear channel was removed last year.

If you want to run tests for version 5.2 you must either not use PHPUnit or find an older version of phpunit and keep it with the repository.

tests/Stripe/ApiRequestorTest.php
chadicus: @bkrausz All tests now extend `Stripe_TestCase`

GrahamCampbell:
This should be squash to one commit too.

composer.json
```diff
@@ -15,13 +15,14 @@
...
-    "php": ">=5.2",
+    "php": ">=5.3.3",
     "ext-curl": "*",
     "ext-json": "*",
     "ext-mbstring": "*"
   },
   "require-dev": {
-    "simpletest/simpletest": "*"
+    "phpunit/phpunit": "~4.0",
+    "squizlabs/php_codesniffer": "~1.5"
```
GrahamCampbell: We should not be running cs checks on travis. Use scrutinizer or something to do that.

chadicus: agreed, but that should not go into this pull request

chadicus:
@GrahamCampbell the separate commits were so that the PR was easier to review, I'm happy to squash the commits if the project owners want me to do so.

chadicus:
Just FYI it looks like to build failed due to connectivity.

chadicus:
@bkrausz I've added a second pull request #125 for the experimental-v2 branch. I'll let the project owners decide which route they want to go.

bkrausz:
I'm fine with this staying as multiple commits.

I'm chatting with the team now on the best way to make sure this gets shipped. I've merged it into experimental-v2 to start, we'll see about getting it into master ASAP.

Thanks again for putting in the work for this!

composer.json
chadicus: @GrahamCampbell  see #128

chadicus:
Probably best to just close this and have it released with version 2

bkrausz:
:+1:

PR #126, opened 2015-01-12
Add namespaces - For discussion ONLY

chadicus:
All classes in Stripe namespace
Use psr-4 autoloading in composer


lib/Account.php
```diff
@@ -1,11 +1,12 @@
 <?php
+namespace Stripe;
```
GrahamCampbell: I prefer the following formatting:

```php
<?php

namespace Stripe;

class ...
```

That style is more popular, for example, symfony use it.

chadicus: fixed.

bkrausz:
:+1: :sheep: :smile:

bkrausz:
Let me know if you wanted to do more work on this, since it's for the experimental branch and seems like an obvious next step I'm happy to have it in :).

chadicus:
Whoa, didn't expect this to get merged. and indeed I have another PR.

PR #127, opened 2015-01-12
Add Support for coveralls code coverage

chadicus:
This adds code coverage reporting using [Coveralls](https://coveralls.io/)

### DIFF
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,0 +1,4 @@
+service_name: travis-ci
+src_dir: .
+coverage_clover: clover.xml
+json_path: coveralls-upload.json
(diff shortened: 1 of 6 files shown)


chadicus:
The account for stripe would need to be set up with coveralls, but it's very simple to do.

bkrausz:
I setup coveralls.io - do I need to do anything else with Travis CI so that coverage gets computed correctly?

chadicus:
@bkrausz the unit tests will need to be updated to use the `@covers` annotations.

Alternatively I think I can remove the `forceCoversAnnotation="true"`  config from the `phpunit.xml` file which should default coverage. I'll add a commit that does that.

chadicus:
@bkrausz I also have a PR for Scrutinizer #128 That account for stripe needs to be created as well if you want to use scrutinizer for phpcs

PR #130, opened 2015-01-12
Use PSR2 standard instead of Zend - Fix all errors in code

chadicus:
Since Zend is even moving away from their own standard, it seems like stripe should do the same.  

The `-n` in the `build.php` is a flag to ignore warnings. This allows the code to keep the underscore prefixes for methods and properties without failing the build

### DIFF
--- a/build.php
+++ b/build.php
@@ -9,7 +9,7 @@ if ($returnStatus !== 0) {
 }
 
 passthru(
-    './vendor/bin/phpcs --standard=Zend lib tests *.php',
+    './vendor/bin/phpcs --standard=PSR2 -n lib tests *.php',
     $returnStatus
 );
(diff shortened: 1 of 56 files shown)


GrahamCampbell:
:+1:

GrahamCampbell:
Zend has been PSR-2 for ages hasn't it?

chadicus:
Yes, since ZF2, they use PSR

bkrausz:
Oh man, I *hate* 4 spaces as opposed to 2. That being said, PSR-2 seems to be the major hotness nowadays, so why not :-P

chadicus:
Awesome!

chadicus:
@bkrausz you can always set you IDE to display the tab as 2 spaces

PR #133, opened 2015-02-06
Updated docblock return type

SteveEdson:
This will provide autocomplete support in popular IDEs such as PHPStorm.

When iterating over the returned plans, each $plan will recognised as a `Stripe_Plan` with the available methods.


bkrausz:
Works for me, thanks :)

Issue #111, opened 2014-11-06
Raise the minimum PHP version to 5.4 so that we can improve the library


alexbilbie:
I just started working with this library today and I was quite shocked at how old the code felt until I noticed that the minimum PHP requirement in the composer.json file is PHP 5.2.

In my opinion version 2.x of this library should be a top down rewrite that requires PHP 5.4 as a minimum so that we can make use of some modern practises like namespaces and DI injection. Version 1.x would still be kicking around so those that really are stuck on PHP 5.2 can still use the library.

Would you up for this? I'd rather not see yet another Stripe library appear on Packagist, I'd prefer to use an officially supported SDK and I'd be happy to contribute some time towards helping improve the library.

shelaghstudio:
having the support stripe plugin in php 5.4 version....
So, please given solution ...
share the file given bottom of my id..

email id: nikhil.volcone@gmail.com

PR #135, opened 2015-02-13
Merge v2 into master

bkrausz:


CHANGELOG.md
```diff
@@ -1,106 +1,116 @@
-=== 1.18.0 2015-01-21
+<<<<<<< HEAD:CHANGELOG.md
```
GrahamCampbell: Ooops...

README.md
```diff
@@ -0,0 +1,71 @@
+# Stripe PHP bindings
+
+[![Build Status](https://travis-ci.org/stripe/stripe-php.svg?branch=master)](https://travis-ci.org/stripe/stripe-php) 
+[![Latest Stable Version](https://poser.pugx.org/stripe/stripe-php/v/stable.svg)](https://packagist.org/packages/stripe/stripe-php)
+[![Total Downloads](https://poser.pugx.org/stripe/stripe-php/downloads.svg)](https://packagist.org/packages/stripe/stripe-php)
+[![License](https://poser.pugx.org/stripe/stripe-php/license.svg)](https://packagist.org/packages/stripe/stripe-php)
+[![Code Coverage](https://coveralls.io/repos/stripe/stripe-php/badge.png?branch=experimental-v2)](https://coveralls.io/r/stripe/stripe-php?branch=experimental-v2)
+
+
+
```
GrahamCampbell: Why so many new lines?

README.md
```diff
@@ -0,0 +1,71 @@
+# Stripe PHP bindings
+
+[![Build Status](https://travis-ci.org/stripe/stripe-php.svg?branch=master)](https://travis-ci.org/stripe/stripe-php) 
+[![Latest Stable Version](https://poser.pugx.org/stripe/stripe-php/v/stable.svg)](https://packagist.org/packages/stripe/stripe-php)
+[![Total Downloads](https://poser.pugx.org/stripe/stripe-php/downloads.svg)](https://packagist.org/packages/stripe/stripe-php)
+[![License](https://poser.pugx.org/stripe/stripe-php/license.svg)](https://packagist.org/packages/stripe/stripe-php)
+[![Code Coverage](https://coveralls.io/repos/stripe/stripe-php/badge.png?branch=experimental-v2)](https://coveralls.io/r/stripe/stripe-php?branch=experimental-v2)
```
GrahamCampbell: Bages need uodating.

VERSION
```diff
@@ -1 +1 @@
-1.18.0
+2.0.0
```
GrahamCampbell: Why do we have this file?

composer.json
```diff
@@ -15,15 +15,17 @@
...
   },
   "require-dev": {
-    "simpletest/simpletest": "*"
+    "phpunit/phpunit": "~4.0",
+    "satooshi/php-coveralls": "~0.6.1",
+    "squizlabs/php_codesniffer": "~2.0"
   },
   "autoload": {
-    "classmap": ["lib/Stripe/"]
+    "psr-4": { "Stripe\\" : "lib" }
```
GrahamCampbell: missing a `/` on the end of `lib`.

lib/Account.php
```diff
@@ -0,0 +1,62 @@
...
+
+namespace Stripe;
+
+class Account extends ApiResource
+{
+    public function instanceUrl()
+    {
+        if ($this['id'] === null) {
+            return '/v1/account';
+        } else {
```
GrahamCampbell: Else block not needed.

bkrausz: Subjective style decision, either way is practically fine (unless PSR-2 has an opinion here).

GrahamCampbell: It's not a cs thing. It's just a micro-optimization.

GrahamCampbell:
For your interest, just run this through my cs checker: https://dev.styleci.io/commits/d2940754c09a7f08ff48a0595b63420b5b768ff6.

GrahamCampbell:
Ignore the short array syntax changes.

GrahamCampbell:
Also, I think it would be worth only supporting php 5.4+ since both php 5.2/5.3 are eol, there's no point in supporting 5.3 still.

VERSION
bkrausz: I suspect part of Stripe's automatic hosting of tags on our servers. @ebroder is this no longer in use?

bkrausz:
Almost all of those are bikesheds around things like spacing around `.` and tab alignment. Unless you're referencing a spec we've agreed to follow, I don't think it's worth changing about 10% of the codebase to align with that particular checker's rules.

Re: 5.3, this was discussed at #111 (and in a number of other issues). Given that packagist is still seeing about 15% usage from 5.3 (which I believe lines up with the 5.3 merchant count on Stripe itself), we're going to support 5.3 for the time being. If, in a few months, the number of 5.3s on the v2.* version of the bindings is quite small, we will know that those folks are unlikely to upgrade anyway, and cut a v3.0 that drops 5.3 support.

GrahamCampbell:
> align with that particular checker's rules.

StyleCI is actually configurable with a .php_cs file. It's just set for psr-2 plus a few extra rules that make sense imo. Many are from symfony's coding style guide.

GrahamCampbell:
I just put that there for interest. That's why I haven't send a pull to change anything.

VERSION
ebroder: Yeah, our tooling for code.stripe.com wants to have a consistent cross-language way of detecting versions, and this is what we use for that.

PR #136, opened 2015-02-13
Added a branch alias

GrahamCampbell:


bkrausz:
Can you elaborate on what the branch alias is for? I assumed all dev would happen on branches and then be merged into master fully function.

GrahamCampbell:
This means people can ask for `~2.0@dev` or `2.0.*@dev` and get the latest stuff that will be a 2.0.x release at some point. People should **never** ask for `dev-master`.

bkrausz:
I see, makes sense :)

OOC, if we cut a 3.0, should we alias the latest 2.0 version to 2.0-dev, or just remove the 2.0 alias entirely and it will auto-pick-up the tagged version when resolving versions?

GrahamCampbell:
Packagist doesn't care what aliases are in commits or tags. It'll just read what ever's currently in the master, so if you ever forget to update the alias before a release, you're not breaking anything.

GrahamCampbell:
> OOC, if we cut a 3.0, should we alias the latest 2.0

You can actually create a `2.0` branch, and just delete the alias, and mark the master as `3.0`.

GrahamCampbell:
I'd recommend creating a `1.18` branch too actually.

bkrausz:
As opposed to the current tagging? I'd much prefer to keep with tagging since it aligns with what every other package repo supports (and packagist seems to support it as well, unless I'm missing something).

GrahamCampbell:
> As opposed to the current tagging?

No, not at all. In addition to tagging. It means people can get the latest upcoming changes before they get tagged. You'd never want to replace tagging with this. :)

GrahamCampbell:
The `1.18` branch would be treated as `1.18.x-dev` under the hood.

PR #140, opened 2015-02-19
`Stripe\Util\Util::convertStripeObjectToArray` not recursively converting Stripe objects

jimrubenstein:
The `\Stripe\Util\Util` class needs to import the `\Stripe\Object` class, as it can not be resolved "naturally" because it's not in the same namespace. (Referenced on [lib/Util/Util:42](https://github.com/stripe/stripe-php/blob/3a7cba2261f97f70b2ca429b62fb1426480851b3/lib/Util/Util.php#L42))

This causes `Util::convertStripeObjectToArray` to fail on recursive objects, which in turn causes `Object::__toArray`, `Object::__toJSON`, and `Object::__toString` all to fail.

To fix the issue, I imported the `\Stripe\Object` into the `\Stripe\Util\Util` class.

### DIFF
--- a/lib/Util/Util.php
+++ b/lib/Util/Util.php
@@ -2,6 +2,8 @@
 
 namespace Stripe\Util;
 
+use Stripe\Object;
+
 abstract class Util
 {
     /**
--- a/tests/ObjectTest.php
+++ b/tests/ObjectTest.php
@@ -40,4 +40,33 @@ class ObjectTest extends TestCase
         $s->foo = 'a';
         $this->assertSame($s->keys(), array('foo'));
     }
+
+    public function testToArray()
+    {
+        $s = new Object();
+        $s->foo = 'a';
+
+        $converted = $s->__toArray();
+
+        $this->assertInternalType('array', $converted);
+        $this->assertArrayHasKey('foo', $converted);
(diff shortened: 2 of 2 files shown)


jimrubenstein:
FWIW this build is failing due to a test in the [CustomerTest.php:159](https://github.com/stripe/stripe-php/blob/3a7cba2261f97f70b2ca429b62fb1426480851b3/tests/CustomerTest.php#L159) file. Which is unreleated to this pull request.

lib/Util/Util.php
```diff
@@ -2,6 +2,8 @@
 
 namespace Stripe\Util;
 
+use \Stripe\Object;
```
GrahamCampbell: No leading slash should be used.

jimrubenstein: I'll drop that out and push back in.

bkrausz:
Hmm...would it be better to just use the absolute reference to \Stripe\Object, since it's only used in one place anyway?

jimrubenstein:
I generally prefer to import at the top of the file, just so all the references are in one place rather than scattered throughout the file. Plus it allows you to alias (if you want/need) without modifying the implementation code below.

bkrausz:
WFM, thanks for the PR!

Issue #138, opened 2015-02-19
Can't/Won't Use Composer


marcusds:
Some of us can't or won't use Composer. Can you not include instructions for not using Composer in the readme as well?

Dylan3304:
Yes, defiantly. I will be stuck on Version 1.8.* till we can use this without composer. Just include the following in your API code on pre-2.
   ` require('path/to/stripe-php/lib/Stripe.php');`

bkrausz:
`Stripe.php` does not load the rest of the files right now. When I get a chance I'll add a file that will load all other classes, and add instructions for using it (happy to take a PR for that as well).

Dylan3304:
Sounds nice :)

marcusds:
Created a PR #141, tried to follow your code styling (as much as I hate starting braces on new lines ;))

bkrausz:
Done, there's an `init.php` file in the root dir that will pre-include everything.

KristenTonnelle:
Hi ! I unzipped the PHP lib (can't use Composer) and got this line in my code :
require_once("www.parkallann.bzh/stripped/stripe-php/init.php");

The server answer with a 500 error ; the console says : 
Failed to load resource: the server responded with a status of 500 (Internal Server Error)

Any idea ? Thanks for reading anyway
[stripe.tar.gz](https://github.com/stripe/stripe-php/files/3000816/stripe.tar.gz)

its-tas:
You can’t use the FQDN(fully qualified domain name) for the require - it
would likely be require_once “/stripped/stripe-php/init.php”;

However you may want to read up on “relative pathing” and “absolute
pathing” and figure out where the lib actually lives in relation to the
file that’s including it, and how to do so properly.

kikloo:
Very bogus gateway, its so difficult to add stripe to an already running website. DON'T USE STRIPE!!

PR #141, opened 2015-02-20
Add an autoloader

marcusds:
As per ticket #138

### DIFF
--- a/autoload.php
+++ b/autoload.php
@@ -3,7 +3,8 @@
 /**
  * @param string $class_name
  */
-function stripe_api_autoload($class_name) {
+function stripe_api_autoload($class_name)
+{
 	$name_prefix = 'Stripe\\'; // The namespace prefix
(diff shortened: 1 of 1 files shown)


bkrausz:
This does not handle sub-namespaces like Util. Also, would be great to add a test for this (likely requires changes to test infra and/or an additional travis test run).

bkrausz:
Added a manual list of files instead, with a test that checks that it's up to date (though it does double the number of tests to run).

hardlynoticable:
Is this an autoloader that users can use instead of using Composer?  

Could you explain why it's needed, what it does, and how to use it?

remistr:
If you don't want to use Composer you don't have to anymore. You can simply load `init.php` which will load all the required files for you. This is covered in the [Manual Installation](https://github.com/stripe/stripe-php#manual-installation) documentation

PR #142, opened 2015-02-20
Note the Composer requirement

matthewarkin:
The Readme makes it seem like composer is optional not required. Lets note the requirement until its removed.


bkrausz:
+1 to the clarity, will remove when we make it optional again. Thanks!

PR #151, opened 2015-03-05
Added support for Alipay

bitmash:
Contains the same methods as the Card resource.


bkrausz:
Thanks for the PR @bitmash (and sorry for the delay). This endpoint is actually only for the Alipay beta. When we move out of beta, Alipay will use the `sources` endpoint just like Bitcoin and cards.

PR #152, opened 2015-03-05
Fix Charge::refund syntax error

mattsnowboard:
EMERGENCY: Fatal Error: Call to undefined method Stripe\Charge::request()

### DIFF
--- a/lib/Charge.php
+++ b/lib/Charge.php
@@ -56,7 +56,7 @@ class Charge extends ApiResource
     public function refund($params = null, $options = null)
     {
         $url = $this->instanceUrl() . '/refund';
-        list($response, $opts) = $this->request('post', $url, $params, $options);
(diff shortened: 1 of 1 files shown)


mattsnowboard:
I wanted to add a unit test for this, but I wasn't sure how to do it. Anyway, I checked the commit that caused this and this seemed to be the only "$this->request()" that should be "$this->_request()"

bkrausz:
You could make a test charge and then refund it. Our PHP tests currently do not stub network requests, which does make these things annoying. Something on my list to fix.

mattsnowboard:
Yeah, I tried that but I also tried to verify the amount. Since there are no network requests as you mentioned, that did not work. I could add a test that just does the refund but didn't seem like it was testing much other than this syntax error. Decoupling the request in a way that is more testable would be really nice.

Anyway, thanks for the quick merge! I needed this fix to keep using the latest version of this library in my application.

PR #159, opened 2015-04-28
Ability to mock requests

bkrausz:
Pulls Curl out into an interface and provides for some basic stubbing.

r? @kyleconroy 
CC @russelldavis


lib/HttpClient/CurlClient.php
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    /**
+     * @param number $errno
+     * @param string $message
+     * @throws Error\ApiConnection
+     */
+    public function handleCurlError($errno, $message)
+    {
+        $apiBase = $this->_apiBase;
```
kyleconroy: I don't believe this is defined for this class.

kyleconroy:
Looks good, but I think you need a few tests around `handleCurlError`

bkrausz:
Unfortunately it's really hard to stub out native methods in PHP, so getting `curl_exec` to return the right thing to trigger handleCurlError isn't easily doable. Fixed the URL reference and manually verified the error returns doesn't itself throw and error (just turned off the wifi on my laptop and ran tests).

Shipping for now and cutting a new version since this is an improvement over the status quo.

PR #160, opened 2015-04-30
Fix Transfer::reverse syntax error

neatness:
EMERGENCY: Fatal Error: Call to undefined method Stripe\Transfer::request()

Hi,

Found this error when trying to process refunds.
Let me know if this fix is OK.

Same fix as was for ffc451ac4700fecd89574647f31ad5b8615a8ef3 which was a Charge::request() fatal error.

Thanks

Ivan


bkrausz:
Sorry about that, thanks for the fix!

bkrausz:
Grepped around and confirmed that this is the last instance of the missing underscore.

PR #165, opened 2015-05-13
Fix path to ca-certificates.crt

### LINKED ISSUE #164: PHP library v2.1.3 has incorrect path to fallback certificates file
Due to a small misconfiguration in my own php.ini file (Ooops, now fixed), the curl code in Stripe's HttpClient code could not find "my" cainfo bundle, so it tried to fall back and use it's own, but failed.  

The problem seems to be that the caBundle function in stripe-php/lib/HttpClient/CurlClient.php contains the line:

return dirname(__FILE__) . '/../data/ca-certificates.crt';
but the "data" directory appears to be one level higher than that.  That is, it's in the stripe-php directory, not the stripe-php/lib directory.  So I think that line should be 

return dirname(__FILE__) . '/../../data/ca-certificates.crt';

Larry

matthewarkin:
Per email to api-discuss and fixes #164

### DIFF
--- a/lib/HttpClient/CurlClient.php
+++ b/lib/HttpClient/CurlClient.php
@@ -127,7 +127,7 @@ class CurlClient implements ClientInterface
 
     private static function caBundle()
     {
-        return dirname(__FILE__) . '/../data/ca-certificates.crt';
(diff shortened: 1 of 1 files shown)


bkrausz:
Gah! Nice catch!

PR #169, opened 2015-06-04
Change order of arguments when throwing exceptions with JSON information

pittyplatsch:
I believe there has been a mix up in the assignment.

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -78,7 +78,7 @@ class ApiRequestor
         if (!is_array($resp) || !isset($resp['error'])) {
             $msg = "Invalid response object from API: $rbody "
               . "(HTTP response code was $rcode)";
-            throw new Error\Api($msg, $rcode, $rbody, $resp);
+            throw new Error\Api($msg, $rcode, $resp, $rbody);
         }
 
         $error = $resp['error'];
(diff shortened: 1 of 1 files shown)


pittyplatsch:
I did not run the test suite. Guess it needs some changes too unless I'm mistaken with this whole issue.

bkrausz:
The order on all of these exceptions is httpBody first and jsonBody second, so the original order seems correct. What are you seeing that makes you think otherwise?

pittyplatsch:
I get JSON when calling getHttpBody on an exception and and array when calling getJsonBody. So to me this seems mixed up...

bkrausz:
`getHttpBody` returns the raw response that the server sent back, which should always be a JSON string anyway. `getJsonBody` returns the parsed JSON response. You almost always want `getJsonBody`, but the other method is there in case, for example, a string that doesn't parse as JSON is returned (again, this should never happen).

Closing out since it sounds like this is working as expected, but feel free to reopen if I'm misunderstanding something.

pittyplatsch:
Thanks for clarifying. However I find it odd that getJsonBody does not deliver JSON while the other method does...

bkrausz:
`getJsonBody` delivers parsed Json, as in what you get back when you run `json_decode` on `getHttpBody`. I agree on the ambiguity there though: it would be better off being named `getParsedBody` or similar.

However, given it's current use I think the better solution is to document the method than rename it.

PR #170, opened 2015-06-16
Add correct return type for BalanceTransaction objects

pittyplatsch:


bkrausz:
Thanks for the fix! Build failure looks transient, lgtm.

PR #174, opened 2015-07-02
Add request IDs to all errors

kyleconroy:
```php
$e = new Stripe\Error\Api();
$e->getHttpHeaders();
$e->getRequestId();
```

### DIFF
--- a/lib/Error/Base.php
+++ b/lib/Error/Base.php
@@ -18,10 +18,10 @@ abstract class Base extends Exception
         $this->httpBody = $httpBody;
         $this->jsonBody = $jsonBody;
         $this->httpHeaders = $httpHeaders;
-        $this->requestID = null;
+        $this->requestId = null;
 
         if ($httpHeaders && isset($httpHeaders['Request-Id'])) {
-          $this->requestId = $httpHeaders['Request-Id'];
+            $this->requestId = $httpHeaders['Request-Id'];
         }
     }
 
@@ -50,4 +50,11 @@ abstract class Base extends Exception
         return $this->requestId;
     }
 
+    public function __toString()
+    {
+        $id = $this->requestId ? " from API request '{$this->requestId}'": "";
+        $message = explode("\n", parent::__toString());
+        $message[0] .= $id;
+        return implode("\n", $message);
+    }
 }
--- a/lib/Error/Base.php
+++ b/lib/Error/Base.php
@@ -10,12 +10,19 @@ abstract class Base extends Exception
         $message,
         $httpStatus = null,
         $httpBody = null,
-        $jsonBody = null
+        $jsonBody = null,
+        $httpHeaders = null
     ) {
         parent::__construct($message);
         $this->httpStatus = $httpStatus;
         $this->httpBody = $httpBody;
         $this->jsonBody = $jsonBody;
+        $this->httpHeaders = $httpHeaders;
+        $this->requestID = null;
+
+        if ($httpHeaders && isset($httpHeaders['Request-Id'])) {
+          $this->requestId = $httpHeaders['Request-Id'];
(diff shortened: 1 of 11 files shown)


lib/Error/Base.php
```diff
@@ -10,12 +10,19 @@ public function __construct(
...
-        $jsonBody = null
+        $jsonBody = null,
+        $httpHeaders = null
     ) {
         parent::__construct($message);
         $this->httpStatus = $httpStatus;
         $this->httpBody = $httpBody;
         $this->jsonBody = $jsonBody;
+        $this->httpHeaders = $httpHeaders;
+        $this->requestID = null;
```
bkrausz: Lowercase `d`

lib/HttpClient/CurlClient.php
```diff
@@ -47,12 +47,21 @@ public function request($method, $absUrl, $headers, $params, $hasFile)
...
+          return strlen($header_line);
+        };
+
         $absUrl = Util\Util::utf8($absUrl);
         $opts[CURLOPT_URL] = $absUrl;
         $opts[CURLOPT_RETURNTRANSFER] = true;
         $opts[CURLOPT_CONNECTTIMEOUT] = 30;
         $opts[CURLOPT_TIMEOUT] = 80;
         $opts[CURLOPT_RETURNTRANSFER] = true;
+        $opts[CURLOPT_HEADERFUNCTION] = $headerCallback;
```
bkrausz: Hmmm...emotionally caught between this option and http://stackoverflow.com/a/9183276/2515035 (split on `\r\n\r\n`). Any reason why you chose this option?

tests/CardErrorTest.php
```diff
@@ -24,6 +30,7 @@ public function testDecline()
             Charge::create($charge);
         } catch (Error\Card $e) {
             $this->assertSame(402, $e->getHttpStatus());
+            $this->assertStartsWith("req_", $e->getRequestId());
```
bkrausz: If this just going to return a boolean, or does PHPUnit magically know that `assert*` methods should raise on false?

lib/HttpClient/CurlClient.php
kyleconroy: I originally split on `\r\n\r\n`, but ran into some strange errors with FileUploads. I think this approach works better because I don't need to munge the response at all. Also, from further down in the SO post: `it won't be affected by additional headers from proxies` which seems like a win as well.

tests/CardErrorTest.php
kyleconroy: Yeah, that's just wrong. Fixed.

lib/Error/Base.php
kyleconroy: Whoops, fixed.

tests/CardErrorTest.php
```diff
@@ -4,6 +4,12 @@
 
 class CardErrorTest extends TestCase
 {
+    public function assertStartsWith($haystack, $needle)
+    {
+        // search backwards starting from haystack length characters from the end
+        $this->assertTrue($needle === "" || strrpos($haystack, $needle, -strlen($haystack)) !== false);
```
bkrausz: Why not `strpos($haystack, $needle) === 0`?

lib/Error/Base.php
```diff
@@ -32,4 +39,21 @@ public function getJsonBody()
...
+    {
+        return $this->httpHeaders;
+    }
+
+    public function getRequestId()
+    {
+        return $this->requestId;
+    }
+
+    public function __toString() {
```
bkrausz: Test failures are for opening brace on newline

bkrausz:
lgtm barring the 2 nits.

PR #177, opened 2015-07-18
Add top level dispute objects

kyleconroy:


lib/Charge.php
```diff
@@ -79,6 +79,8 @@ public function capture($params = null, $options = null)
      * @param array|null $params
      * @param array|string|null $options
      *
+     * @deprecated Use the `save` method on the Dispute object
+     *
```
bkrausz: Does `$this->dispute` load an actual `Dispute` now? Do we want to just glue things through as one would expect: update the `Dispute` with all of the `$params` and call `save` on it?

lib/Charge.php
```diff
@@ -92,6 +94,8 @@ public function updateDispute($params = null, $options = null)
     /**
      * @param array|string|null $options
      *
+     * @deprecated Use the `close` method on the Dispute object
+     *
```
bkrausz: Ditto

tests/DisputeTest.php
```diff
@@ -0,0 +1,91 @@
...
+        );
+        $c = Charge::retrieve($c->id);
+
+        $attempts = 0;
+
+        while ($c->dispute === null) {
+            if ($attempts > 5) {
+                throw "Charge is taking too long to be disputed";
+            }
+            sleep(10);
```
bkrausz: Given that a consumer is unlikely to be between 10 and 50 seconds behind (i.e. if it's 10s behind it's probably a few minutes behind), do we want to make this 1s?

Alternatively, we now have the infra to stub out responses in PHPUnit, so we could just stub out a dispute.

bkrausz:
See comments, but generally lgtm.

If you don't have any urgency to cut a new release, I have 2 minor breaking fixes to make for a 3.0 release, happy to roll this into it.

kyleconroy:
Yeah, no urgency here. We shouldn't merge this in until disputes have shipped to everyone.

lib/Charge.php
kyleconroy: I'd rather note, as that may have unexpected changes. I think tying the deprecated methods to the deprecated endpoints makes sense.

tests/DisputeTest.php
kyleconroy: I'd rather not stub things out, so I'll change the sleep to 1

PR #179, opened 2015-07-24
Fixed null reference return

ThisIsAreku:
This PR fix a E_NOTICE 'Only variable references should be returned by reference' when trying to access an inexistent property using the object notation (```$s->myprop```)
It also add unit tests to cover this case

### DIFF
--- a/lib/Object.php
+++ b/lib/Object.php
@@ -92,6 +92,8 @@ class Object implements ArrayAccess
     }
     public function &__get($k)
     {
+        // function should return a reference, using $nullval to return a reference to null
+        $nullval = null;
         if (array_key_exists($k, $this->_values)) {
             return $this->_values[$k];
         } else if ($this->_transientValues->includes($k)) {
@@ -104,11 +106,11 @@ class Object implements ArrayAccess
                     . "probably as a result of a save(). The attributes currently "
                     . "available on this object are: $attrs";
             error_log($message);
-            return null;
+            return $nullval;
         } else {
             $class = get_class($this);
             error_log("Stripe Notice: Undefined property of $class instance: $k");
-            return null;
+            return $nullval;
         }
     }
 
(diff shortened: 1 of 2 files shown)


lib/Object.php
```diff
@@ -104,11 +105,11 @@ public function &__get($k)
                     . "probably as a result of a save(). The attributes currently "
                     . "available on this object are: $attrs";
             error_log($message);
-            return null;
+            return $nullval;
         } else {
             $class = get_class($this);
             error_log("Stripe Notice: Undefined property of $class instance: $k");
-            return null;
+            return $nullval;
```
GrahamCampbell: Should just be `return;` with no value.

GrahamCampbell: Your tests will still pass with that change.

blaugueux:
:+1:

lib/Object.php
ThisIsAreku: On my side they don't pass, using ```return;``` is strictly equals to ```return null;``` and constants can't be passed by reference.

GrahamCampbell: Hmmm. PHP's a bit rubbish sometimes.

tests/ObjectTest.php
```diff
@@ -69,4 +69,16 @@ public function testRecursiveToArray()
         $this->assertArrayHasKey('foo', $converted['child']);
         $this->assertEquals('a', $converted['child']['foo']);
     }
+
+    public function testInexistentProperty()
```
bkrausz: Sorry for the super nit, but can we call this `testNonexistentProperty` or `testPropertyDoesNotExists`. `Inexistent` is valid english, but is fairly rarely used.

lib/Object.php
```diff
@@ -92,6 +92,7 @@ public function __unset($k)
     }
     public function &__get($k)
     {
+        $nullval = null;
```
bkrausz: Since this looks kind of weird, mind adding a comment as to why it's necessary?

bkrausz:
See my nits, otherwise lgtm

PR #181, opened 2015-07-28
Rename Object to StripeObject

### LINKED ISSUE #172: Stripe\Object class name is incompatible with PHP 7
While I don't use PHP 7 yet, its quickly coming down the release pipeline and many will want to switch to it for its massive performance improvement and exhaustive list of new features.

Unfortunately, however, the `object` (case-insensitive) name [will be reserved in PHP 7][php-7-rfc-reserve-types] (to allow for type-hinting on generic objects in the future) and so the `Stripe\Object` class will be technically incompatible with PHP 7.

I'd suggest renaming it to something a bit more package specific. Maybe `Stripe\Entity` or something.

[php-7-rfc-reserve-types]: https://wiki.php.net/rfc/reserve_even_more_types_in_php_7

bkrausz:
Fixes #172 

See Issue for context

r? @kyleconroy

### DIFF
--- a/init.php
+++ b/init.php
@@ -22,7 +22,7 @@ require(dirname(__FILE__) . '/lib/Error/InvalidRequest.php');
 require(dirname(__FILE__) . '/lib/Error/RateLimit.php');
 
 // Plumbing
-require(dirname(__FILE__) . '/lib/Object.php');
+require(dirname(__FILE__) . '/lib/StripeObject.php');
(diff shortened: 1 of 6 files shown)


kyleconroy:
LGTM

This is a breaking change though, so we should probably bump a version.

bkrausz:
Yup, 3.0 inbound :)

PR #182, opened 2015-07-28
Rename Error code/param to stripeCode/stripeParam

### LINKED ISSUE #178: Stripe\Error\Card::getCode() is incompatible with PHP 5.6
`Stripe\Error\Card` sets the Exception's `code` field to a string. In PHP 5.6, this field is cast to an integer after the exception is thrown (the required return type from `Exception::getCode()`). The result is that when running on PHP 5.6, accessing this via getCode() always returns 0.

For example, the following code snippet mimics how `Stripe\Error\Card` sets the `code` field.
```php
<?php

class CodeException extends Exception
{
    public function __construct($message, $code) {
        parent::__construct($message);
        $this->code = $code;
    }
}

$exception = new CodeException("message", "123string that will be cast to int\n");
print $exception->getCode();
try {
    throw $exception;
} catch (Exception $e) {
}
print $exception->getCode();
```

Saving this to a file named "test.php", in PHP 5.5, this code will output:
```
$ php -v
PHP 5.5.25 (cli) (built: May 15 2015 09:30:28) 
Copyright (c) 1997-2015 The PHP Group
php test.php 
123string that will be cast to int
123string that will be cast to int
```

In PHP 5.6, however, the exception code is cast to an integer after being thrown:
```
$ php -v
PHP 5.6.11-1+deb.sury.org~trusty+1 (cli) 
Copyright (c) 1997-2015 The PHP Group
$ php test.php
123string that will be cast to int
123
```

In order to maintain access to the string value in 5.6, the `code` parameter has to be saved to a different field and a new accessor provided.

bkrausz:
Fixes #178

r? @kyleconroy

### DIFF
--- a/lib/Error/Card.php
+++ b/lib/Error/Card.php
@@ -6,15 +6,25 @@ class Card extends Base
 {
     public function __construct(
         $message,
-        $param,
-        $code,
+        $stripeParam,
+        $stripeCode,
         $httpStatus,
         $httpBody,
         $jsonBody,
         $httpHeaders = null
     ) {
         parent::__construct($message, $httpStatus, $httpBody, $jsonBody, $httpHeaders);
-        $this->param = $param;
-        $this->code = $code;
+        $this->stripeParam = $stripeParam;
+        $this->stripeCode = $stripeCode;
+    }
+
+    public function getStripeCode()
+    {
+        return $this->stripeCode;
+    }
+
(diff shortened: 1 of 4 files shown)


bkrausz:
Post-merge review is fine, I want to start testing 3.0.0 so I can get it released today.

kyleconroy:
:+1:

PR #186, opened 2015-07-29
Remove RandomString, increase PHP unit test entropy

shalecraig:
Apparently we have intermittent tests because we create so many coupons, and these things collide.

Remove low-entropy `randomString` method, replace it with a pre-existing `generateRandomString` method.

R? @bkrausz  (feel free to :arrow_right: without talking to me if :green_heart: )


bkrausz:
:llamayay:

PR #190, opened 2015-08-24
Fixed PHPDoc and CS

xavismeh:


bkrausz:
Thanks for the PR! I have a preference to keep the spaces around the string concat operator, I believe it's more readable (and more analogous to other binary operators). Is there a particular style guide you're following for that? We don't specify anything beyond PSR-2.

A lot of the personal nits I have no strong opinion about (e.g. if/else returns vs. if/end of method returns), if we're going to have an opinion it should be one enforced by our phpcs. Otherwise there's no real point in changing them.

xavismeh:
Hello @bkrausz,
All fixes are the result of an analyze using https://github.com/FriendsOfPhp/PHP-CS-Fixer default values (PSR2 + couple of others).

For `if/else return` subject, I personally prefer to avoid "obvious" statements, lowering files' number lines of code and allowing to make it easier to read/debug. My 2¢

kyleconroy:
@xavismeh Could you rebase and re-run your fixer? I agree with @bkrausz that we should only be looking for fixes from PSR-2 and our phpcs.

xavismeh:
Hello @kyleconroy. All changed have been rerun and now take into consideration the new commits made between my PR and the last code modifications. Let me know if you need anything else.

kyleconroy:
@xavismeh this is still a very invasive change, for little benefit. I'd be interested in a change that just updated the PHPDoc, instead of changing white space throughout the entire codebase.

xavismeh:
@kyleconroy when you talk about withe spaces, do you mean https://github.com/stripe/stripe-php/pull/190/files#diff-5dadb2515d8061f081193fcd74a72591R5 for instance ?

kyleconroy:
@xavismeh Indeed, I don't think we're looking to accept this if it includes all those whitespace changes. Sorry!

PR #191, opened 2015-08-29
Added `Request-ID` to API calls.

joshluongo:
I needed to get access to the `Request-ID` that is [sent from the API](https://stripe.com/docs/api#request_ids) for a project I was working on.

In order to make this work I have made two small changes.

* The `Request-ID` is stored in the `RequestOptions` as `$requestId`.
* In order to access `$requestId` I have added added basic check to the magic method in StripeObject.

You can access it by getting `_requestid` from any stripe object.

I hope this comes in handy for someone!

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -57,8 +57,10 @@ class ApiRequestor
         }
         list($rbody, $rcode, $rheaders, $myApiKey) =
         $this->_requestRaw($method, $url, $params, $headers);
-        $resp = $this->_interpretResponse($rbody, $rcode, $rheaders);
-        return array($resp, $myApiKey);
+
+        list($respBody, $requestId) = $resp = $this->_interpretResponse($rbody, $rcode, $rheaders);
+
+        return array($respBody, $myApiKey, $requestId);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


bkrausz:
I agree that we should have a way to get response headers, but I don't think storing it on `RequestOptions` is the right way to expose it: they are meant to be sent per-request, and are fairly independent from the response. For example, they get glued through to future requests (ex: calling `retrieve` and then `save` uses the same `RequestOptions` by default, though I believe it does make a new options object).

I think I'd rather expose a `getLastResponseHeaders` accessor on each object, that is overridden on each refresh. Then we wouldn't need the `_` key magic, and it's relatively clear that each refresh will update it. It also covers all headers, so things like `Stripe-Version` would be fetchable as well.

The other option would be a static method on `ApiRequestor` that does the same thing, but static methods are less appealing.

@kyleconroy - any thoughts on this?

joshluongo:
@bkrausz I agree with you, My method is not the right way to achieve this, I needed the functionality asap for my implementation so it met my requirements.

I think having `getLastResponseHeaders` would be a worthwhile change to make.

kyleconroy:
@bkrausz I think we should probably make this more generic by adding a `getResponse` method that returns the actually API response, status code, body and all.

kyleconroy:
See the work done in #206

PR #192, opened 2015-09-09
add 429 rate limit error and test

jxu577:
r? @kyleconroy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,3 @@
-### 3.3.0 2015-09-08
-
-* Add support for 429 RateLimit errors
-
 ### 3.2.0 2015-08-17
 
 * Add refund listing and retrieval without an associated charge
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### 3.3.0 2015-09-08
+
+* Add support for 429 RateLimit errors
+
 ### 3.2.0 2015-08-17
 
 * Add refund listing and retrieval without an associated charge
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -72,6 +72,8 @@ class ApiRequestor
      *    permissions.
      * @throws Error\Card if the error is the error code is 402 (payment
      *    required)
+     * @throws Error\RateLimit if the error is caused by too many requests
+     *    hitting the API.
      * @throws Error\Api otherwise.
      */
     public function handleApiError($rbody, $rcode, $rheaders, $resp)
@@ -89,10 +91,6 @@ class ApiRequestor
 
         switch ($rcode) {
             case 400:
-                if ($code == 'rate_limit') {
-                    throw new Error\RateLimit($msg, $param, $rcode, $rbody, $resp, $rheaders);
-                }
-
                 // intentional fall-through
             case 404:
                 throw new Error\InvalidRequest($msg, $param, $rcode, $rbody, $resp, $rheaders);
@@ -100,6 +98,8 @@ class ApiRequestor
                 throw new Error\Authentication($msg, $rcode, $rbody, $resp, $rheaders);
             case 402:
                 throw new Error\Card($msg, $param, $code, $rcode, $rbody, $resp, $rheaders);
+            case 429:
+                throw new Error\RateLimit($msg, $rcode, $rbody, $resp, $rheaders);
             default:
                 throw new Error\Api($msg, $rcode, $rbody, $resp, $rheaders);
         }
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -91,6 +91,10 @@ class ApiRequestor
 
         switch ($rcode) {
             case 400:
+                if ($code == 'rate_limit') {
+                    throw new Error\RateLimit($msg, $param, $rcode, $rbody, $resp, $rheaders);
+                }
+
                 // intentional fall-through
             case 404:
                 throw new Error\InvalidRequest($msg, $param, $rcode, $rbody, $resp, $rheaders);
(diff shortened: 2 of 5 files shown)


tests/RateLimitErrorTest.php
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Stripe;
+
+class RateLimitErrorTest extends TestCase
+{
+    public function testRateLimit()
+    {
+        try {
+            for ($x = 0; $x <= 100; $x++) {
+                Customer::create();
```
bkrausz: This is going to make our tests take ~forever (specifically adding 30s to each of our 12 builds). Can we do something else? We have the ability to mock HTTP responses now.

bkrausz:
Are we changing the response code behind a version gate? If so, do we need to leave the old 400 interpreter for older versions?

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+### 3.3.0 2015-09-08
+
+* Add support for 429 RateLimit errors
+
 ### 3.2.0 2015-08-17
```
jxu577: is this update necessary?

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+### 3.3.0 2015-09-08
+
+* Add support for 429 RateLimit errors
```
jxu577: Is this update necessary?

tests/RateLimitErrorTest.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace Stripe;
+
+class RateLimitErrorTest extends TestCase
+{
+    public function testRateLimit()
+    {
+        try {
+            throw new Error\RateLimit(
```
bkrausz: This code is essentially testing that the compiler does what's expected. @nelhage gave a [talk about the concept](https://docs.google.com/presentation/d/1QBJ4EntC63FX-R4THZgSFRZNi_R0H7FovN5JfJmgcVA/edit#slide=id.g42d6fffa0_00)

If you want to test this, you should mock out the response so that we'd actually be testing the response parsing. Something similar to https://github.com/stripe/stripe-php/blob/master/tests/TestCase.php#L32-L39

jxu577:
@bkrausz sg, I made some tweaks to mock out the response 
The response code is behind a version gate, so I left the old 400 handling also.

CHANGELOG.md
bkrausz: We usually add this line in a separate PR. Given that this is a 100% backwards compat change I don't think it warrants a version bump, so I'd remove this.

lib/ApiRequestor.php
```diff
@@ -100,6 +102,8 @@ public function handleApiError($rbody, $rcode, $rheaders, $resp)
                 throw new Error\Authentication($msg, $rcode, $rbody, $resp, $rheaders);
             case 402:
                 throw new Error\Card($msg, $param, $code, $rcode, $rbody, $resp, $rheaders);
+            case 429:
```
bkrausz: Can we add a comment on the 400 rate limit that identifies that it's backwards compat for API versions older than X

tests/RateLimitErrorTest.php
```diff
@@ -0,0 +1,23 @@
...
+    private function rateLimitErrorResponse()
+    {
+        return array(
+            'error' => array(),
+        );
+    }
+
+    public function testRateLimit()
+    {
+        try {
```
bkrausz: This will not fail if the exception is not thrown. PHPUnit supports an `expectedException` annotation which will help: https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.exceptions.examples.ExceptionTest.php

kyleconroy:
@bkrausz the change is behind a version gate, so we need to account for both 400s and 429s

tests/RateLimitErrorTest.php
```diff
@@ -0,0 +1,26 @@
...
+    {
+        return array(
+            'error' => array(),
+        );
+    }
+
+    /**
+     * @expectedException Error\RateLimit
+     */
+    public function testRateLimit()
```
kyleconroy: This test isn't correct. I'd split this up into two tests.

```php
**
 * @expectedException Error\RateLimit
 */
public function testRateLimit()
{
    $this->mockRequest('GET', '/v1/accounts/acct_DEF', array(), $this->rateLimitErrorResponse(), 429);
    Account::retrieve('acct_DEF');
}
```
and another than just checks that `Error\RateLimit` returns 429 for `getHttpStatus`.

bkrausz: I don't think the second test is actually necessary, since we pass through whatever HTTP code we get in the response (400 is pre-version-gate, 429 post).

bkrausz: But no strong preference, I just find the HTTP code test to be a little "test that PHP runs properly"-ish (though not blatantly so).

kyleconroy: If we're just passing the code through, I agree that we don't need a second test.

jxu577:
Cool, I added in a comment about the backwards compatibility, added a fail for if the exception is not thrown, and went with not writing the 2nd test.

kyleconroy:
:+1: once tests pass

lib/ApiRequestor.php
```diff
@@ -89,6 +91,8 @@ public function handleApiError($rbody, $rcode, $rheaders, $resp)
 
         switch ($rcode) {
             case 400:
+                // 'rate_limit' code is depreciated, but left here for backwards compatibility
```
mmcgrana: "deprecated"

jxu577: fixed: https://github.com/stripe/stripe-php/commit/10c6eea0834240024b2876f2c57d08162a7c41e0

PR #194, opened 2015-09-10
Adding refund() support to BitcoinReceiver

leonar15:
These changes add SDK support for refunding bitcoin receivers directly. This is needed because sometimes receivers end up with the wrong amount of money or the payment couldn't be completed for some other reason. This also includes a simple test, though I discovered that the BitcoinReceiver object doesn't include any indication of its own refund state (maybe something to add to a future API?).

### DIFF
--- a/lib/BitcoinReceiver.php
+++ b/lib/BitcoinReceiver.php
@@ -63,4 +63,18 @@ class BitcoinReceiver extends ExternalAccount
     {
         return self::_create($params, $opts);
     }
+
+    /**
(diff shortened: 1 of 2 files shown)


bkrausz:
Looks good to me, thanks for the PR!

leonar15:
Thanks for accepting my PR :+1: 

Now to the other point I brought up, is it possible to expose an indicator the refund status on the BR object itself? Possibly a "refunds" array, or at the very least an "amount_refunded" param?

PR #197, opened 2015-09-21
exp_month format corrected

MamunHoque:
here need two digit exp_month. otherwise got exception.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ This legacy version may be included via `require_once("/path/to/stripe-php/lib/S
 
 ```php
 Stripe::setApiKey('d8e8fca2dc0f896fd7cb4cb0031ba249');
-$myCard = array('number' => '4242424242424242', 'exp_month' => 5, 'exp_year' => 2015);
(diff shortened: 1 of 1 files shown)


matthewarkin:
What exception did you get? Prefixing integers with a 0 will cause really weird issues in PHP as PHP will interpret 05 as not base 10 but base 8. Thus 010 is not 10 but 8

MamunHoque:
Ok then make it 12. :)

matthewarkin:
though it doesn't change the fact that 5 and 05 should return the same error, putting 9 should work even though it is single digits, but yes that doc example does use a month in the past

MamunHoque:
ok then write a working example.

matthewarkin:
Sure, I'll submit a pull request, I was more curious into why you say that
"5" as a month gave you an error but not "05"

kyleconroy:
Fixed here https://github.com/stripe/stripe-php/commit/983887d49d9caa96079a57858ace88e6ae3b89bf. @matthewarkin is correct that prefixing with `0` is not the correct thing to do.

PR #176, opened 2015-07-16
Added bank account verification bindings

jackdent:

PR #202, opened 2015-10-08
Add BankAccount::verify method

aaronbski:
Adds BankAccount::verify needed for the ACH beta.

### DIFF
--- a/lib/BankAccount.php
+++ b/lib/BankAccount.php
@@ -4,5 +4,17 @@ namespace Stripe;
 
 class BankAccount extends ExternalAccount
 {
-
+    /**
+     * @param array|null $params
+     * @param array|string|null $options
+     *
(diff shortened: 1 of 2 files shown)


matthewarkin:
Some previous discussion on this #176

bkrausz:
@matthewarkin since the last discussion we've renamed the endpoint so that `.../sources/BA_TOKEN/verify` now works, which this PR uses. 

@aaronbski thanks for the PR! It looks like you're using short array syntax in your test, which PHP 5.3 doesn't support (thus the test failures). If you update this PR to fix that, I'd be very help to merge it in :).

aaronbski:
@bkrausz Fixed. Thanks.

bkrausz:
LGTM, thanks!!

PR #205, opened 2015-10-18
Update installation instructions.

NoUseFreak:
This allows users to install the library in one simple step.

```sh
$ composer require stripe/stripe-php
Using version ^3.4 for stripe/stripe-php
./composer.json has been updated
Loading composer repositories with package information
...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,20 +14,10 @@ PHP 5.3.3 and later.
 
 ## Composer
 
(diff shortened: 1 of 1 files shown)


brandur:
A quick perusal of some other PHP libraries seems to suggest that this is a pretty common method for recommended installation:

* https://github.com/Imangazaliev/DiDOM
* https://github.com/hassankhan/config
* https://github.com/fzaninotto/Faker

So looks good to me. Thanks @NoUseFreak!

PR #206, opened 2015-10-20
Save the last response made on StripeObjects

kyleconroy:
Supersedes #191

### DIFF
--- a/init.php
+++ b/init.php
@@ -22,6 +22,7 @@ require(dirname(__FILE__) . '/lib/Error/InvalidRequest.php');
 require(dirname(__FILE__) . '/lib/Error/RateLimit.php');
 
 // Plumbing
+require(dirname(__FILE__) . '/lib/ApiResponse.php');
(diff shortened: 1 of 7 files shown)


brandur:
Nice! +1. Maybe one missing piece is that it might be worth adding tests on `ApiResource` to verify the new behavior there.

Another slightly more general question: if getting at the request ID is the main goal here, should we bother exposing the entire response? I guess I could see it either way.

kyleconroy:
@brandur while the impetus for this change was accessing the Request-Id, I think the real request is having access to the response. I didn't want to special case Request-Id and then have someone come back and ask for response status or another header (such as Stripe version).

bkrausz:
Nice, lgtm. Was considering suggesting `getFoo` style accessor methods instead of public vars, but I have no strong preference. Other suggestion is to thread `ApiResponse` further up to `_requestRaw`, but again, no preference.

PR #209, opened 2015-11-16
Correct URL used for refunding application fees

### LINKED ISSUE #208: ApplicationFee::refund doesn't return ApplicationFeeRefund obj
Hi.

According to Stripe docs https://stripe.com/docs/api/php#create_fee_refund ApplicationFee::refund() function should return ApplicationFeerefund object. Instead if this it returns ApplicationFee object like this

public function refund($params = null, $opts = null)
    {
        ...
        return $this;
    }

brandur:
Fixes #208.

See also stripe/stripe-ruby#351.

### DIFF
--- a/lib/ApplicationFee.php
+++ b/lib/ApplicationFee.php
@@ -45,7 +45,7 @@ class ApplicationFee extends ApiResource
      */
     public function refund($params = null, $opts = null)
     {
-        $url = $this->instanceUrl() . '/refund';
+        $url = $this->instanceUrl() . '/refunds';
         list($response, $opts) = $this->_request('post', $url, $params, $opts);
         $this->refreshFrom($response, $opts);
         return $this;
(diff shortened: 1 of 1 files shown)


brandur:
@bkrausz Would you mind taking a look at this approach?

Essentially what we need to do here when `refund()` is called on an application fee is to create a _different_ type of resource (an application fee refund), and I couldn't really find any prior art for this kind of thing on other resources. I thought about trying to reuse `ApiResource`'s `_create` method, but it's pretty tightly keyed to the current class from which it fetches a URL. I could modify `_create` to take an optional URL, but I'm not sure that it's worth it for this one instance. Any thoughts/opinions you have on this would certainly be appreciated. Thanks!

bkrausz:
`->refund()` should be deprecated, instead using `->refunds->create()`. The former doesn't actually return a new object, which is why it's weird. The latter will do what #208 wants.

brandur:
I've also updated this pull to use a similar approach to the one proposed in stripe/stripe-ruby#354.

brandur:
As with stripe/stripe-ruby#354 there's an additional optimization to be had here, but I'm going to pull this in so that we can get onto the new path.

PR #211, opened 2015-11-18
Fix the test suite

brandur:
Unfortunately, stripe-php also seems to run against the live API
service, and as one may expect, the passage of time means that master no
longer passes. This patch updates the broken test with new parameters
that are now expected in the API.


kyleconroy:
:+1: once the tests pass

brandur:
@kyleconroy Thanks!

PR #214, opened 2015-12-17
Allow user-defined override of timeouts

iansltx:
The default connect and response timeouts baked into the Stripe API Curl client are way too long (see https://github.com/stripe/stripe-ruby/issues/46 as wlel), exceeding the standard TTL for an ELB, for example, and are not configurable in any way by the user.

This PR, while a bit of a hack (I dislike static client configuration as little as the next dev), allows the user to override the CurlClient's default connect timeout and total timeout values to something more sane. Getters are also provided, in case you want to switch the timeout for a single call, then set it back to where it was later. Defaults are stored as constants for easier inspection.

### DIFF
--- a/tests/CurlClientTest.php
+++ b/tests/CurlClientTest.php
@@ -6,6 +6,23 @@ use Stripe\HttpClient\CurlClient;
 
 class CurlClientTest extends TestCase
 {
+    public function testTimeout()
+    {
+        $curl = new CurlClient();
+        $this->assertSame(CurlClient::DEFAULT_TIMEOUT, $curl->getTimeout());
+        $this->assertSame(CurlClient::DEFAULT_CONNECT_TIMEOUT, $curl->getConnectTimeout());
+
+        // implicitly tests whether we're returning the CurlClient instance
+        $curl = $curl->setConnectTimeout(1)->setTimeout(10);
+        $this->assertSame(1, $curl->getConnectTimeout());
+        $this->assertSame(10, $curl->getTimeout());
+
+        $curl->setTimeout(-1);
+        $curl->setConnectTimeout(-999);
+        $this->assertSame(0, $curl->getTimeout());
+        $this->assertSame(0, $curl->getConnectTimeout());
+    }
+
     public function testEncode()
     {
         $a = array(
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ echo $charge;
 
 ## Custom Request Timeouts
 
-*NOTE:* If you're decreasing timeouts on non-read-only calls, make sure to use idempotency tokens to avoid executing the same transaction twice as a result of timeout retry logic.
+*NOTE:* We do not recommend decreasing the timeout for non-read-only calls (e.g. charge creation), since even if you locally timeout, the request on Stripe's side can still complete. If you are decreasing timeouts on these calls, make sure to use [idempotency tokens](https://stripe.com/docs/api/php#idempotent_requests) to avoid executing the same transaction twice as a result of timeout retry logic.
 
 To modify request timeouts (connect or total, in seconds) you'll need to tell the API client to use a CurlClient other than its default. You'll set the timeouts in that CurlClient.
 
--- a/README.md
+++ b/README.md
@@ -62,6 +62,27 @@ $charge = Stripe_Charge::create(array('card' => $myCard, 'amount' => 2000, 'curr
 echo $charge;
 ```
 
+## Custom Request Timeouts
+
+*NOTE:* If you're decreasing timeouts on non-read-only calls, make sure to use idempotency tokens to avoid executing the same transaction twice as a result of timeout retry logic.
+
+To modify request timeouts (connect or total, in seconds) you'll need to tell the API client to use a CurlClient other than its default. You'll set the timeouts in that CurlClient.
+
+```php
+// set up your tweaked Curl client
+$curl = new \Stripe\HttpClient\CurlClient();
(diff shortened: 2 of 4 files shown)


iansltx:
If y'all want to set a higher minimum, let me know and I'll revise. But we'll be using this internally with 5-10 second timeouts whether this gets merged or not to avoid having external effects any time y'all go down (like right now).

matthewarkin:
One thing to note about this: The client side timeout value is meant to match the time that the Stripe API will decide to timeout server-side. Changing the value means that there is now an increased likelihood that the Stripe library will return a timeout despite the API request potentially going through.

If you plan on doing this, I'd recommend using idempotency tokens so that you can repeat the API request without risking duplicate charges.

brandur:
@iansltx Nice! Thanks for the contribution here! This looks pretty reasonable to me given that we do allow these [to be set in other bindings languages like Ruby](https://github.com/stripe/stripe-ruby/blob/c5ba20275ef30fe0749ed9cc9977f03f7889a2cd/lib/stripe.rb#L80) and that the defaults are quite optimistic. Also, +1 @matthewarkin in that idempotency keys would be highly recommended in situation with reduced timeouts especially.

One thing that catches me as slightly unusual is that the `CurlClient` is _both_ a singleton and takes static configuration. What do you think about potentially putting the configuration in as instance variables so that you could potentially instantiate multiple `CurlClient`s with different timeout configurations without affecting global state?

It also looks like the builds didn't quite make it through. Mind taking a quick look at that in case there's something worrying?

Will also defer to @bkrausz for general thoughts on the idea.

matthewarkin:
I believe the build failures were causes by the api outage last night, but re-running them would be a good idea.



—

brandur:
> I believe the build failures were causes by the api outage last night, but re-running them would be a good idea.

Good point! I just bumped all the failed builds to see what happens.

matthewarkin:
Builds fixed



—

iansltx:
Fair apoint about the static/instance var bits. Just have to figure out a way to have the instance vars apply when the main interfaces to the lib are static calls. Maybe a setInstance method?

brandur:
> Fair apoint about the static/instance var bits. Just have to figure out a way to have the instance vars apply when the main interfaces to the lib are static calls. Maybe a setInstance method? Will update this PR with what I'm thinking later.

I might be misunderstanding something, but I think the library as is favors non-static calls to `CurlClient`. The flow:

1. The `ApiRequestor` provides a private `httpClient()` method that [defers to the `CurlClient` singleton unless a client object has been set](https://github.com/stripe/stripe-php/blob/64d5a80356cb568f67cde3231fa0ca76e1537b1d/lib/ApiRequestor.php#L227,L233).
2. The `CurlClient`'s `request()` instance method is [then called by the `ApiRequestor` as part of its standard request flow](https://github.com/stripe/stripe-php/blob/64d5a80356cb568f67cde3231fa0ca76e1537b1d/lib/ApiRequestor.php#L173,L179).

iansltx:
Okay...so the static HTTP client setter is what should be used. Which means the CurlClient either needs constructor injection or timeout override instance methods. Which do y'all prefer?

brandur:
> Okay...so the static HTTP client setter is what should be used. Which means the CurlClient either needs constructor injection or timeout override instance methods. Which do y'all prefer?

Personally I'd go with with instance setters. Hopefully we can choose defaults that will work for most people, so hopefully these won't need to be frequently changed.

iansltx:
Revised to use instance methods, added tests, added docs, bugfixed. What do y'all think?

tests/CurlClientTest.php
```diff
@@ -6,6 +6,23 @@
 
 class CurlClientTest extends TestCase
 {
+    public function testTimeout()
```
brandur: Nice! Thanks for adding in the test.

brandur:
@iansltx Nice job! +1 from me.

I'll leave this open for a bit longer to give others from @stripe/api-libraries a chance to comment, but will pull this in absent other feedback.

Thanks again for tackling the issue.

bkrausz:
We have had a lot of concerns in the past about people setting there timeouts very low. Charge creation, given the need to talk to the card network, can easily take 5s when talking to more obscure banks in further away countries.

That said, the bar for doing this seems about right: having to make a new `CurlRequest` object. I'm pretty happy with this.

Can we extend the note in docs to include a warning against a low timeout (in addition to the idempotency suggestion)? 5s is definitely too low for charge creation, we usually warn against anything less than 30s, but maybe suggesting 10s as a lower bar is a better compromise?

CC @remistr if he has support FUD.

iansltx:
Add wording here onc ey'all decide what's appropriate and I'll update the readme in the PR.

Another approach would be to, for transactions that only ping internal Syrioe resourcrs, such as connected account balance inquiry, have a lower timeout. Probably a future enhancement though.

iansltx:
:+1: or :-1: on this? What should the readme's cautionary tales look like?

bkrausz:
Sorry for the delay. Let's change the NOTE message to this, then I'm happy to merge:

    *NOTE:* We do not recommend decreasing the timeout for non-read-only calls (e.g. charge creation), since even if you locally timeout, the request on Stripe's side can still complete. If you are decreasing timeouts on these calls, make sure to use [idempotency tokens](https://stripe.com/docs/api/php#idempotent_requests) to avoid executing the same transaction twice as a result of timeout retry logic.

iansltx:
Changed. Thx in advance for the merge!

bkrausz:
No problem, thanks again for the PR!

PR #216, opened 2016-01-07
Add curl option to encourage earlier curl versions to negotiate TLS 1.2

tetrakai:
This causes curl on RedHat to upgrade from TLS 1.0 to TLS 1.2 by default.
r? @kyleconroy

### DIFF
--- a/lib/HttpClient/CurlClient.php
+++ b/lib/HttpClient/CurlClient.php
@@ -102,6 +102,13 @@ class CurlClient implements ClientInterface
         if (!Stripe::$verifySslCerts) {
             $opts[CURLOPT_SSL_VERIFYPEER] = false;
         }
+        // Opt into TLS 1.x support on older versions of curl. This causes some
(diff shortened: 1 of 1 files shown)


lib/HttpClient/CurlClient.php
```diff
@@ -102,6 +102,13 @@ public function request($method, $absUrl, $headers, $params, $hasFile)
         if (!Stripe::$verifySslCerts) {
             $opts[CURLOPT_SSL_VERIFYPEER] = false;
         }
+        // Opt into TLS 1.x support on older versions of curl. This causes some
+        // curl versions, notably on RedHat, to upgrade the connection to TLS
+        // 1.2, from the default TLS 1.0.
+        if (!defined('CURL_SSLVERSION_TLSv1')) {
```
kyleconroy: This is annoying, but our code linter is barfing on these constants because they aren't all uppercase, even those these are the correct values to use. I'm going to commit a fix for the linter.

PR #218, opened 2016-01-26
Update CA bundle

brandur:
Also switches to the more readily available Mozilla bundle.

See stripe/stripe-ruby#372 for details.


brandur:
Released as 3.7.1.

PR #221, opened 2016-02-05
Set the stripe_account globaly when working on managed account

rikless:
This will automatically add Stripe-Account header and the value if  ```php \Stripe\Stripe::setAccountId() ``` has been used.

This can be very useful if you set your api key with a service provider, and have to deal with managed accounts on many requests.

### DIFF
--- a/lib/Stripe.php
+++ b/lib/Stripe.php
@@ -16,6 +16,9 @@ class Stripe
     // @var string|null The version of the Stripe API to use for requests.
     public static $apiVersion = null;
 
+    // @var string|null The account ID for connected accounts requests.
+    public static $accountId = null;
+
     // @var boolean Defaults to true.
     public static $verifySslCerts = true;
 
@@ -70,5 +73,21 @@ class Stripe
     public static function setVerifySslCerts($verify)
     {
         self::$verifySslCerts = $verify;
+    }
+
+    /**
+     * @return string | null The Stripe account ID for connected accounts requests.
+     */
+    public static function getAccountId()
+    {
+        return self::$accountId;
+    }
+
+    /**
+     * @param string $accountId The Stripe account ID to set for connected accounts requests
+     */
+    public static function setAccountId($accountId)
+    {
+        self::$accountId = $accountId;
     }
 }
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -146,6 +146,11 @@ class ApiRequestor
         if (Stripe::$apiVersion) {
             $defaultHeaders['Stripe-Version'] = Stripe::$apiVersion;
         }
+
+        if (Stripe::$accountId) {
+            $defaultHeaders['Stripe-Account'] = Stripe::$accountId;
+        }
+
         $hasFile = false;
         $hasCurlFile = class_exists('\CURLFile', false);
         foreach ($params as $k => $v) {


brandur:
> This can be very useful if you set your api key with a service provider, and have to deal with managed accounts on many requests.

Do you tend to use the same managed account for the entire lifetime of your program?

My only worry here is that people will use this more as a helper for making batches of requests for a single account more convenient. This is a bit problematic because since it's global state, it's not entirely safe in that one thread may still be making a request while another resets this account out from under it.

rikless:
No at all. Actually, a Service Provider can determine if the current request is owned by a managed account. If it's true, then you can use this setter to push all stripe requests with correct configuration, instead of setting multiple parameters on different places.

The service provider will set the value, if needed, only on current request life cycle.

To be more explicit, all services will just have to be written this way : 

```php
//StripeServiceProvider class

\Stripe\Stripe::setApiKey(PLATFORM_SECRET_KEY);

if ($request->isManagedAccount()) {

    \Stripe\Stripe::setAccountId($account_id);

}
```

```php
\Stripe\Customer::create(
  array("description" => "example@stripe.com")
);
```

Instead of
```php

if($this->managedAccount())
    \Stripe\Customer::create(
      array("description" => "example@stripe.com"),
      array("stripe_account" => CONNECTED_STRIPE_ACCOUNT_ID)
    );
}
else{
     \Stripe\Customer::create(
          array("description" => "example@stripe.com"),
        );
}
```

This can save a lot of time when building an app to abstract this management.

matthewarkin:
```` 
  \Stripe\Customer::create(
      array("description" => "example@stripe.com"),
      array("stripe_account" => NULL)
    );
````

should work fine

**Comment by rikless** [comment]
Sure it works.

This is just a way to let developer making the stripe_account management on 1 place for all the application.

When the Service Provider is loaded, I can use all stripe objects without testing if I'm in a managed account request, or not, and find the correct stripe_account id, because this logic is made on one place.

**Comment by brandur** [comment]
Cool! Alright, I'm still not crazy about global state, but admittedly we still don't have a great story around separate Stripe client objects that could local settings, so I'm okay with pulling this in with the caveat that it could be retooled at some point down the line.

@rikless Would you mind squashing your commits? A `git rebase -i master` is generally the easiest/fastest way to do this. Thanks!

File: lib/Stripe.php [hunk 1]
```diff
@@ -71,4 +74,20 @@ public static function setVerifySslCerts($verify)
     {
         self::$verifySslCerts = $verify;
     }
+
+    /**
+     * @return string | null The stripe_account id for requests.
```
brandur: And actually, mind just changing these two comments to read "The Stripe account ID for managed account requests." (Mostly just changing the casing on "stripe_account".)

lib/Stripe.php
```diff
@@ -16,6 +16,9 @@ class Stripe
     // @var string|null The version of the Stripe API to use for requests.
     public static $apiVersion = null;
 
+    // @var string|null The account id for managed account requests.
```
bkrausz: Sorry for the chunder...this is not managed accounts only. This should be `connected account`. It also works with your own account ID, but the purpose is connected accounts, whether managed or standalone.

rikless:
Done

brandur:
Thanks!

brandur:
Amended comments a bit in 2d85b1a1a651476fb8b3be29be2efa86420df0eb and released in 3.9.0.

rikless:
Thank you @brandur :)

brandur:
> Thank you @brandur :)

Anytime! Thanks again for the pull!

PR #223, opened 2016-02-06
Added support for CountrySpec to the bindings

remistr:
Added support for `CountrySpec` and basic tests.

r? @brandur 
cc @gregsabo @bkrausz  for awareness

I wasn't sure whether we wanted to hardcode some assumptions about US since we don't rely on mocked data in the bindings.


brandur:
Amazing work Remi!! Thanks!

brandur:
Released as 3.8.0.

PR #226, opened 2016-02-17
Fixed error messages to remove mentions of support@stripe.com

olivierbellone:
This PR fixes error messages that mention support@stripe.com. Users sometimes surface the error messages to customers, and customers are understandably confused when told to contact Stripe's support.

Users should already know how to contact support anyway so they shouldn't be affected by this.

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -121,9 +121,9 @@ class ApiRequestor
 
         if (!$myApiKey) {
             $msg = 'No API key provided.  (HINT: set your API key using '
-              . '"Stripe::setApiKey(<API-KEY>)".  You can generate API keys from '
-              . 'the Stripe web interface.  See https://stripe.com/api for '
-              . 'details, or email support@stripe.com if you have any questions.';
+              . '"Stripe::setApiKey(<API-KEY>)".  You can find your API keys '
+              . 'in your Stripe dashboard: '
+              . 'https://dashboard.stripe.com/account/apikeys.';
             throw new Error\Authentication($msg);
         }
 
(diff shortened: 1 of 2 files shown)


olivierbellone:
r? @brandur 
cc @stripe/api-libraries

lib/ApiRequestor.php
```diff
@@ -121,9 +121,9 @@ private function _requestRaw($method, $url, $params, $headers)
 
         if (!$myApiKey) {
             $msg = 'No API key provided.  (HINT: set your API key using '
-              . '"Stripe::setApiKey(<API-KEY>)".  You can generate API keys from '
-              . 'the Stripe web interface.  See https://stripe.com/api for '
-              . 'details, or email support@stripe.com if you have any questions.';
+              . '"Stripe::setApiKey(<API-KEY>)".  You can find your API keys '
```
russelldavis: If "Users sometimes surface the error messages to customers" is the problem, mentioning the stripe dashboard is also going to cause confusion.

matthewarkin: Don't a number of api returned messages return mentions of support and the dashboard? Like `unable to use live transfers, please email support` or `managed accounts not enabled please register your platform`? It seems like this is poor error handling on behalf of the person integrating Stripe

brandur:
This seems reasonable to me if it's still the direction that we want to go.

@olivierbellone Would you mind taking another look at this one and fielding Russell's question above? Thanks!

remistr:
I let this slide for a few weeks to think about this more. Recent downtime/issue doesn't seem to have been impacted as much as it was 6+ months ago. Let's close for now and revisit in a future iteration if needed.

PR #230, opened 2016-02-25
Remove duplicated line on CurlClient.php

loslch:
It's duplicated code on 96 and 99 of CurlClient.php.


brandur:
+1! Thanks @loslch!

PR #232, opened 2016-03-03
Fix fatal error when setting a $nestedUpdatableAttributes attr more than once

saber3005:
Currently updating any nested attribute (which is not already set) more than once causes a fatal error because `StripeObject` assumes that it must be an `AttachedObject` when it's only an array. I've added an `instanceof` check to ensure this doesn't happen.


brandur:
@saber3005 Awesome fix, and thanks for the test!

PR #234, opened 2016-03-21
Allow setting default CURLOPT_* on HTTP calls

iansltx:
Can be supplied as a static array or as a callable, which will be
invoked with the same arguments the request method is invoked with, and
should return an array (same format as curl_setopt_array()).

### DIFF
--- a/lib/HttpClient/CurlClient.php
+++ b/lib/HttpClient/CurlClient.php
@@ -23,7 +23,7 @@ class CurlClient implements ClientInterface
     /**
      * CurlClient constructor.
      *
-     * Pass in a callable to $defaultIptions that returns an array of CURLOPT_* values to start
+     * Pass in a callable to $defaultOptions that returns an array of CURLOPT_* values to start
      * off a request with, or an flat array with the same format used by curl_setopt_array() to
      * provide a static set of options. Note that many options are overridden later in the request
      * call, including timeouts, which can be set via setTimeout() and setConnectTimeout().
--- a/lib/HttpClient/CurlClient.php
+++ b/lib/HttpClient/CurlClient.php
@@ -82,7 +82,7 @@ class CurlClient implements ClientInterface
 
         $opts = array();
         if (is_callable($this->defaultOptions)) { // call defaultOptions callback, set options to return value
-            $opts = call_user_func($this->defaultOptions, $method, $absUrl, $headers, $params, $hasFile);
+            $opts = call_user_func_array($this->defaultOptions, func_get_args());
             if (!is_array($opts)) {
                 throw new Error\Api("Non-array value returned by defaultOptions CurlClient callback");
             }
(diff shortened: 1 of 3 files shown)


lib/HttpClient/CurlClient.php
```diff
@@ -18,6 +18,31 @@ public static function instance()
         return self::$instance;
     }
 
+    protected $defaultOptions;
+
+    /**
+     * CurlClient constructor.
+     *
+     * Pass in a callable to $defaultIptions that returns an array of CURLOPT_* values to start
```
brandur: Typo here: should be `$defaultOptions`.

brandur: Thanks for adding the documentation though! We don't have enough of it :$

brandur:
@iansltx Looks good to me! Thanks for getting in those tests and documentation.

Just a couple questions:

1. Is the closure option largely a hypothetical addition at the moment or is this something that you needed?
2. Regarding the closure again: what do you think about possible passing a class or associative array as an argument instead of individual args for each characteristic of the request? I'm a tad worried that we'll need to add another argument at some point in the future and that this will break compatibility for anyone that's currently using the feature. A class or array would future-proof the call.

Thanks again for putting this together!

And lastly, calling @bkrausz in case he wants to take a pass on this as well!

iansltx:
1. All I need is the static array option; the closure was what I thought of first because that allows hooking into the request method which may have other uses (debug/logging), but it's not necessary for my use case.
2. The only time that call would change is if the signature of request() changed, which would probably result in a change to ClientInterface anyway, resulting in a BC break for the whole thing, unless you're just adding an optional param to request(), in which case the call_user_func call would just tack on that extra param, and any closure using those params would continue working.

That said, it's probably clearer (and requires less code changes in the future) if that call explicitly passes on the parameters of the request() call, rather than using the individual vars. Making that change now.

Lastly, if y'all don't see any value in having the callable option, I don't mind re-PRing without it; just say the word :) Again, my own use case is covered by the non-callable option...I just promised I'd show a PoC for the callable one.

brandur:
> The only time that call would change is if the signature of request() changed, which would probably result in a change to ClientInterface anyway, resulting in a BC break for the whole thing, unless you're just adding an optional param to request(), in which case the call_user_func call would just tack on that extra param, and any closure using those params would continue working.

Cool. I'd normally say that the arguments to request are largely package-internal while this is external, but I guess we've been recommending that people override `ClientInterface` anyway, so that's not really true.

> That said, it's probably clearer (and requires less code changes in the future) if that call explicitly passes on the parameters of the request() call, rather than using the individual vars. Making that change now.

+1. That looks cleaner.

> Lastly, if y'all don't see any value in having the callable option, I don't mind re-PRing without it; just say the word :) Again, my own use case is covered by the non-callable option...I just promised I'd show a PoC for the callable one.

Got it. I'd normally try to reduce the surface area of anything that we expose publicly, but given that it's already done and seems useful, I don't mind having the extra hook in there.

I'll wait ~1 day for comments from anybody else on the patch, but barring anything else, will bring it in tomorrow.

brandur:
Released as 3.11.0.

iansltx:
Thanks! Just updated project deps to point back at this repo. :+1:

PR #235, opened 2016-03-22
include $params and $options when refunding an application fee

MDrollette:
This fixes a regression introduced by https://github.com/stripe/stripe-php/pull/209

Note that partial application fee refunds were failing silently (or rather, succeeding with the incorrect refund amount) since that change.


brandur:
Ugh. I'm very sorry about the trouble here. Thanks a lot for the patch!

brandur:
Fix released in 3.10.1.

PR #241, opened 2016-03-31
Use correct opts parameter

theboolean:
`$options` variable is not declared anywhere in the method. I think this is simply a typo: use passed `$opts` instead.


brandur:
Thanks for all the fixes @theboolean! They've all been released as part of 3.12.0.

theboolean:
You're welcome @brandur!

PR #242, opened 2016-04-05
Update dispute test to better match docs

matthewarkin:
mainly doing this to not have to install php 5.4


matthewarkin:
just wanted to run that through travis

PR #244, opened 2016-04-13
allow retrieval, listing, and creation of subs, change URL to v1/subs

jxu577:
cc @kyleconroy @cupcait

### DIFF
--- a/lib/Subscription.php
+++ b/lib/Subscription.php
@@ -5,26 +5,36 @@ namespace Stripe;
 class Subscription extends ApiResource
 {
     /**
-     * @return string The API URL for this Stripe subscription.
+     * @param string $id The ID of the subscription to retrieve.
+     * @param array|string|null $opts
+     *
+     * @return Subscription
      */
-    public function instanceUrl()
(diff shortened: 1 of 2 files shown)


brandur:
@jxu577 Looks promising so far!

I'm not totally sure how you want to proceed here, so let me know what you're thinking.

And also just a heads up: it looks like at least one of the test failures is related to your changes.

jxu577:
Yep, that is a known error. I was going to tag the subs team for review on these, so you don't need to worry about reviewing things :) unless you would like to for consistency sake.

GrahamCampbell:
These changes are major and breaking though.

brandur:
> These changes are major and breaking though.

@GrahamCampbell Hey! How so? This is largely an internal re-arrangement of the API and shouldn't really have much of an effect at all on the API exposed by the bindings (i.e. retrieval will still work, but now it goes to a different URL).

kyleconroy:
@jxu577 this looks good! Tests will pass once the discount endpoint is up and running.

kyleconroy:
:+1: :ship: once we're ready to go

Issue #251, opened 2016-05-05
Detect if mbstring is available and output an explicit error message if it isn't


olivierbellone:
A relatively common issue with the PHP bindings is that they depend on the `mbstring` extension ([here](https://github.com/stripe/stripe-php/blob/e40211d1d5325d99d3ce167c9ee1ade0473a608a/lib/Util/Util.php#L119)), which isn't necessarily available everywhere.

In particular, on RedHat / CentOS systems, it doesn't come out of the box with PHP and needs to be installed via a separate package.

I was thinking it could be helpful to add a check when the bindings are loaded. Something like:
```php
if (!extension_loaded('mbstring')) {
    die("It looks like the mbstring extension is not loaded. "
      . "The Stripe PHP bindings cannot work properly without this extension. "
      . "Please ask your system administrator to enable the mbstring extension, "
      . "or email support@stripe.com if you have any questions.");
}
```

called from `init.php`.

While we're at it, we could do a similar check for the `curl` extension (although it's much less common for this extension to not be available).

Also, we should explicitly list these dependencies in the README.

Thoughts? @stripe/api-libraries

kyleconroy:
Seems like a reasonable addition. @bkrausz may have more state on if this is the best

bkrausz:
The only mbstring method we use is `mb_detect_encoding` to decide if we should `utf8_encode`. We can instead add a `function_exists('mb_detect_encoding')` to that call (preferably with class var caching since I think the perf of `function_exists` sucks and it'd be called a lot). May still be worth a warning about poor UTF8 handling, but if they don't have `mbstring` installed I think they already have UTF8 problems.

Also putting in `init` would only work for those not using Composer, so we'd likely want somewhere else (like in ApiRequestor, which always gets called on request).

I think we used to do a check for Curl, but Composer enforces its existence, and we now provide pluggable request handlers, so you could opt to swap in your own non-Curl requestor if you're so inclined.

matthewarkin:
does the ext-mbstring dependency in the composer.json (https://github.com/stripe/stripe-php/blob/master/composer.json#L21) not handle the composer case?

GrahamCampbell:
Composer already deals with this. Nothing else is needed IMO.

olivierbellone:
Tentatively fixed by #251.

PR #252, opened 2016-05-06
mbstring availability check

olivierbellone:
This is a tentative PR for #251.

Upon the first call to `utf8()`, it will check if mbstring is enabled, and trigger a warning if it isn't.

I have slight FUD that users will miss the warning however. What do you think? Maybe raise an exception / `die` rather than just a warning?

r? @bkrausz 
cc @stripe/api-libraries

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,16 @@ If you do not wish to use Composer, you can download the [latest release](https:
 require_once('/path/to/stripe-php/init.php');
 ```
 
+## Dependencies
+
+The bindings require the following extension in order to work properly:
+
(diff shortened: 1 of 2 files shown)


bkrausz:
I'm pretty OK still working for non-UTF-8 here. If they don't have mbstring installed I think they'll have more int'l problems than just us.

lgtm - can you test the warning code and make sure it works properly in PHP 5.3? Assuming it does 🚢 .

olivierbellone:
I tested it with PHP 5.3.29 with no issues.

PR #260, opened 2016-05-25
Support order returns.

rasmusrygaard:
r? @stripe/api-libraries 
cc @jimdanz

### DIFF
--- a/init.php
+++ b/init.php
@@ -54,6 +54,7 @@ require(dirname(__FILE__) . '/lib/FileUpload.php');
 require(dirname(__FILE__) . '/lib/Invoice.php');
 require(dirname(__FILE__) . '/lib/InvoiceItem.php');
 require(dirname(__FILE__) . '/lib/Order.php');
+require(dirname(__FILE__) . '/lib/OrderReturn.php');
 require(dirname(__FILE__) . '/lib/Plan.php');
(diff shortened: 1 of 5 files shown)


rasmusrygaard:
I think we need to upgrade the API version to avoid these test failures, thoughts on whether or not that's doable? I believe we'll need the new version in the Java libraries as well

brandur:
Looks good to me! +1.

> I think we need to upgrade the API version to avoid these test failures, thoughts on whether or not that's doable? I believe we'll need the new version in the Java libraries as well

[I was just checking the changes](https://stripe.com/docs/upgrades?since=2015-08-19#api-changelog) and nothing seems insurmountable. I recently went through the process of upgrading the version on the Go bindings and it was time consuming, but ultimately not too bad.

I think we should just go for it and fix anything that does happen to come up.

brandur:
I just triggered the upgrade and restarted the build.

brandur:
@rasmusrygaard I think you might have branched off quite an old `master`. If you pull the latest and rebase, you should be able to get the test suite to pass.

rasmusrygaard:
That could very well be. Rebased and it looks like we're in the clear

brandur:
Released as 3.14.0.

PR #262, opened 2016-05-31
Add few class' constants corresponding to subscription statuses

bocharsky-bw:


brandur:
Hi @bocharsky-bw,

This looks useful, thank-you! Could we also do the following:

1. Alphabetize the list so if new statuses are added, they'll have a logical home.
2. Prefix the names with `STATUS_` to make it obvious what they are and to avoid any future naming collisions.
3. Adding a comment above the list of constants indicating that they're possible representations of the status field.

brandur:
Hey @bocharsky-bw, thanks for making those changes!

It looks like CI is failing here due to some whitespace being detected at the end of one of the new lines. Would you mind taking a look (and squashing the change) if possible? Thanks!

And feel free to bump the PR when you're done so that I don't forget to take a look at it again :)

bocharsky-bw:
@brandur Fixed!

brandur:
LGTM.

Thanks!

brandur:
Released as 3.18.0.

PR #263, opened 2016-06-03
add SKU inventory to $nestedUpdatableAttributes

guillaumehenriot:
r? @brandur 

Added the SKU inventory to the $nestedUpdatableAttributes array to allow updates.
Also I added inventory as to a new line to avoid putting it in the middle of the account ones but it's not the cleanest.


brandur:
LGTM.

Thanks!

brandur:
Released as 3.14.2.

PR #266, opened 2016-06-15
Add static `update` method

kyleconroy:
You can now update API resources with first calling retrieve.

### DIFF
--- a/lib/ApiResource.php
+++ b/lib/ApiResource.php
@@ -66,7 +66,7 @@ abstract class ApiResource extends StripeObject
     /**
      * @return string The instance endpoint URL for the given class.
      */
-    public static function classInstanceUrl($id)
+    public static function resourceUrl($id)
     {
         if ($id === null) {
             $class = get_called_class();
@@ -85,7 +85,7 @@ abstract class ApiResource extends StripeObject
      */
     public function instanceUrl()
     {
-        return static::classInstanceUrl($this['id']);
+        return static::resourceUrl($this['id']);
     }
 
     private static function _validateParams($params = null)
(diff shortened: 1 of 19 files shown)


lib/ApiResource.php
```diff
@@ -64,11 +64,10 @@ public static function classUrl()
     }
 
     /**
-     * @return string The full API URL for this API resource.
+     * @return string The instance endpoint URL for the given class.
      */
-    public function instanceUrl()
+    public static function classInstanceUrl($id)
```
brandur: I _think_ you could just call this `instanceUrl` as well. The disambiguation is kind of nice, but I'm not crazy about starting a convention to prefix static functions with `class*`.

lib/ApiResource.php
```diff
@@ -151,6 +158,18 @@ protected static function _create($params = null, $options = null)
         return $obj;
     }
 
+    protected static function _update($id, $params = null, $options = null)
```
brandur: Just in the interest of leaving things better than we found them, mind putting some documentation on this function?

brandur:
@kyleconroy Left a couple minor comments above, but overall looks good to me. Thanks!

lib/ApiResource.php
kyleconroy: +1 can do

lib/ApiResource.php
kyleconroy: Sadly you can't have a static function and non-static function with the same name.

lib/ApiResource.php
```diff
@@ -64,11 +64,10 @@ public static function classUrl()
     }
 
     /**
-     * @return string The full API URL for this API resource.
+     * @return string The instance endpoint URL for the given class.
      */
-    public function instanceUrl()
+    public static function resourceUrl($id)
```
brandur: +1 on the new name! Thanks.

brandur:
@kyleconroy LGTM!

Let me know if you have any particular ideas around how/when you want to bring this in. Otherwise, I can coordinate a release if you like.

kyleconroy:
I was planning on merging to master and doing a release. Is there anything else that needs to go out?

brandur:
> I was planning on merging to master and doing a release. Is there anything else that needs to go out?

Nope! Aside from these changes, master should be clean right now.

lib/ApiResource.php
```diff
@@ -151,6 +158,25 @@ protected static function _create($params = null, $options = null)
...
+     * @param string $id The ID of the API resource to update.
+     * @param array|null $params
+     * @param array|string|null $opts
+     *
+     * @return ApiResource the updated API resource
+     */
+    protected static function _update($id, $params = null, $options = null)
+    {
+        self::_validateParams($params);
+        $base = static::baseUrl();
```
arcanedev-maroc: I think you can remove this line

PR #267, opened 2016-06-20
Empty Expect: header to prevent 100-Continue responses

nickma42:
I spent a few days recently debugging a series of seemingly inexplicable `Unexpected error communicating with Stripe. If this problem persists, let us know at support@stripe.com. (Network error [errno 52]: Empty reply from server)` errors we were noticing on our servers recently.

It seemed to happen repeatedly for certain users while other users were fine which ruled out any connectivity issue and nothing crazy turned up in a network packet capture but it turns out cURL by default sends an `Expect: 100-continue` header and for whatever reason, the subsequent request with the POST data never made it to the Stripe API.

There's a pretty good write-up here about the behavior: https://support.urbanairship.com/entries/59909909--Expect-100-Continue-Issues-and-Risks

We found setting `Expect:` as a header stopped the problem.

### DIFF
--- a/lib/HttpClient/CurlClient.php
+++ b/lib/HttpClient/CurlClient.php
@@ -126,6 +126,9 @@ class CurlClient implements ClientInterface
             return strlen($header_line);
         };
 
+        // Send empty expect header to avoid 100-Continue responses
+        array_push($headers, 'Expect: ');
+
         $absUrl = Util\Util::utf8($absUrl);
         $opts[CURLOPT_URL] = $absUrl;
         $opts[CURLOPT_RETURNTRANSFER] = true;


brandur:
Thanks!

Looking into this a little more, it seems that cURL will only add the "Expects: 100-continue" header [if the request body size is greater than 1024 bytes](https://github.com/curl/curl/blob/5f2e3b886759e0822ff31c36ef10ca8df59fcf59/lib/http.c#L2619,L2622). That would explain why you're the first to have reported this problem — most request bodies are probably small enough that they're been flying under the radar.

Technically, this is actually a different class of bug in that our bindings should be able to handle the empty server response and then retry, but until that can be fixed, this seems like quite a reasonable interim fix. Thanks

LGTM.

brandur:
Released as version 3.14.3.

I've also added commit ad16ae7 in order to better document the problem and solution.

nickma42:
@brandur thanks - fastest PR response I think I've ever seen! We're possibly a bit too liberal with our metadata, and the >1024b condition explains why it was only happening for certain users/purchases. Thanks again, much appreciated.

Issue #250, opened 2016-05-05
Documenting All Classes


MarkVaughn:
I'd volunteer to add class level phpdoc blocks to all classes.

Example: 
Customer.php
```php
<?php

namespace Stripe;

/**
 * Class Customer
 *
 * @property string $id
 * @property string $object
 * @property int $account_balance
 * @property string $business_vat_id
 * @property string $created
 * @property string $currency
 * @property string $default_source
 * @property bool $delinquent
 * @property string $description
 * @property mixed $discount
 * @property string $email
 * @property bool $livemode
 * @property array $metadata
 * @property mixed $shipping
 * @property array $sources
 * @property array $subscriptions
 *
 * @package Stripe
 */
class Customer extends ApiResource
```

This would help with development in IDE's such as phpstorm
![autocompletion](https://cloud.githubusercontent.com/assets/39970/15052384/691afa02-12cb-11e6-9ada-9a19d7cc3e80.png)

Again I'd volunteer to do it and put up a PR, but I don't want to do it all and then get some reason why you guys would reject it for some reason.
Let me know

arcanedev-maroc:
Check this one: [ARCANEDEV/Stripe](https://github.com/ARCANEDEV/Stripe)

brandur:
Hi @MarkVaughn, and I guess the proposal here is mostly to just get types in rather than an additional description as well?

My only fear here is that we don't really have a great way of ensuring the accuracy or timeliness of update of such an addition. That said, I wouldn't be too adverse to the idea if you want to go for a pull request.

MarkVaughn:
Yea, just some sort of bare minimum better than nothing.
Also knowing what fields are potentially available. I'll get on it then. 
Cheers

brandur:
>  Also knowing what fields are potentially available. I'll get on it then. 

Makes sense! And cool, thanks for the effort here.

jarrodbell:
Any update on this @MarkVaughn ? I'd love this for working with the API in PhpStorm

MarkVaughn:
@jarrodbell @brandur made a pr, but apparently it is "useless"

GrahamCampbell:
> made a pr, but apparently it is "useless"

That wasn't what I said. I made an inline comment on a specific line.

MarkVaughn:
[...]

PR #268, opened 2016-07-04
Added basic class level docs

MarkVaughn:
Added some basic properties for #250

This provides some minimal structure, is not comprehensive and can be improved upon.

### DIFF
--- a/lib/Account.php
+++ b/lib/Account.php
@@ -2,6 +2,38 @@
 
 namespace Stripe;
 
+/**
+ * Class Account
+ *
+ * @property string $id
+ * @property string $object
+ * @property mixed $business_logo
+ * @property string $business_name
+ * @property mixed $business_url
+ * @property bool $charges_enabled
+ * @property string $country
+ * @property bool $debit_negative_balances
+ * @property mixed $decline_charge_on
+ * @property string $default_currency
+ * @property bool $details_submitted
+ * @property string $display_name
+ * @property string $email
+ * @property mixed $external_accounts
+ * @property mixed $legal_entity
+ * @property bool $managed
+ * @property mixed $product_description
+ * @property mixed $statement_descriptor
+ * @property mixed $support_email
+ * @property mixed $support_phone
+ * @property string $timezone
+ * @property mixed $tos_acceptance
+ * @property mixed $transfer_schedule
+ * @property bool $transfers_enabled
+ * @property mixed $verification
+ * @property mixed $keys
+ *
+ * @package Stripe
+ */
 class Account extends ApiResource
 {
     public function instanceUrl()
--- a/lib/Customer.php
+++ b/lib/Customer.php
@@ -2,6 +2,28 @@
 
 namespace Stripe;
 
+/**
+ * Class Customer
+ *
+ * @property string $id
+ * @property string $object
+ * @property int $account_balance
+ * @property string $business_vat_id
+ * @property string $created
+ * @property string $currency
+ * @property string $default_source
+ * @property bool $delinquent
+ * @property string $description
+ * @property mixed $discount
+ * @property string $email
+ * @property bool $livemode
+ * @property array $metadata
+ * @property mixed $shipping
+ * @property array $sources
+ * @property array $subscriptions
+ *
+ * @package Stripe
+ */
 class Customer extends ApiResource
 {
     /**
--- a/lib/Customer.php
+++ b/lib/Customer.php
@@ -19,8 +19,8 @@ namespace Stripe;
  * @property bool $livemode
  * @property array $metadata
  * @property mixed $shipping
- * @property array $sources
- * @property array $subscriptions
+ * @property Collection $sources
+ * @property Collection $subscriptions
  *
  * @package Stripe
  */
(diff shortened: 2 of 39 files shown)


MarkVaughn:
not sure why adding comments to code is causing travis to throw unknown errors.

lib/Account.php
```diff
@@ -2,6 +2,38 @@
 
 namespace Stripe;
 
+/**
+ * Class Account
```
GrahamCampbell: These comments are useless

lib/Account.php
```diff
@@ -2,6 +2,38 @@
...
+ * @property mixed $support_email
+ * @property mixed $support_phone
+ * @property string $timezone
+ * @property mixed $tos_acceptance
+ * @property mixed $transfer_schedule
+ * @property bool $transfers_enabled
+ * @property mixed $verification
+ * @property mixed $keys
+ *
+ * @package Stripe
```
GrahamCampbell: Incorrect

lib/Account.php
MarkVaughn: they are not useless please see #250 and why this helps

lib/Account.php
MarkVaughn: Care to elaborate or are you just having a bad day?

lib/Account.php
GrahamCampbell: I was referring ONLY to that line I commented on. Not the annotations. :)

lib/Account.php
GrahamCampbell: `@package` is only meant to emulate namespaces on PHP 5.2. It has no other use.

lib/Customer.php
```diff
@@ -2,6 +2,28 @@
...
+ * @property string $default_source
+ * @property bool $delinquent
+ * @property string $description
+ * @property mixed $discount
+ * @property string $email
+ * @property bool $livemode
+ * @property array $metadata
+ * @property mixed $shipping
+ * @property array $sources
+ * @property array $subscriptions
```
dannyhajj: I believe `$sources` and `$subscriptions` are of type `\Stripe\Collection`. Using the collection class instead of `array`, would allow IDEs to recommend the methods that we can use on those two properties. Like `retrieve`, `create`, etc.

lib/Account.php
GrahamCampbell: > Care to elaborate or are you just having a bad day?

Nope, I'm telling you that I wouldn't put this there because it's not correct.

MarkVaughn: So the correct package is `stripe/stripe-php` ?

brandur:
Hi @MarkVaughn, thanks a lot for the huge amount of effort here. It's definitely appreciated!

So I hate to reneg a bit on what I said in the other thread, but I think there's going to be two basic problems with this approach:

1. The PHP bindings are designed to be dynamic such that they can handle responses from a number of different API versions that the user may have set (or to put that better, they're not tied to a particular version like the Go bindings are for example). This may result in a user seeing properties in an IDE that don't exist for their API version, or the opposite problem where they have properties that are not documented in the IDE.
2. As mentioned elsewhere, this will be a bit of a maintenance problem in that we don't currently have any sort of working process for keeping these up-to-date when we make changes in the API.

Given that these are just hints for PHP IDEs, they're not the end of the world, but I mainly wanted to point out that some subtle confusion is possible.

I think merging would be okay, but with the caveat that we may revert if we have too many confused users writing in. I'm not sure how like this is.

Let me know if you have any other thoughts here, and thanks again!

MarkVaughn:
I vote for any documentation is better than no documentation.

---

Review status: 0 of 39 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[lib/Account.php, line 6 \[r1\]](https://reviewable.io:443/reviews/stripe/stripe-php/268#-KMZQUmf01RMZ8wSUKkD-r1-6:-KMZQsK03tSRBnT829oe:-1826134749) ([raw file](https://github.com/stripe/stripe-php/blob/d414571b18c2ba87279c7083acfdc6497f630da8/lib/Account.php#L6)):*
<details><summary><i>Previously, GrahamCampbell (Graham Campbell) wrote…</i></summary>
> I was referring ONLY to that line I commented on. Not the annotations. :)
</details>
The purpose is differentiating between classes interfaces and traits. It's a header usually followed by a description paragraph describing the purpose which I did not write. My intention is to get the ball rolling on starting some sort of template here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/stripe/stripe-php/268#-:-KMZQYiakUBWiM0Sab3g:-1352569415)*

brandur:
@MarkVaughn Cool! Alright, I'm pulling this in (with the small caveat above). Thanks again for working on it!

brandur:
Released as 3.17.0.

lib/Account.php
GrahamCampbell: No, that's not what the package annotation means.

GrahamCampbell:
> I vote for any documentation is better than no documentation.

No it's not? The class name is already self documenting.

PR #273, opened 2016-07-14
Throw exception in DisputeTest.php

dino001:
Throw exception object instead of string.


brandur:
LGTM! Thanks @dino001.

Unfortunately, another pull seemed to have introduced a build problem on master. I've recently pulled a fix in by way of #274, but I think it's failing your branch here. Would you mind rebasing on master just to double check that the tests pass?

brandur:
Oops, sorry, I missed that the changes here had been pushed. Thank-you!

Small point for future reference: it's often preferable to pull down the latest version of master and then rebase against it rather than merging master into your new branch. For example:

``` sh
git checkout master
git pull origin master
git checkout my-branch
git rebase master
git push -f origin my-branch
```

PR #274, opened 2016-07-14
Remove some newlines that php_codesniffer doesn't like

brandur:
Master is breaking due to what looks like a style problem.


brandur:
Super minor change. Pulling this in.

LGTM.

PR #278, opened 2016-07-27
I guess you forgot to set the $item

Nakilon:


olivierbellone:
@Nakilon: Could you provide an example of the issue that this PR fixes?

According to the [PHP docs](https://secure.php.net/manual/en/function.next.php), `next()` should return `FALSE` when there are no more elements to return:
> Returns the array value in the next place that's pointed to by the internal array pointer, or FALSE if there are no more elements.

Nakilon:
1) I suppose we inherit the Iterator class, not Array: http://php.net/manual/en/iterator.next.php  
2) I'm not coding in PHP and I don't have the environment to run this stuff.  
I'm just reviewing my colleague's code and when he used `foreach` and `[]=` to convert the whole iterator to array, I asked him why he didn't use `[iterator_to_array()](http://php.net/manual/en/function.iterator-to-array.php)` -- he says "it didn't work" and refuses to explain. So I looked into this function and assumed to find a bug.

Nakilon:
I don't understand how `foreach` "did work" for him though.

olivierbellone:
@Nakilon: the auto-paging iterator iterates over both items and pages. With your change, the iterator only returns the first item from each page: https://travis-ci.org/stripe/stripe-php/jobs/147779064#L272

You are correct that `iterator_to_array()` doesn't seem to work correctly with the auto-paging iterator. I'll dig further.

Nakilon:
Probably that's not "the first item from each page", but "all the items without the first item of each page starting from the second one", that means _smth_ is already calling the `next()` method that I thought was missing.

olivierbellone:
You're correct, your patch skips the first item on every page besides the first.

I think `AutoPagingIterator`'s implementation of the `Iterator` interface is a bit unorthodox. It works for `foreach` but not for `iterator_to_array`. After digging a bit, @remistr found that this is because the `key()` method returns the key of each item within its own page, so if there's more than one page the same key might match different items over the entire collection.

I'm going to close this PR, but feel free to reply here or on #279.

PR #279, opened 2016-07-27
The key() method should return a unique value

remistr:
When you use iterator_to_array() it expects you to return keys by default.
Each item needs to have its own key to avoid overwriting another element.
We now keep track of the offset for each element we access based on the current page.

@olivierbellone what do you think of this fix?

### DIFF
--- a/lib/Util/AutoPagingIterator.php
+++ b/lib/Util/AutoPagingIterator.php
@@ -6,6 +6,7 @@ class AutoPagingIterator implements \Iterator
 {
     private $lastId = null;
     private $page = null;
+    private $pageOffset = 0;
     private $params = array();
 
     public function __construct($collection, $params)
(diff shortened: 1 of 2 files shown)


olivierbellone:
See #278 for some context around the issue this fixes.

lib/Util/AutoPagingIterator.php
```diff
@@ -7,6 +7,7 @@ class AutoPagingIterator implements \Iterator
     private $lastId = null;
     private $page = null;
     private $params = array();
+    private $pageOffset = 0;
```
olivierbellone: Could you put this one line above so that the properties are in alphabetical order?

olivierbellone:
I added a test to make sure `iterator_to_array()` works as expected.

r? @brandur

brandur:
LGTM!

@remistr Thanks for getting this assembled and @olivierbellone thanks for the great test! Pulling this in.

brandur:
Sorry about the delay here, but now released as 3.17.1.

PR #280, opened 2016-08-05
add decline code to stripe api Card error

thecrogers:
This PR will add direct access to decline_code from Error\Card. To give a better understanding of Card Errors.

### DIFF
--- a/lib/Error/Card.php
+++ b/lib/Error/Card.php
@@ -16,6 +16,7 @@ class Card extends Base
         parent::__construct($message, $httpStatus, $httpBody, $jsonBody, $httpHeaders);
         $this->stripeParam = $stripeParam;
         $this->stripeCode = $stripeCode;
+        $this->stripeDeclineCode = isset($jsonBody["error"]["decline_code"]) ? $jsonBody["error"]["decline_code"] : null;
     }
 
     public function getStripeCode()
@@ -26,5 +27,10 @@ class Card extends Base
     public function getStripeParam()
     {
         return $this->stripeParam;
+    }
+
+    public function getStripeDeclineCode()
+    {
+        return $this->stripeDeclineCode;
     }
 }


lib/Error/Card.php
```diff
@@ -27,4 +28,9 @@ public function getStripeParam()
     {
         return $this->stripeParam;
     }
+
+    public function getStripeDeclineCode()
+    {
+        return $this->stripeDeclineCode;
```
GrahamCampbell: This property is not defined?

thecrogers: @GrahamCampbell are you asking a question or making a statement.

GrahamCampbell: I'm saying I think you forgot to define it.

thecrogers: @GrahamCampbell thanks for pointing that out, but I am following the style that is used in the other exception classes. Where the property requestId is not defined either. See this file for reference https://github.com/stripe/stripe-php/blob/master/lib/Error/Base.php#L21

brandur:
@thecrogers Thanks for the pull!

Looks mostly good to me, but I'm not too crazy about the inconsistency of pulling `decline_code` out of `jsonBody` in the card error's constructor when other special fields are extracted in `handleApiError` and then injected in. What do you think about changing that around?

thecrogers:
@brandur Thanks for the quick response. 

That is what I had thought initially as well, but was unsure if changing the constructor was the best idea. If you are ok with changing the constructor for Card Error, I am good with that as well.

brandur:
Hey @thecrogers, actually hmm, that's a good point. I was thinking/hoping that our errors might be an "internal enough" interface that we could change it, but there's probably a good number of users who are initializing them in tests and the like.

I'm going to pull this in as is and add a comment that we should probably change it on the next major version bump.

brandur:
Actually, I added one more tweak in #281 to rename `StripeDeclineCode` to `DeclineCode`. I can see that an effort was made to follow the naming conventions of the other parameters in this class, but I think that those were given a "Stripe" prefix to disambiguate them (i.e. a Stripe "code" versus an HTTP "code"). Because a decline code is not ambiguous, it's safe to strip the prefix.

brandur:
Released as 3.20.0.

PR #287, opened 2016-08-24
Add ability to manipulate sources

wbronitsky:
r? @brandur

### DIFF
--- a/init.php
+++ b/init.php
@@ -60,6 +60,7 @@ require(dirname(__FILE__) . '/lib/Product.php');
 require(dirname(__FILE__) . '/lib/Recipient.php');
 require(dirname(__FILE__) . '/lib/Refund.php');
 require(dirname(__FILE__) . '/lib/SKU.php');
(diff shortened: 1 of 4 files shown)


brandur:
Hey @wbronitsky, looks good!

A couple things:

1. I think you're going to need to add this to `init.php`.
2. It's pretty unusual to add a new model without a corresponding test file with at least a little meat. Can you put one in? (Copying a basic template like coupon and adding a test for verify is enough.) I have a feeling that if we had one, we may have caught the bug in (1) in CI.

wbronitsky:
@brandur updated per code review. Please take another look when you can

brandur:
@wbronitsky Thanks!

Unfortunately Travis seems to be having some pretty serious trouble today, and I don't think it's likely that CI is going to be passing anytime soon. I've confirmed that the suite runs locally, and am pulling this in.

LGTM.

brandur:
Released as 3.21.0.

PR #294, opened 2016-09-13
Added custom application information to the User-Agent headers

olivierbellone:
r? @brandur 
cc @stripe/api-libraries 

This PR adds a `setAppInfo` method to the static `Stripe` class. The method takes three parameters:

- `$appName` (mandatory): the application's name
- `$appVersion` (optional): the application's version
- `$appUrl` (optional): the URL for the application's site

The method is meant to be used by plugin developers. After the method has been called, the headers sent with every request will be modified like this:

- `User-Agent` will have the additional information appended at the end, e.g. `Stripe/v1 PhpBindings/3.21.0` would become `Stripe/v1 PhpBindings/3.21.0 MyAwesomeApp/1.2.34 (https://myawesomeapp.example)`
- `X-Stripe-Client-User-Agent` will have an additional `application` key with the same string that's appended to `User-Agent`, e.g. `MyAwesomeApp/1.2.34 (https://myawesomeapp.example)`

For more context about the motivation behind this PR, see the DX-120 issue internally.

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -117,6 +117,51 @@ class ApiRequestor
         }
     }
 
+    private static function _formatAppInfo($appInfo)
+    {
+        if ($appInfo !== null) {
+            $string = $appInfo['name'];
+            if ($appInfo['version'] !== null) {
+                $string .= '/' . $appInfo['version'];
+            }
+            if ($appInfo['url'] !== null) {
+                $string .= ' (' . $appInfo['url'] . ')';
+            }
+            return $string;
+        } else {
+            return null;
+        }
+    }
+
+    private static function _defaultHeaders($apiKey)
+    {
+        $appInfo = Stripe::getAppInfo();
+
+        $uaString = 'Stripe/v1 PhpBindings/' . Stripe::VERSION;
+
+        $langVersion = phpversion();
+        $uname = php_uname();
+        $appInfo = Stripe::getAppInfo();
+        $ua = array(
+            'bindings_version' => Stripe::VERSION,
+            'lang' => 'php',
+            'lang_version' => $langVersion,
+            'publisher' => 'stripe',
+            'uname' => $uname,
+        );
+        if ($appInfo !== null) {
+            $uaString .= ' ' . self::_formatAppInfo($appInfo);
+            $ua['application'] = $appInfo;
+        }
+
+        $defaultHeaders = array(
+            'X-Stripe-Client-User-Agent' => json_encode($ua),
+            'User-Agent' => $uaString,
+            'Authorization' => 'Bearer ' . $apiKey,
+        );
+        return $defaultHeaders;
+    }
+
     private function _requestRaw($method, $url, $params, $headers)
     {
         $myApiKey = $this->_apiKey;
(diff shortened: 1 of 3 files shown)


lib/ApiRequestor.php
```diff
@@ -117,6 +117,31 @@ public function handleApiError($rbody, $rcode, $rheaders, $resp)
...
+        $ua = array(
+            'bindings_version' => Stripe::VERSION,
+            'lang' => 'php',
+            'lang_version' => $langVersion,
+            'publisher' => 'stripe',
+            'uname' => $uname,
+        );
+        if ($appInfo !== null) {
+            $uaString .= ' ' . $appInfo;
+            $ua['application'] = $appInfo;
```
floatingLomas: Is it possible to do nested values in here - i.e., leave the 'object' as an object?  It might be easier to grok [0] for analytics etc. if it didn't require parsing the string.

[0] _This might actually be the first time I've typed `grok` on the internet._

olivierbellone: Excellent point. It definitely should be possible, I'll give it a try!

lib/ApiRequestor.php
```diff
@@ -134,20 +179,7 @@ private function _requestRaw($method, $url, $params, $headers)
...
-            'lang_version' => $langVersion,
-            'publisher' => 'stripe',
-            'uname' => $uname,
-        );
-        $defaultHeaders = array(
-            'X-Stripe-Client-User-Agent' => json_encode($ua),
-            'User-Agent' => 'Stripe/v1 PhpBindings/' . Stripe::VERSION,
-            'Authorization' => 'Bearer ' . $myApiKey,
-        );
+        $defaultHeaders = $this->_defaultHeaders($myApiKey);
```
brandur: Nice. Way cleaner here now.

tests/ApiRequestorTest.php
```diff
@@ -40,4 +40,29 @@ public function testHttpClientInjection()
         $injectedCurl = $method->invoke(new ApiRequestor());
         $this->assertSame($injectedCurl, $curl);
     }
+
+    public function testDefaultHeaders()
+    {
+        $reflector = new \ReflectionClass('Stripe\\ApiRequestor');
+        $method = $reflector->getMethod('_defaultHeaders');
+        $method->setAccessible(true);
```
brandur: Oh man, TBH, I find that using reflection to access internal methods is testing practice that's kind of terrible on deep level, but I see that you're just following the existing convention here :$ Hopefully we can refactor this a bit later to be more palatable.

brandur:
@olivierbellone Awesome! Thanks again for picking this up and running with it; it should be hugely helpful in tracking down future compatibility problems.

LGTM.

brandur:
Released as 3.22.0.

snez:
Hey guys, could we please get a port of this for the older 1.x libraries? If a 1.19.0 is under consideration, could you please throw in 3D Secure as well based on https://github.com/snez/stripe-php/commit/15df5ebc06e5bf57e60d46b8ffc4e6f6d9552b36

Thanks!

brandur:
@snez Sorry for the answer that's sure to be disappointing, but 1.x is not maintained and we don't backport changes to it. Is there anything that we can do to help you upgrade? Being on such an ancient release is going to cause other trouble in the long run like not getting other more minor bug fixes, etc.

PR #298, opened 2016-09-16
Add SubscriptionItem object and tests

jxu577:
Add SubscriptionItem object that supports create, update, retrieval, list, and deletion.


kyleconroy:
👍

Issue #289, opened 2016-08-31
Setting CURLOPT_SSLVERSION explicitly causes it to fail (curl error 35).


BrunoDeBarros:
Working on a client's server today (I develop a self-hosted app that uses the Stripe SDK for payments), Stripe's SDK failed with the error: Unexpected error communicating with Stripe. If this problem persists, let us know at support@stripe.com. (Network error [errno 35]: Unsupported SSL protocol version)

They contacted Stripe and they were told to run [this code](https://gist.github.com/anonymous/506e7818d1a7d7f4d687c5436cbd8ac8), which outputted "TLS 1.2" correctly. So TLS 1.2 is supported properly.

The client was running Stripe 3.20.0 (3.21 only adds the Source stuff so it wouldn't have changed anything), so this is an issue with the latest code. After digging into it a bit, I figured out that the problem lay with [these lines](https://github.com/stripe/stripe-php/blob/master/lib/HttpClient/CurlClient.php#L171-L183). This server has CURL_SSLVERSION_TLSv1_2 defined (PHP 5.6), and OpenSSL is current enough to pass that `if`, so the code was running `$opts[CURLOPT_SSLVERSION] = CURL_SSLVERSION_TLSv1_2;`, which for some reason causes the problem, because changing it to `CURL_SSLVERSION_TLSv1 ` makes the problem go away. Somehow, explicitly requiring 1.2 is a problem while just asking for any TLS is fine. Removing all the lines linked to above (i.e. not setting CURLOPT_SSLVERSION at all) also resolves the problem. It's also worth noting that PHP's docs themselves recommend not setting CURLOPT_SSLVERSION at all (http://php.net/manual/en/function.curl-setopt.php "Note: Your best bet is to not set this and let it use the default."). So it could be that the best thing to do is to just remove it.

This is not a Stripe-specific issue because in the TLS-checking gist linked above, if I add the line `curl_setopt($c, CURLOPT_SSLVERSION, CURL_SSLVERSION_TLSv1_2);`, it makes it fail. So it's just something with curl itself on this particular server configuration. But since Stripe does depend on it working correctly, I thought it might be useful to report this, just in case you want to debug and find a workaround for it to make sure others don't have this problem. 

I have FTP access to the server and the client's permission to work on debugging this issue so I'm happy to look at whatever is needed and provide whatever information is required.

[...]

PR #299, opened 2016-09-21
Better TLS (hopefully)

olivierbellone:
r? @brandur 
cc @stripe/api-libraries @remistr 

This PR aims to solve the various TLS issues that some of our users have encountered, usually caused by weird interactions between software versions (most notably curl and its underlying SSL library, typically but not always OpenSSL) and the way we configure TLS in the library.

To summarize:

- the library does not set `CURLOPT_SSLVERSION` at all anymore. From the [PHP docs](https://secure.php.net/manual/en/function.curl-setopt.php): "Your best bet is to not set this and let it use the default. "

- users who need to set this option (on some older RedHat / CentOS versions, the server is capable of using TLS 1.2 but will use TLS 1.0 unless `CURLOPT_SSLVERSION` is explicitly set to either `CURL_SSLVERSION_TLSv1` or `CURL_SSLVERSION_TLSv1_2`) can now do so by passing the option themselves (see README)

- curl constants definitions have been moved out of the `request()` method so that we don't check if the constants are already defined for every request

- the `X-Stripe-Client-User-Agent` header now includes curl version and the SSL library name and version

This update will unfortunately break some users' integrations (for those cases where the library happens to work now, but will require manually setting `CURLOPT_SSLVERSION` after the update), and so should be released with a major version bump.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,17 @@ $curl = new \Stripe\HttpClient\CurlClient(array(CURLOPT_PROXY => 'proxy.local:80
 
 Alternately, a callable can be passed to the CurlClient constructor that returns the above array based on request inputs. See `testDefaultOptions()` in `tests/CurlClientTest.php` for an example of this behavior. Note that the callable is called at the beginning of every API request, before the request is sent.
 
+### SSL / TLS compatibility issues
+
+Stripe's API now requires that [all connections use TLS 1.2](https://stripe.com/blog/upgrading-tls). Some systems (most notably some older CentOS and RHEL versions) are capable of using TLS 1.2 but will use TLS 1.0 or 1.1 by default. In this case, you'd get an `invalid_request_error` with the following error message: "Stripe no longer supports API requests made with TLS 1.0. Please initiate HTTPS connections with TLS 1.2 or later. You can learn more about this at https://stripe.com/blog/upgrading-tls.".
+
+The recommended course of action is to [upgrade your cURL and OpenSSL packages](https://support.stripe.com/questions/how-do-i-upgrade-my-stripe-integration-from-tls-1-0-to-tls-1-2#php) so that TLS 1.2 is used by default, but if that is not possible, you might be able to solve the issue by setting the `CURLOPT_SSLVERSION` option to either `CURL_SSLVERSION_TLSv1` or `CURL_SSLVERSION_TLSv1_2`:
+
+```php
+$curl = new \Stripe\HttpClient\CurlClient(array(CURLOPT_SSLVERSION => CURL_SSLVERSION_TLSv1));
+\Stripe\ApiRequestor::setHttpClient($curl);
+```
+
 ## Development
 
 Install dependencies:
@@ -127,3 +138,17 @@ Or to run an individual test file:
 ```bash
 ./vendor/bin/phpunit tests/UtilTest.php
 ```
+
+## Attention plugin developers
+
+Are you writing a plugin that integrates Stripe and embeds our library? Then please use the `setAppInfo` function to identify your plugin. For example:
+
+```php
+\Stripe\Stripe::setAppInfo("MyAwesomePlugin", "1.2.34", "https://myawesomeplugin.info");
+```
+
+The method should be called once, before any request is sent to the API. The second and third parameters are optional.
+
+### SSL / TLS configuration option
+
+See the "SSL / TLS compatibility issues" paragraph above for full context. If you want to ensure that your plugin can be used on all systems, you should add a configuration option to let your users choose between different values for `CURLOPT_SSLVERSION`: none (default), `CURL_SSLVERSION_TLSv1` and `CURL_SSLVERSION_TLSv1_2`.
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -141,6 +141,7 @@ class ApiRequestor
 
         $langVersion = phpversion();
         $uname = php_uname();
+        $curlVersion = curl_version();
         $appInfo = Stripe::getAppInfo();
         $ua = array(
             'bindings_version' => Stripe::VERSION,
@@ -148,6 +149,8 @@ class ApiRequestor
             'lang_version' => $langVersion,
             'publisher' => 'stripe',
             'uname' => $uname,
+            'httplib' => 'curl ' . $curlVersion['version'],
+            'ssllib' => $curlVersion['ssl_version'],
         );
         if ($appInfo !== null) {
             $uaString .= ' ' . self::_formatAppInfo($appInfo);
--- a/lib/HttpClient/CurlClient.php
+++ b/lib/HttpClient/CurlClient.php
@@ -6,6 +6,22 @@ use Stripe\Stripe;
 use Stripe\Error;
 use Stripe\Util;
 
+// cURL constants are not defined in PHP < 5.5
+
+// @codingStandardsIgnoreStart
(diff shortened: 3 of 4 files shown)


lib/HttpClient/CurlClient.php
```diff
@@ -6,6 +6,22 @@
...
+
+// Note the values 1 and 6 come from their position in the enum that
+// defines them in cURL's source code.
+if (!defined('CURL_SSLVERSION_TLSv1')) {
+    define('CURL_SSLVERSION_TLSv1', 1);
+}
+if (!defined('CURL_SSLVERSION_TLSv1_2')) {
+    define('CURL_SSLVERSION_TLSv1_2', 6);
+}
+// @codingStandardsIgnoreEnd
```
brandur: In retrospect, this definitely fits best up in the header of this file. Nice!

README.md
```diff
@@ -106,6 +106,17 @@ $curl = new \Stripe\HttpClient\CurlClient(array(CURLOPT_PROXY => 'proxy.local:80
...
+
+Stripe's API now requires that [all connections use TLS 1.2](https://stripe.com/blog/upgrading-tls). Some systems (most notably some older CentOS and RHEL versions) are capable of using TLS 1.2 but will use TLS 1.0 or 1.1 by default.
+
+The recommended course of action is to [upgrade your cURL and OpenSSL packages](https://support.stripe.com/questions/how-do-i-upgrade-my-stripe-integration-from-tls-1-0-to-tls-1-2#php) so that TLS 1.2 is used by default, but if that is not possible, you might be able to solve the issue by setting the `CURLOPT_SSLVERSION` option to either `CURL_SSLVERSION_TLSv1` or `CURL_SSLVERSION_TLSv1_2`:
+
+```php
+$curl = new \Stripe\HttpClient\CurlClient(array(CURLOPT_SSLVERSION => CURL_SSLVERSION_TLSv1));
+\Stripe\ApiRequestor::setHttpClient($curl);
+```
+
```
remistr: Minor comment but I'd recommend that you add the exact error message people might get on this page. This is helpful for someone looking for the exact error either on the page or via Google.

lib/HttpClient/CurlClient.php
```diff
@@ -6,6 +6,22 @@
 use Stripe\Error;
 use Stripe\Util;
 
+// cURL constants are not defined in PHP < 5.5
+
+// @codingStandardsIgnoreStart
+// PSR2 requires all constants be upper case. Sadly, the CURL_SSLVERSION
+// constants to not abide by those rules.
```
remistr: minor nit "constants to not" => "constants do not"

lib/ApiRequestor.php
```diff
@@ -141,13 +141,16 @@ private static function _defaultHeaders($apiKey)
 
         $langVersion = phpversion();
         $uname = php_uname();
+        $curlVersion = curl_version();
```
remistr: have we confirmed that calling this has a minimal impact? We've discussed this before in private but can't remember seeing clear results

remistr:
Added a few minor comments here to address first but otherwise I think we're ready to ship. Since this is a major change, it might be worth delaying it a bit more and trying to think of anything we've wanted to change or fix for some time but couldn't because it would require a major version release

pderksen:
Not sure if this helps or is too late...

Though they don't use this Stripe PHP library, the WooCommerce Stripe payment gateway added this:
https://github.com/woothemes/woocommerce-gateway-stripe/commit/877dcabf5c60ef6e41f770ecf654274fa1fc0ce5

remistr:
@pderksen this is the original code we had in our library too as far as I can tell (they even mention olivier's name directly in the comments).

Unfortunately, we've ended up in situations where some users have PHP built with one specific version of curl and then uses a completely different one in production. When that happens, forcing the constant sometimes doesn't work at all producing errors such as the one documented in https://github.com/stripe/stripe-php/issues/289.

This PR is a different attempt to avoid this issue. We now rely on curl doing the right thing by default and then provide an option for users to explicitly force a different constant otherwise if they need to.

brandur:
> Since this is a major change, it might be worth delaying it a bit more and trying to think of anything we've wanted to change or fix for some time but couldn't because it would require a major version release

@remistr Given that major version releases tend to be slightly intimidating as is, what do you think about trying to keep this one fairly minimal? I'm a little worried that if there are too many other changes they might present a bit of a barrier to upgrade.

remistr:
@brandur well that seems the opposite of what a major release should be no? Otherwise we'll keep not changing things we want to fix because it's too minor for a major release and at the same time have to release major releases every time something less minor changes.

I can't remember a specific thing we wanted to fix but couldn't recently. I'm just trying to advocate for taking this a bit slower than usual and trying to make sure we're not missing a quick win we could get and avoid a 5.X in a few weeks.

lib/ApiRequestor.php
olivierbellone: I ran some basic benchmarking, and it takes approximately 1.3 seconds to run 1 million calls to `curl_version` on my laptop. Given that ~all of our users will never send more than 100 req/s, I think we're fine.

README.md
olivierbellone: Great point. Done!

lib/HttpClient/CurlClient.php
olivierbellone: Fixed.

olivierbellone:
Hey @pderksen, thanks. As @remistr pointed out, this is how we previously set the option in our own library, and I'm the one who suggested that they do the same [here](https://github.com/woothemes/woocommerce-gateway-stripe/issues/45#issuecomment-238919177). I've left another comment to let them know about our latest findings and this new update.

brandur:
> I can't remember a specific thing we wanted to fix but couldn't recently. I'm just trying to advocate for taking this a bit slower than usual and trying to make sure we're not missing a quick win we could get and avoid a 5.X in a few weeks.

Well I should point out that we're not burning through major releases at an incredible rate (the last was almost a year and a half ago), so it doesn't seem like we'd be starting a 5.x line anytime soon. But cool, it would be nice to get some of the people on #289 fixed up, but aside from that I guess we're not in a particular rush. Let me know when you think it would be wise to move forward on this one.

remistr:
@brandur Then I think we're okay merging and releasing this!

olivierbellone:
I added a paragraph at the end of the README for plugin devs, to let them know about `setAppInfo` (this should have been in #294) and that they should plan for a configuration option to set `CURLOPT_SSLVERSION` if they want to ensure that their plugin can be used on all systems.

Feel free to tweak the wording or just rewrite the whole thing!

README.md
```diff
@@ -106,6 +106,17 @@ $curl = new \Stripe\HttpClient\CurlClient(array(CURLOPT_PROXY => 'proxy.local:80
 
 Alternately, a callable can be passed to the CurlClient constructor that returns the above array based on request inputs. See `testDefaultOptions()` in `tests/CurlClientTest.php` for an example of this behavior. Note that the callable is called at the beginning of every API request, before the request is sent.
 
+### SSL / TLS compatibiliy issues
```
brandur: Ugh, pretty minor, but I just noticed this typo on "compatibility". Mind fixing that?

olivierbellone: Done!

olivierbellone:
@brandur Sorry for the delay, GitHub failed to notify me for some reason. Typo is fixed, let's :ship: this!

brandur:
Released as 4.0.0. Fingers crossed :)

PR #303, opened 2016-10-14
Support 403 status codes

jflinter:
r? @brandur

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -76,7 +76,7 @@ class ApiRequestor
      * @throws Error\InvalidRequest if the error is caused by the user.
      * @throws Error\Authentication if the error is caused by a lack of
      *    permissions.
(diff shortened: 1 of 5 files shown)


lib/ApiRequestor.php
```diff
@@ -110,6 +112,8 @@ public function handleApiError($rbody, $rcode, $rheaders, $resp)
                 throw new Error\Authentication($msg, $rcode, $rbody, $resp, $rheaders);
             case 402:
                 throw new Error\Card($msg, $param, $code, $rcode, $rbody, $resp, $rheaders);
+            case 403:
+                throw new Error\Permissions($msg, $rcode, $rbody, $resp, $rheaders);
```
brandur: Definitely a nit, but it looks like many of these other errors are named in the singular form. Any particular reason to prefer `Permissions` over `Permission` here?

jflinter:
re-r? @brandur

PR #305, opened 2016-10-17
Fix TransferReversal to have the proper class

remistr:
Before that fix, accessing a TransferReversal gave you a StripeObject.

r? @brandur


brandur:
Thanks!

brandur:
Released as 4.0.1.

PR #309, opened 2016-11-02
Added a template for GitHub issues

ob-stripe:
r? @brandur

As we discussed, this adds a default "template" for GitHub issues, though in our case it's less of an actual template and more of a notice so people will route their issues to the appropriate channel.


brandur-stripe:
Awesome. Thanks again @ob-stripe!

brandur:
Awesome. Thanks again @ob-stripe!

PR #310, opened 2016-11-21
Force cURL to use TLS 1.2 when TLS 1.2 is supported but not used by default

robertotremonti:
Since July 2016, Stripe requires TLS 1.2 for all connections.
I suggest to force cURL to use TLS 1.2 on those servers where TLS 1.2 is supported but not used by default.


ob-stripe:
@robertotremonti Thanks for the PR, but we'll have to decline it. We've actually went back and forth on this issue, as we previously did exactly what you suggested, but it ended up having the opposite of the intended effect on some systems.

We eventually decided to do what the PHP docs recommend and not set the `CURLOPT_SSLVERSION` at all, which should result in the highest available TLS version being used on most systems, and provide a way for users on those (rare) systems where that's not the case to pass the options themselves (cf. [README](https://github.com/stripe/stripe-php#ssl--tls-compatibility-issues)).

For more information, see the following PRs: #216, #277, #299 and their related issues.

PR #311, opened 2016-11-21
Added 3DS retrieve endpoint

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @bct 

Adds a method for retrieving 3DS auths.


brandur:
Ugh, it looks like our master build is broken (which has caused the problem here). This looks good though.

brandur:
Released as 4.2.0.

PR #313, opened 2016-12-05
Fixed Upgrading TLS blog post url

smottt:
Removed the trailing dot in the Upgrading TLS blog post url.


ob-stripe:
@smottt Nice catch, thanks!

PR #320, opened 2017-01-20
Remove subscriptions annotation from customer.php

jack-stripe:
r? @brandur-stripe


brandur:
Looks generally good. Thanks!

PR #324, opened 2017-02-19
Fix: Uncought exception in ApiRequestor.php

ankurk91:
Add namespace so that Exception can be catched in [_interpretResponse](https://github.com/stripe/stripe-php/blob/master/lib/ApiRequestor.php#L259) method of this class.

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -2,6 +2,8 @@
 
 namespace Stripe;
 
+use Exception;
+
 /**
  * Class ApiRequestor
  *


brandur-stripe:
Thanks @ankurk91! Could you describe the problem that you're seeing right now? (Like a backtrace and anything else.) Thanks!

ankurk91:
Hi, @brahn-stripe 
I am talking about this function -
```php
 private function _interpretResponse($rbody, $rcode, $rheaders)
    {
        try {
            $resp = json_decode($rbody, true);
        } catch (Exception $e) {
            $msg = "Invalid response body from API: $rbody "
              . "(HTTP response code was $rcode)";
            throw new Error\Api($msg, $rcode, $rbody);
        }
        if ($rcode < 200 || $rcode >= 300) {
            $this->handleApiError($rbody, $rcode, $rheaders, $resp);
        }
        return $resp;
    }

```
As you can see, we are trying to access ```Exception``` class which does not exist in ```Stripe``` namespace. The ```catch``` will never happen, another solutions is -
```
 } catch (\Exception $e) {
```
notice the slash ```\``` before accessing ```Exception``` class.
A better way to detect malformed JSON is [json_last_error](http://php.net/manual/en/function.json-last-error.php)

brandur-stripe:
Ah, I see. Thanks! Crazy that this sort of problem could have lingered for so long.

I'd have a slight preference for just putting the `\` in front of `Exception` to suggest the root namespace as you suggest above.

@bkrausz-stripe Would you mind also giving this one a quick sanity check?

jlomas-stripe:
Isn't `Exception` a global PHP class though?

bkrausz-stripe:
While you're correct that this should be `\Exception` to catch exceptions, `json_decode` doesn't actually raise on decoding failure, so that `catch` can never be hit regardless.

We should change this to a `$resp === null` check which, given that we'll never actually return `null` as our response body, is the correct way to tell if `json_decode` failed.

ankurk91:
I won't mind if you guys close this pull request and implement best solution here.

PR #325, opened 2017-02-24
Properly check if json_decode failed

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @bkrausz-stripe @ankurk91

Implements the fix suggested by @bkrausz-stripe [here](https://github.com/stripe/stripe-php/pull/324#issuecomment-281799778).

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -254,11 +254,11 @@ class ApiRequestor
 
     private function _interpretResponse($rbody, $rcode, $rheaders)
     {
-        try {
-            $resp = json_decode($rbody, true);
-        } catch (Exception $e) {
+        $resp = json_decode($rbody, true);
+        $jsonError = json_last_error();
+        if ($resp === null && $jsonError !== JSON_ERROR_NONE) {
             $msg = "Invalid response body from API: $rbody "
-              . "(HTTP response code was $rcode)";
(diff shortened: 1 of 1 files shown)


brandur-stripe:
This looks great! Unfortunately the build is failing, but it looks unrelated ... I just started a rebuild on master to find out.

lib/ApiRequestor.php
```diff
@@ -254,9 +254,8 @@ private function _processResourceParam($resource, $hasCurlFile)
 
     private function _interpretResponse($rbody, $rcode, $rheaders)
     {
-        try {
-            $resp = json_decode($rbody, true);
-        } catch (Exception $e) {
+        $resp = json_decode($rbody, true);
+        if ($resp === null) {
```
remi-stripe: We should use [json-last-error()](http://www.php.net/manual/en/function.json-last-error.php) and/or [json-last-error-msg](http://php.net/manual/en/function.json-last-error-msg.php) at least in our logs to help our users debug. Flagging here instead of in private to avoid merging until then.

ob-stripe: `json_last_error_msg()` is only available in PHP >= 5.5, but I updated the PR to use `json_last_error()`.

brandur-stripe:
Nice. Master is broken unfortunately, so merging this as is :/

brandur-stripe:
Released as 4.4.1.

PR #326, opened 2017-02-24
Make StripeObject not switch NULLs to empty strings when it shouldn't

jlomas-stripe:
So this is a bit of a weird one.

Basically, if I try to add an additional `additional_owners` entry:

```
$account = \Stripe\Account::retrieve($account_id);

$count = count($account->legal_entity->additional_owners);
$account->legal_entity->additional_owners[$count] = array(
  'first_name' => 'Frank',
  'last_name' => 'Lastnameson'
);

$account->save();
```

I get this:

```Fatal error: Uncaught Stripe\Error\InvalidRequest: You passed an empty string for 'legal_entity[additional_owners][1][dob]'. We assume empty values are an attempt to unset a parameter; however 'legal_entity[additional_owners][1][dob]' cannot be unset. You should remove 'legal_entity[additional_owners][1][dob]' from your request or supply a non-empty value. in /Users/jlomas/fiddling/php/vendor/stripe/stripe-php/lib/ApiRequestor.php:110 from API request 'req_ABCDDc8cQM5NDy'```

Because what actually gets assembled/POSTed to Stripe is this:

```{
  "legal_entity": {
    "additional_owners": {
      "1": {
        "address": "",
        "dob": "",
        "verification": "",
        "first_name": "Frank",
        "last_name": "Lastnameson"
      }
    }
  }
}
```

...because `address`, `dob` and `verification` are considered [updatable (nullable) nested properties by the library](https://github.com/stripe/stripe-php/blob/master/lib/StripeObject.php#L29-L35), but are not considered as such by the API.

I'm not sure if these property names collide with something else in the API that _is_ nullable though - thus why it is a weird one.

r? @brandur-stripe
cc @stripe/api-libraries

### DIFF
--- a/lib/StripeObject.php
+++ b/lib/StripeObject.php
@@ -22,17 +22,33 @@ class StripeObject implements ArrayAccess, JsonSerializable
      *    the parent class's URL (e.g. metadata).
      */
     public static $nestedUpdatableAttributes;
+    /**
+     * @var Util\Set Attributes that are nested but cannot be unset via
+     *    an empty string.
+     */
+    public static $nonNullableAttributes;
 
     public static function init()
     {
         self::$permanentAttributes = new Util\Set(array('_opts', 'id'));
         self::$nestedUpdatableAttributes = new Util\Set(array(
-            'metadata', 'legal_entity', 'address', 'dob', 'transfer_schedule', 'verification',
-            'tos_acceptance', 'personal_address',
+            'metadata', 'legal_entity',
+            'transfer_schedule', 'tos_acceptance',
+
+            // These all can't be unset via an empty string.
+            'dob', 'verification',
+            'address',
+            'personal_address',
+
             // will make the array into an AttachedObject: weird, but works for now
             'additional_owners', 0, 1, 2, 3, 4, // Max 3, but leave the 4th so errors work properly
             'inventory',
             'owner',
+        ));
+        self::$nonNullableAttributes = new Util\Set(array(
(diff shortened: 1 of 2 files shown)


brandur-stripe:
@jlomas-stripe Nice work! Thanks a lot for tackling this one, and sorry about the really late reply.

So I see that you modeled this after what was already there in `nestedUpdatableAttributes`. It probably doesn't make a huge practical different, but the fragility of this implementation concerns me a bit in that we're just matching off field names and nothing else (not on you, the original implementation was a little misguided :/).

In stripe-ruby, what I did eventually is assume that `additional_owners` is going to be perpetually a one-off mistake and implemented a [special serialization method on `Account` that handles it specifically](https://github.com/stripe/stripe-ruby/blob/8f817b83653f75560029d4a3d52e029674a5f3a3/lib/stripe/account.rb#L105,L131) so that we can maintain the integrity of our base class and shared modules.

What do you think of trying an alternative like that instead?

jlomas-stripe:
@brandur-stripe No worries.  :)

I think the alternative makes a lot of sense, since this is a weird case.  I'll take a run at that in the next little while.

Thanks!

jlomas-stripe:
@brandur-stripe This issue is actually a red herring caused by #329.  There's no problem adding an additional owner in the right place, but `count($account->legal_entity->additional_owners);` doesn't return the right value - it always returns 1 - so it's viewed as 'Update object at index 1' instead of 'Add an object at index $count'.

PR #327, opened 2017-02-27
Fix for updating sources' owner attributes

ob-stripe:
r? @brandur-stripe
cc @stripe/api-libraries

Adds `owner` to the list of nested updatable attributes to allow for updating sources' `owner` attributes: https://stripe.com/docs/api#update_source-owner


brandur-stripe:
That field list is so hacky :$

LGTM. Thanks!

brandur-stripe:
Released as 4.4.2.

PR #329, opened 2017-03-01
Adding a count() implementation for StripeObjects

gh-stripe:
This PR adds the Countable interface as well as a `count()` implementation to `StripeObjects.php`. Right now `count()` on a StripeObject property always returned `1` due to the way we handle values in it.

There was at least one example in our documentation that was failing due to this as we require to count the number of `legal_entity.additional_owners` in a managed account before adding a new one to the right index.

r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/lib/StripeObject.php
+++ b/lib/StripeObject.php
@@ -3,6 +3,7 @@
 namespace Stripe;
 
 use ArrayAccess;
+use Countable;
 use InvalidArgumentException;
 
 /**
@@ -10,7 +11,7 @@ use InvalidArgumentException;
  *
  * @package Stripe
  */
-class StripeObject implements ArrayAccess, JsonSerializable
+class StripeObject implements ArrayAccess, JsonSerializable, Countable
 {
     /**
      * @var Util\Set Attributes that should not be sent to the API because
@@ -288,6 +289,11 @@ class StripeObject implements ArrayAccess, JsonSerializable
         } else {
             return $this->_values;
         }
+    }
+
+    public function count()
+    {
+        return count($this->_values);
     }
 }
 
--- a/tests/AccountTest.php
+++ b/tests/AccountTest.php
@@ -295,6 +295,7 @@ class AccountTest extends TestCase
 
         $account->legal_entity->additional_owners[1] = array('first_name' => 'Jane');
         $account->save();
+        $this->assertSame(2, count($account->legal_entity->additional_owners));
         $this->assertSame('Jane', $account->legal_entity->additional_owners[1]->first_name);
     }
 }


brandur-stripe:
@gh-stripe Nice!! This seems like such an obviously good idea.

Just with respect to an edge case: what happens for a `StripeObject` that's not a list. For example, a single instance of `Charge`. I think I know the answer, but I just wanted to double-check that it makes sense.

gh-stripe:
@brandur-stripe That's a good question. I actually completely forgot to check how it would behave on different objects and types of properties. Right now it gives the number of properties for the first level of the object you pass. For example a random Stripe\Account returns 27.

The object where this is a bit confusing is for Stripe\Collections, returned from ::all() methods, where this gives back 4 and not the number of Stripe\Object in it. Although you can still get it if you do count($list->data); instead.

I'm not sure if we want to adapt the implementation for those or if this is the expected behavior?

brandur-stripe:
Cool, thanks!

> @brandur-stripe That's a good question. I actually completely forgot to check how it would behave on different objects and types of properties. Right now it gives the number of properties for the first level of the object you pass. For example a random Stripe\Account returns 27.

So I'm not aware enough of idiomatic PHP to know what a perfect answer is here, but I could see it either returning the number of fields (as it does already) or potentially throwing an exception if it can determine that there's no good reason to ever call count on the object (i.e. a non-list).

> The object where this is a bit confusing is for Stripe\Collections, returned from ::all() methods, where this gives back 4 (it just counts the top object properties object, data, has_more and url) and not the number of Stripe\Object in it. Although you can still get it if you do count($list->data); instead.

What do you think about detecting a list object and returning the correct count of data in it, and to throw an error in other cases where it doesn't make sense?

One last thing: could we get some tests right on `tests/StripeObjectTest.php` (i.e. as opposed to in account) so that we can verify the expectations for the change directly?

Also cc @bkrausz-stripe in case he has opinions on how this should behave.

bkrausz-stripe:
It depends on how much we want to treat our objects as Arrays. If we want them to be super Array-y, then count() would return the number of keys, regardless of if it's a hash or array.

I'm actually a bit disinclined to add more magic to our keys. If the only issue here is `additional_owners` updating, we should fix that in particular. There's been a lot of talk about making `additional_owners` a subresource to avoid the lurky array-update semantics.

brandur-stripe:
> I'm actually a bit disinclined to add more magic to our keys. If the only issue here is additional_owners updating, we should fix that in particular. There's been a lot of talk about making additional_owners a subresource to avoid the lurky array-update semantics.

+1 to this being the right solution. There's been talk of retooling `additional_owners` for quite a while though, and so far it hasn't materialized, so we should probably patch this in some other way for today.

@gh-stripe Would it be possible to modify this to just address the `additional_owners` edge case in particular?

jlomas-stripe:
@gh-stripe @brandur-stripe I literally just landed on this while working on #326.

Could we move this from `StripeObject` to `AttachedObject` and solve the problem?

jlomas-stripe:
Also, I totally agree with wanting to solve this particular (edge) case in a specific way - i.e., properly implementing an `AdditionalOwner` object (and maybe a `LegalEntity` object to contain it within the `Account` object).

As it stands now, we've already ['hacked' StripeObject to work as an array](https://github.com/stripe/stripe-php/blob/master/lib/StripeObject.php#L33-L34) so it might be worth just making that hack 'complete' (by merging this) so this can work the way folks expect it to now, and then doing it 'right' when we have the time/resources.

Anyways, just a thought.

ob-stripe:
Superseded by #350, closing this one.

PR #331, opened 2017-03-17
Adds support for detaching sources from customers

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @will-stripe

Adds support for detaching sources from customers.

I changed the visibility of `_validateParams` so that derived classes can use it in order to be consistent with the implementation of `_delete` in `ApiResource` (though the API request to detach sources from customers does not expect any parameters).

### DIFF
--- a/lib/Source.php
+++ b/lib/Source.php
@@ -66,6 +66,41 @@ class Source extends ApiResource
 
     /**
      * @param array|null $params
+     * @param array|string|null $opts
+     *
+     * @return Source The deleted source.
+     */
+    public function delete($params = null, $options = null)
+    {
+        self::_validateParams($params);
+
+        $id = $this['id'];
+        if (!$id) {
+            $class = get_class($this);
+            $msg = "Could not determine which URL to request: $class instance "
+             . "has invalid ID: $id";
+            throw new Error\InvalidRequest($msg, null);
+        }
+
+        if ($this['customer']) {
+            $base = Customer::classUrl();
+            $parentExtn = urlencode(Util\Util::utf8($this['customer']));
+            $extn = urlencode(Util\Util::utf8($id));
+            $url = "$base/$parentExtn/sources/$extn";
+
+            list($response, $opts) = $this->_request('delete', $url, $params, $options);
+            $this->refreshFrom($response, $opts);
+            return $this;
+        } else {
+            $message = "Source objects cannot be deleted, they can only be "
+               . "detached from customer objects. This source object does not "
+               . "appear to be currently attached to a customer object.";
+            throw new Error\Api($message);
+        }
+    }
+
+    /**
+     * @param array|null $params
      * @param array|string|null $options
      *
      * @return BankAccount The verified bank account.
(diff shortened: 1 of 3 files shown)


lib/Source.php
```diff
@@ -66,6 +66,41 @@ public function save($opts = null)
...
+        if (!$id) {
+            $class = get_class($this);
+            $msg = "Could not determine which URL to request: $class instance "
+             . "has invalid ID: $id";
+            throw new Error\InvalidRequest($msg, null);
+        }
+
+        if ($this['customer']) {
+            $base = Customer::classUrl();
+            $parentExtn = urlencode(Util\Util::utf8($this['customer']));
```
brandur-stripe: Thanks for tracking and following all these crazy localized naming conventions :)

lib/Source.php
```diff
@@ -66,6 +66,41 @@ public function save($opts = null)
...
+    {
+        self::_validateParams($params);
+
+        $id = $this['id'];
+        if (!$id) {
+            $class = get_class($this);
+            $msg = "Could not determine which URL to request: $class instance "
+             . "has invalid ID: $id";
+            throw new Error\InvalidRequest($msg, null);
+        }
```
brandur-stripe: Just out of curiosity, what exactly is this block protecting against? Just an uninitialized source object?

brandur-stripe:
Added one question above, but otherwise looks good to me. Thanks for all these patches @ob-stripe!

lib/Source.php
ob-stripe: Yeah, I guess -- I mostly just copied the existing logic when deleting API resources: [`ApiResource::_delete`](https://github.com/stripe/stripe-php/blob/11c6041f7aea18905c121e46adf09dc4427cd14a/lib/ApiResource.php#L194) calls [`instanceUrl`](https://github.com/stripe/stripe-php/blob/11c6041f7aea18905c121e46adf09dc4427cd14a/lib/ApiResource.php#L91), which in turn calls [`resourceUrl`](https://github.com/stripe/stripe-php/blob/11c6041f7aea18905c121e46adf09dc4427cd14a/lib/ApiResource.php#L74) which has [this block](https://github.com/stripe/stripe-php/blob/11c6041f7aea18905c121e46adf09dc4427cd14a/lib/ApiResource.php#L76-L81).

For sources, we can't use the existing method because sources behave differently than existing customer payment sources (ugh naming), i.e. cards & bank accounts:
- their normal instance URL (for retrieve & update requests) is `/v1/sources/src_foo`
- sources cannot be deleted, but when attached to a customer, they can be detached by sending a `DELETE` request to `/v1/customers/cus_bar/sources/src_foo`.

If you want, I can refactor to avoid duplicating the code, or just get rid of the check entirely as it doesn't seem terriblye useful.

brandur-stripe: Thanks!

> If you want, I can refactor to avoid duplicating the code, or just get rid of the check entirely as it doesn't seem terriblye useful.

I'd say that it's okay given that it was probably added at some point for some weird error condition. Let's just merge as is.

brandur-stripe:
Released as 4.5.0.

PR #332, opened 2017-03-20
Remove dependency on curl as we shouldn't require it.

remi-stripe:
It's possible for users to use their own interface instead of curl to make API requests. We recently introduced a dependency to curl which is not expected.

r? @stripe/api-libraries

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -145,7 +145,16 @@ class ApiRequestor
 
         $langVersion = phpversion();
         $uname = php_uname();
-        $curlVersion = curl_version();
+
+        $httplib = 'unknown';
+        $ssllib = 'unknown';
+
+        if (function_exists('curl_version')) {
+            $curlVersion = curl_version();
+            $httplib = 'curl ' . $curlVersion['version'];
+            $ssllib = $curlVersion['ssl_version'];
+        }
+
         $appInfo = Stripe::getAppInfo();
         $ua = array(
             'bindings_version' => Stripe::VERSION,
@@ -153,8 +162,8 @@ class ApiRequestor
             'lang_version' => $langVersion,
             'publisher' => 'stripe',
             'uname' => $uname,
-            'httplib' => 'curl ' . $curlVersion['version'],
-            'ssllib' => $curlVersion['ssl_version'],
+            'httplib' => $httplib,
+            'ssllib' => $ssllib,
         );
         if ($appInfo !== null) {
             $uaString .= ' ' . self::_formatAppInfo($appInfo);


lib/ApiRequestor.php
```diff
@@ -145,16 +145,26 @@ private static function _defaultHeaders($apiKey)
 
         $langVersion = phpversion();
         $uname = php_uname();
-        $curlVersion = curl_version();
+
+        $httplib = 'curl not installed';
+        $ssllib = 'unknown';
+
+        if (function_exists('curl_version'))
+        {
```
brandur-stripe: Nit, but I think most of the conditional curly braces (as opposed to the ones for function) are on the same line. Mind changing over to that here for consistency?

lib/ApiRequestor.php
```diff
@@ -145,16 +145,26 @@ private static function _defaultHeaders($apiKey)
 
         $langVersion = phpversion();
         $uname = php_uname();
-        $curlVersion = curl_version();
+
+        $httplib = 'curl not installed';
```
brandur-stripe: Should this also just be "unknown" like you have for `$ssllib`? Otherwise might look a little strange in our user agent header.

ob-stripe:
Aw, that's my bad.

That said, it makes little sense to send info about curl to the API if the client doesn't actually use curl (in the event that curl is installed but a non-curl client is used).

The values for `httplib` and `ssllib` should instead be provided by the client, but that would probably require modifying the interface :(

lib/ApiRequestor.php
remi-stripe: so `curl uknown` or just `unknown`? I was assuming that if we parsed the header we'd expect to see a `curl prefix`.

remi-stripe:
@ob-stripe I considered that and wonder if it was worth changing the interface. We collect this just for TLS really and not to action the info do it didn't seem worth it.

lib/ApiRequestor.php
brandur-stripe: Hm, if `curl_version` doesn't exist we pretty much have no idea what they'll be using right? For that reason I was leaning toward just `unknown`.

remi-stripe:
@brandur-stripe PTAL

brandur-stripe:
LGTM.

@ob-stripe Definitely agreed that it would be better if even custom clients could hook into this. Let me know if you can think of a reasonable way to allow that to happen.

ob-stripe:
Maybe the clients could provide `$httplib` + `$ssllib` public variables, and then `ApiRequestor` would check for these via `property-exists`?

This way, if the variables don't exist (as will most likely be the case in custom clients), we would simple omit the `httplib` + `ssllib` entries in the `X-Stripe-Client-User-Agent` hash.

remi-stripe:
I'm not sure I understand the benefits of this if we end up just defaulting to empty values and not really enforcing that implementation from users.

I'm happy to add something along those lines, but it feels unneeded. Especially as we're tracking those for the TLS deprecation and wouldn't glean a lot of details from the users.

It's likely that no one does this today beyond the user who reported the issue otherwise we'd have had more reports right?

remi-stripe:
@ob-stripe @brandur wdyt? Should we merge as is or do you want me to add something custom to `ClientInterface`?

brandur-stripe:
Lets get this in. @ob-stripe's idea here sounds pretty plausible to me, but I think it could easily come in on a separate PR without trouble.

LGTM.

brandur-stripe:
Released as 4.5.1.

PR #333, opened 2017-03-22
Backport setAppInfo to legacy version of the bindings

remi-stripe:
We have a lot of third-party plugins/platforms that use the legacy version. The main reason is to support old PHP installations. We still want them to explicitly showcase who they are so that we can better help their users.

This is a first attempt at bringing the functionality in for a potential release as 1.18.1.

I'm still not convinced we should release this and instead we should nudge plugin developers to use the latest and more reliable version but at least the functionality is available if we do need to merge it in the future.

As a note: tests on 1.18.0 are failing locally on a lot of files even without my changes making it hard to make the merge safe but at least my test works and I've confirmed on a request that it does the right thing.

@stripe/api-libraries  what do you think?

### DIFF
--- a/lib/Stripe/ApiRequestor.php
+++ b/lib/Stripe/ApiRequestor.php
@@ -156,6 +156,51 @@ class Stripe_ApiRequestor
     }
   }
 
+  private static function _formatAppInfo($appInfo)
+  {
+      if ($appInfo !== null) {
+          $string = $appInfo['name'];
+          if ($appInfo['version'] !== null) {
+              $string .= '/' . $appInfo['version'];
+          }
+          if ($appInfo['url'] !== null) {
+              $string .= ' (' . $appInfo['url'] . ')';
+          }
+          return $string;
+      } else {
+          return null;
+      }
+  }
+
+  private static function _defaultHeaders($apiKey)
+  {
+      $appInfo = Stripe::getAppInfo();
+
+      $uaString = 'Stripe/v1 PhpBindings/' . Stripe::VERSION;
+
+      $langVersion = phpversion();
+      $uname = php_uname();
+      $appInfo = Stripe::getAppInfo();
+      $ua = array(
+          'bindings_version' => Stripe::VERSION,
+          'lang' => 'php',
+          'lang_version' => $langVersion,
+          'publisher' => 'stripe',
+          'uname' => $uname,
+      );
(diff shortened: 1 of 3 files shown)


brandur-stripe:
Wow, nice Remi! Thanks for tackling this one.

So as mentioned previously, I'm pretty on the fence about this: I definitely don't want to get into the habit of backporting features or maintaining ancient releases, but if this is going to be pretty much a one time thing, then I guess it would make sense to make the effort.

> As a note: tests on 1.18.0 are failing locally on a lot of files even without my changes making it hard to make the merge safe but at least my test works and I've confirmed on a request that it does the right thing.

Yeah, given that the suite is dependent on live calls, I imagine there's probably quite a few things that would need to be updated to get that back to green. I'd suggest that we don't worry about it too much and just try to build some basic confidence that the patch works as expected through a couple manual tests.

BTW, you're probably aware, but we won't be able to pull this back into master. I think what we'd want to do is branch off of the 1.18.0 tag, target that new branch from this PR, and then ship a new release off of it.

remi-stripe:
> BTW, you're probably aware, but we won't be able to pull this back into master. I think what we'd want to do is branch off of the 1.18.0 tag, target that new branch from this PR, and then ship a new release off of it.

Yes I did checkout away from `v1.18.0` already. I wanted to PR to be available if needed. My opinion is that we should agree on the code but not release it for now. We'll release it if we encounter plugin developers that refuse to move to 4.X. We'd simply close the PR but know the code is here if needed (and potentially shared as a reference if devs want to pull this in locally).

What do you think?

brandur-stripe:
> Yes I did checkout away from v1.18.0 already. I wanted to PR to be available if needed.

Oh right! I just meant that I think we'd have to target a non-`master` branch with this PR. While you branched off the right release tag, you're still targeting the codebase of the latest version. (I think this is why you're seeing conflicting files above.)

> We'll release it if we encounter plugin developers that refuse to move to 4.X. We'd simply close the PR but know the code is here if needed (and potentially shared as a reference if devs want to pull this in locally).
>
> What do you think?

Yep, that works for me.

remi-stripe:
> Oh right! I just meant that I think we'd have to target a non-master branch with this PR. While you branched off the right release tag, you're still targeting the codebase of the latest version. (I think this is why you're seeing conflicting files above.)

PR #337, opened 2017-04-05
Peg the test suite's API version

brandur-stripe:
Pegs the test suite's API version so that we can upgrade it
independently of the account's implicit version. It's also to check how
much breaks when we move to a relatively modern version of the API.

See also https://github.com/stripe/stripe-java/pull/360 and https://github.com/stripe/stripe-python/pull/298.

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -25,6 +25,11 @@ class TestCase extends \PHPUnit_Framework_TestCase
     protected function setUp()
     {
         ApiRequestor::setHttpClient(HttpClient\CurlClient::instance());
+
+        // Peg the API version so that it can be varied independently of the
+        // one set on the test account.
+        Stripe::setApiVersion('2017-02-14');
+
         $this->mock = null;
         $this->call = 0;
     }


brandur-stripe:
I also noticed that stripe-node just sets `latest`. That might also be a plausible approach here. `latest` has the advantage of giving us automatic protection against bitrot, but it'll also be prone to break at random times on PRs issued by users or whomever.

brandur-stripe:
That was surprisingly easy.

@remi-stripe Thoughts on this approach?

remi-stripe:
That sounds okay though I wonder if we need to also start adding custom tests for backwards compatibility that we want to keep supporting? Not fully sure how this would look like though. It's likely that our test suite is not exhaustive enough to catch most of those anyway

brandur-stripe:
> That sounds okay though I wonder if we need to also start adding custom tests for backwards compatibility that we want to keep supporting? Not fully sure how this would look like though. It's likely that our test suite is not exhaustive enough to catch most of those anyway

Yeah, I think that kind of thing might be a good idea. It should be relatively easy to override the API version on a per-test case basis.

Bringing this in for now to prepare for the payouts/transfers PR.

PR #339, opened 2017-04-07
Remove raw PANs from our test suite

remi-stripe:
This PR aims to use the new test tokens such as `tok_visa` instead or hardcoding raw PANs in the test suite.

There's one last PAN we can't remove for disputes as it hasn't been released yet.

r? @brandur-stripe

### DIFF
--- a/lib/ApiResource.php
+++ b/lib/ApiResource.php
@@ -99,8 +99,7 @@ abstract class ApiResource extends StripeObject
             $message = "You must pass an array as the first argument to Stripe API "
                . "method calls.  (HINT: an example call to create a charge "
                . "would be: \"Stripe\\Charge::create(array('amount' => 100, "
-               . "'currency' => 'usd', 'card' => array('number' => "
-               . "4242424242424242, 'exp_month' => 5, 'exp_year' => 2015)))\")";
+               . "'currency' => 'usd', 'source' => 'tok_1234'))\")";
             throw new Error\Api($message);
         }
     }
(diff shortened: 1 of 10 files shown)


remi-stripe:
Unfortunately the tokens can't be used in parallel which seems like a bug with the feature. It means that if you have 2 separate scripts using `tok_visa` at the same time, the second one will fail. This is why the test suite is breaking right now. Working on a fix internally!

lib/ApiResource.php
```diff
@@ -99,8 +99,7 @@ protected static function _validateParams($params = null)
             $message = "You must pass an array as the first argument to Stripe API "
                . "method calls.  (HINT: an example call to create a charge "
                . "would be: \"Stripe\\Charge::create(array('amount' => 100, "
-               . "'currency' => 'usd', 'card' => array('number' => "
-               . "4242424242424242, 'exp_month' => 5, 'exp_year' => 2015)))\")";
+               . "'currency' => 'usd', 'source' => 'tok_visa'))\")";
```
brandur-stripe: Minor, but should we change this to something like `tok_123` to help suggest that this isn't supposed to be a static string of some sort?

remi-stripe: Nice catch, just put `tok_1234` instead!

brandur-stripe: Sweet, thanks. LGTM pending backend fix!

remi-stripe:
@brandur-stripe  PTAL

brandur-stripe:
I can't believe that out of six PRs, a full three of them failed tests due to some intermittent issues. This is really not good.

PR #341, opened 2017-04-19
Allow a PSR-3 logger to be configured

### LINKED ISSUE #340: Use a PSR-3 logger instead or using error_log directly
It would be nice if the client had a way to attach an external error logger with a [PSR-3](http://www.php-fig.org/psr/psr-3/) interface ([Monolog](https://github.com/Seldaek/monolog) for example) instead of directly calling error_log:

https://github.com/stripe/stripe-php/blob/master/lib/StripeObject.php#L138

brandur-stripe:
Adds a configuration option so that a PSR-3 compatible logger can be
set on the library so that messages will end up there instead of
`error_log`.

Fixes #340.

@bkrausz-stripe Any thoughts on this approach? See context in #340.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,16 @@ $curl = new \Stripe\HttpClient\CurlClient(array(CURLOPT_PROXY => 'proxy.local:80
 
 Alternately, a callable can be passed to the CurlClient constructor that returns the above array based on request inputs. See `testDefaultOptions()` in `tests/CurlClientTest.php` for an example of this behavior. Note that the callable is called at the beginning of every API request, before the request is sent.
 
+### Configuring a Logger
+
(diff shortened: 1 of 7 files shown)


bkrausz-stripe:
Conceptually a big fan...why do we have `context` is nothing inside of our error logging uses it?

brandur-stripe:
> Conceptually a big fan...why do we have context is nothing inside of our error logging uses it?

Basically just to match up with the PSR-3 interface's signature. I figure it might also be useful someday if someone wants to write a quick implementation.

bkrausz-stripe:
Ah, didn't realize there was an interface there. This looks great to me then!

brandur-stripe:
Sweet. Thanks Brian! Taking this in.

PR #347, opened 2017-05-10
Update retrieve method documentation to accept arrays

### LINKED ISSUE #346: Retrieve functions doc blocks should allow an array as well as a string as ID property
The current doc blocks have the ID property of retrieve functions as only accepting strings but this needs updating to allow an array as when passing the expand option this needs to be done inline as an array containing the ID and an array with the expand options.

Example:
`Payout::retrieve(["id" => $payoutId, "expand" => ["destination", "balance_transaction"]]`

brandur-stripe:
Retrieve methods can take an array in some cases like when an expansion is requested. This patch updates documentation to better show that.

Fixes #346.

r? @bkrausz-stripe

### DIFF
--- a/lib/Account.php
+++ b/lib/Account.php
@@ -47,7 +47,7 @@ class Account extends ApiResource
     }
 
     /**
-     * @param string|null $id
+     * @param array|string|null $id
      * @param array|string|null $opts
      *
      * @return Account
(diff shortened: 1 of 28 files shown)


bkrausz-stripe:
Should we also mention "or an array with a key of `'id'`"?

brandur-stripe:
@bkrausz Good point. I fixed the doc strings as well now too. Want to take one more look?

bkrausz-stripe:
lgtm

brandur-stripe:
Thanks!

PR #350, opened 2017-06-04
Adds count() implementation to AttachedObject

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @gh-stripe @jlomas-stripe 

Replaces #329.

This PR adds a `count()` method to the `AttachedObject` class, which is used to store metadata as well as the infamous `additional_owners` attribute for account objects. Both `metadata` and `addtional_owners` can have a varying number of elements, so having a `count()` method available is useful.

This lets users write:

```php
count($account->additional_owners)
```

instead of:

```php
count($account->additional_owners->keys())
```

### DIFF
--- a/lib/AttachedObject.php
+++ b/lib/AttachedObject.php
@@ -2,6 +2,8 @@
 
 namespace Stripe;
 
+use Countable;
+
 /**
  * Class AttachedObject
  *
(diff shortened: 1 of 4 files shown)


ob-stripe:
Tests are failing because of an unrelated Travis issue with HHVM -- I'll tackle separately.

brandur-stripe:
Nice! Thanks for remembering that we should do this :)

Yeah, I like how this is much more narrowly scoped than #329. I'll wait on merge until we can get that build figured out, but LGTM.

ob-stripe:
Rebased on master -- tests are passing now.

brandur-stripe:
Released as 4.11.0.

PR #351, opened 2017-06-04
Use Trusty distribution (required for HHVM)

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Travis no longer supports HHVM on Precise (the default dist). This PR updates `.travis.yml` to use Trusty for HHVM builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,38 @@
 language: php
+sudo: false
 
-php:
-  - 5.3
-  - 5.4
-  - 5.5
-  - 5.6
-  - 7.0
-  - 7.1
-  - hhvm
-
-env:
-  - AUTOLOAD=1
-  - AUTOLOAD=0
+matrix:
+  include:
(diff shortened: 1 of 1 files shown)


brandur-stripe:
@ob-stripe Should we just bump to use Trusty everywhere? Kind of make sense to march forward here if possible.

ob-stripe:
That's what I initially tried to do, but unfortunately, PHP 5.3 is no longer supported on Trusty :(

Basically, we have to use Precise for 5.3, Trusty for HHVM, and either for all other versions. Since Trusty support is still in beta for Travis, I opted to use it only for HHVM.

brandur-stripe:
Thanks for explaining @ob-stripe! That reasoning sounds sane to me.

As mentioned in the issue you linked, if 5.3's really been out of support for a few years, we should consider dropping it as well and just going to all Trusty. We can make that a separate project though.

LGTM.

PR #352, opened 2017-06-05
Let clients add info to the X-Stripe-Client-User-Agent header

ob-stripe:
r? @brandur-stripe 

This is a followup to #332. Currently, we fill the `httplib` and `ssllib` keys in the `X-Stripe-Client-User-Agent` header with `curl_version()`, even if curl is not actually being used.

This PR lets HTTP clients provide an (optional) `getUserAgentInfo()` method. The method should return an array, and the contents of the array are appended to the `X-Stripe-Client-User-Agent` header, so HTTP clients are now responsible for providing information about whichever underlying library they use.

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -137,7 +137,7 @@ class ApiRequestor
         }
     }
 
-    private static function _defaultHeaders($apiKey)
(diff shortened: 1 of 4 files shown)


brandur-stripe:
Nice clean up! LGTM.

brandur-stripe:
Released as 4.12.0.

PR #353, opened 2017-06-12
Add support for ephemeral keys

grey-stripe:
This adds the `EphemeralKey` resource to the bindings.  Currently, the only supported operations are creation and deletion.

Creation is a bit of a special case, in that it requires `Stripe-Version` to be set explicitly (because this will be requested by the backend for use in the frontend, which may be expecting an arbitrary API version).  This means that it is an error if the header is set by the globally configured API version, and it must be provided explicitly in the `create` call.  As such, `create` will throw a `InvalidArgumentException` if `stripe_version` is not specified.

Everything else should be fairly standard.


lib/EphemeralKey.php
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Class EphemeralKey
+ *
+ * @property string $id
+ * @property string $object
+ * @property int $created
+ * @property string $purpose
+ * @property int $size
+ * @property string $type
```
brandur-stripe: This list isn't right is it?

brandur-stripe:
Left a comment above, but otherwise +1. Thanks!

PR #356, opened 2017-06-21
Fix namespaces

akeeman:
Currently, these refer to `\Stripe\HttpClient\Error\{Api,ApiConnection}`, which is not right and throws warnings in my IDE.


brandur-stripe:
Looks good to me. Thank you!

PR #357, opened 2017-06-22
Power test suite with external API stub

brandur:
Move to using an external API stub in the test suite instead of issuing live API calls so that we can have tests that run more quickly and more reliably.

For the time being, I've just converted the plan test suite over to use the new scheme.

This is an experimental prototype for demonstrative purposes and should not be merged.

See also:

* https://github.com/stripe/stripe-java/pull/384
* https://github.com/stripe/stripe-go/issues/424
* https://github.com/stripe/stripe-ruby/pull/553

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,9 +34,28 @@ matrix:
       dist: trusty
       env: AUTOLOAD=0
 
+env:
+  global:
+    - STRIPE_STUB_PORT=6065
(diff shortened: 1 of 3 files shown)


deontologician-stripe:
Commenting to appease checker for now

ob-stripe:
Closing this one since we ported the tests to stripe-mock in #389.

PR #358, opened 2017-06-22
Cache composer files

brandur:
It seems that the build's speed is pretty sizably impacted by installing
Composer dependencies from `build.php`. Lets try caching them in Travis
to see if we get any speed up.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,5 +34,9 @@ matrix:
       dist: trusty
       env: AUTOLOAD=0
 
+cache:
+  directories:
+    - $HOME/.composer/cache/files
+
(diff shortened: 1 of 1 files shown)


brandur-stripe:
So it's a little hard to tell, but I think this did help.

You can see in the build matrix below that the first build had pretty standard timing, but the subsequent do seem to be faster, even when they were in the same version:
screenshot 2017-06-22 12 57 04

Compare this to an older build where the timing was fairly uniform:
screenshot 2017-06-22 13 04 08

brandur-stripe:
Anyway, I doubt it'll help much, but going to bring this is in as a very minor improvement.

PR #359, opened 2017-06-27
Add parameters when calling `pay` on an invoice

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @ark-stripe 

Users can now provide a `source` parameter when paying invoices, but the `pay` method currently does not let users provide any parameters. This PR fixes that.

Unfortunately, this is a breaking change because if someone is currently passing a `$opts` argument, they'll need to modify their code since it's now the second argument. I didn't want to add `$params` as the second argument since it would be inconsistent with ~every other method in the library.

### DIFF
--- a/lib/Invoice.php
+++ b/lib/Invoice.php
@@ -83,10 +83,10 @@ class Invoice extends ApiResource
     /**
      * @return Invoice The paid invoice.
      */
-    public function pay($opts = null)
+    public function pay($params = null, $opts = null)
     {
         $url = $this->instanceUrl() . '/pay';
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Yeah, the only other alternative I can think of is to add a `pay_with_params` or something like that, but it's pretty ugly. Given that many users will probably not be broken by this (only those providing custom `$opts`), I agree that this is the best way forward.

LGTM.

brandur-stripe:
Released as 5.0.0.

PR #360, opened 2017-06-29
Add OAuth methods

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

This PR adds support for sending requests to the OAuth API directly from the PHP library.

I've tried to stay as close as possible to the existing Ruby and Python implementations:
- https://github.com/stripe/stripe-ruby/pull/523
- https://github.com/stripe/stripe-python/pull/323

Nothing major to note, except that I moved the `CurlClient::encode` function to `Util\Util::urlEncode` because I needed to use it (for `authorizeUrl`) and it made little sense to have `OAuth` depend on `CurlClient`. Now both `OAuth` and `CurlClient` depend on `Util`, which seems more reasonable IMHO.

### DIFF
--- a/examples/oauth.php
+++ b/examples/oauth.php
@@ -1,0 +1,55 @@
+<?php
+
+require('../init.php');
+
+\Stripe\Stripe::setApiKey(getenv('STRIPE_SECRET_KEY'));
+\Stripe\Stripe::setClientId(getenv('STRIPE_CLIENT_ID'));
+
+
+if (isset($_GET['code'])) {
+    // The user was redirected back from the OAuth form with an authorization code.
+    $code = $_GET['code'];
+
+    try {
+        $resp = \Stripe\OAuth::token(array(
+            'grant_type' => 'authorization_code',
+            'code' => $code,
+        ));
+    } catch (\Stripe\Error\OAuth\OAuthBase $e) {
+        exit("Error: " . $e->getMessage());
+    }
+
+    $accountId = $resp->stripe_user_id;
+
+    echo "<p>Success! Account <code>$accountId</code> is connected.</p>\n";
+    echo "<p>Click <a href=\"?deauth=$accountId\">here</a> to disconnect the account.</p>\n";
+
+} elseif (isset($_GET['error'])) {
+    // The user was redirect back from the OAuth form with an error.
+    $error = $_GET['error'];
+    $error_description = $_GET['error_description'];
+
+    echo "<p>Error: code=$error, description=$error_description</p>\n";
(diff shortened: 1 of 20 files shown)


examples/oauth.php
```diff
@@ -0,0 +1,46 @@
...
+        ));
+    } catch (\Stripe\Error\OAuth\OAuthBase $e) {
+        exit("Error: " . $e->getMessage());
+    }
+
+    echo "<p>Success! Account <code>$accountId</code> is disonnected.</p>\n";
+    echo "<p>Click <a href=\"?\">here</a> to restart the OAuth flow.</p>\n";
+
+} else {
+    $url = \Stripe\OAuth::authorizeUrl(array(
```
remi-stripe: You're not handling the error case here for example when the customer decides not to Connect or if something wrong happens. There would be no `code` in the URL but there would be an `error` parameter that should be handled.

ob-stripe: Good point, I added an `if` clause to handle errors.

lib/OAuth.php
```diff
@@ -0,0 +1,98 @@
...
+        list($response, $apiKey) = $requestor->request(
+            'post',
+            '/oauth/deauthorize',
+            $params,
+            null
+        );
+        return Util\Util::convertToStripeObject($response->json, $opts);
+    }
+
+    private static function _getClientId($params = null)
```
brandur-stripe: I just noticed that in terms of using an underscore prefix for `private` methods, we're doing it about half and half throughout the rest of code right now. I'm not sure if PHP has a general convention, but I'd guess it shouldn't be as important given that it has real visibility modifiers (unlike Python et all). Either way, fine to leave as is.

brandur-stripe:
Incredible work OB!

This looks really solid to me. Is it ready for release?

ob-stripe:
@brandur-stripe I think so. Since the implementation is so close to the Ruby & Python ones, I don't expect any bad surprises. I also wrote a basic but complete example to test the entire OAuth flow (redirect to the OAuth form, use the authorization code to connect the account, then disconnect the account) and it worked fine.

brandur-stripe:
> Since the implementation is so close to the Ruby & Python ones, I don't expect any bad surprises. I also wrote a basic but complete example to test the entire OAuth flow (redirect to the OAuth form, use the authorization code to connect the account, then disconnect the account) and it worked fine.

Thanks for the diligence here :) Awesome, shipping this out then.

brandur-stripe:
Released as 5.1.0.

PR #361, opened 2017-07-02
OrderReturn: override parent className function to add underscore

danasf:
`OrderReturn` appears to be broken, is trying to call `/v1/orderreturns` when the endpoint should contain an underscore.

For example `\Stripe\OrderReturn::all(array("limit" => 3))` will throw

```
Fatal error: Uncaught exception 'Stripe\Error\InvalidRequest' with message 'Unrecognized request URL (GET: /v1/orderreturns)
```

### DIFF
--- a/lib/OrderReturn.php
+++ b/lib/OrderReturn.php
@@ -10,6 +10,17 @@ namespace Stripe;
 class OrderReturn extends ApiResource
 {
     /**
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Thanks @danasf!

brandur-stripe:
Released as 5.1.1.

PR #364, opened 2017-08-01
Fixes phpdoc declared type to qualify namespace correctly

asgrim:
When using `Collection#autoPagingIterator()` I get type warnings in PhpStorm as the `@return` type is hinted incorrectly.

This will fix this warning in PhpStorm inside Collection:

15bcaa

And this warning when using the `autoPagingIterator` method to fetch the iterator, for example:

113eba

### DIFF
--- a/lib/Collection.php
+++ b/lib/Collection.php
@@ -56,7 +56,7 @@ class Collection extends ApiResource
     }
 
     /**
-     * @return AutoPagingIterator An iterator that can be used to iterate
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Oops, yep, that looks better. Thanks!

brandur-stripe:
Released as 5.1.2.

asgrim:
Thanks for the quick response! :metal:

Issue #366, opened 2017-08-14
Cannot update address_kana (and more) on JP custom account


floriangaerber:
| Component | Version |
| --- | --- |
| PHP | 7.0.10 |
| `stripe/stripe-php` | 4.13.0 |
| Stripe API | 2017-06-05 (latest) (test mode) |

I am trying to update a JP custom accounts `legal_entity.address_kana.postal_code` property.

The request is submitted without error, but also without the `legal_entity.address_kana.postal_code` property.

I've used `Symfony\VarDumper` to inspect the `Stripe\Account` object and found the following:
- `$account->legal_entity` is an instance of `Stripe\AttachedObject`
- `$account->legal_entity->address` of a DE custom account is an instance of `Stripe\AttachedObject`
  - After updating a value, the corresponding field is listed in `unsavedValues` property of the `AttachedObject`
  - After using `$account->save();`, `unsavedValues` is empty again
  - Updating a DE custom account's address works (address parameters are submitted)
- `$account->legal_entity->address_kana` of a JP custom account is an instance of `Stripe\StripeObject`
  - After updating a value, the corresponding field is listed in `unsavedValues` property of the `StripeObject`
  - After using `$account->save();`, `unsavedValues` still lists the changed value, when no other fields were updated.
    - When also updating another value, in the same call, such as `first_name`, `address_kana` is reset to its previous state.
  - Updating a JP custom account does work (i.e. `legal_entity.first_name`, ...), but updating the address does not. (Neither kana nor kanji)

I had a brief look into it and believe the Stripe client fails to collect `address_kana` and `address_kanji` in `Stripe\StripeObject::serializeParameters()` [here](https://github.com/stripe/stripe-php/blob/master/lib/StripeObject.php#L235), because they aren't listed in `Stripe\StripeObject::nestedUpdatableAttributes` [here](https://github.com/stripe/stripe-php/blob/master/lib/StripeObject.php#L29).

### Appendix
<details>

**Minimal Example:**
```php
$account = Stripe\Account::retrieve($id);
$account->legal_entity->address_kana->postal_code = '1500001';
$account->save();
```

Result: No API call is made (checked logs), likely because client did not detect any changes.

**Verification:**

```php
$account = Stripe\Account::retrieve($id);
$account->legal_entity->address_kana->postal_code = '1500001';
$account->legal_entity->first_name = 'Max';
$account->save();
```
Result: API call is made successfully, logs show "Request POST body":
```json
{
  "legal_entity": {
[...]

PR #367, opened 2017-08-14
Submit updated address_kana and address_kanji for Custom accounts

remi-stripe:
Before that change, updating the fields under the `address_kana` or `address_kanji` in the `legal_entity` hash of a Custom JP account would not do anything as the data was ignored.

Fixes https://github.com/stripe/stripe-php/issues/366

r? @brandur-stripe


floriangaerber:
That's what I just tried locally and it works just fine! Thanks for the quick fix!

brandur-stripe:
Thanks!

Test failure here is unrelated. Pulling this in.

PR #371, opened 2017-09-08
Add `shipping` to the list of nested attributes that can be updated

remi-stripe:
This ensures that updating the shipping details on an order and then saving the order properly updates the order.

The issue is that when doing this, updating a customer's shipping address and passing only one field such as `name` does not work anymore.

This PR is here to clarify what's going on with `nestedUpdatableAttributes` and what the right approach would be.

@brandur-stripe do you know why the order update would start to work and not the customer one?

Here's some example code where `customer[shipping][name]` and `customer[shipping][address][line1]` are set already on the customer:

```
$customer = \Stripe\Customer::retrieve("cus_XXXXXX");
$customer->shipping->name = "myname";
$customerUpdated = $customer->save();
```

It worked before this PR and now returns an error:
> Missing required param: shipping[address].

I think this is expected behaviour and we've seen similar issues with the Charge object before but the fact that it used to work leads to two questions:
* is this a breaking change?
* why was it working on the customer and not Order?

### DIFF
--- a/lib/StripeObject.php
+++ b/lib/StripeObject.php
@@ -28,7 +28,7 @@ class StripeObject implements ArrayAccess, JsonSerializable
         self::$permanentAttributes = new Util\Set(array('_opts', 'id'));
         self::$nestedUpdatableAttributes = new Util\Set(array(
             'metadata', 'legal_entity', 'address', 'dob', 'payout_schedule', 'transfer_schedule', 'verification',
-            'tos_acceptance', 'personal_address', 'address_kana', 'address_kanji',
+            'tos_acceptance', 'personal_address', 'address_kana', 'address_kanji', 'shipping',
             // will make the array into an AttachedObject: weird, but works for now
             'additional_owners', 0, 1, 2, 3, 4, // Max 3, but leave the 4th so errors work properly
             'inventory',


brandur-stripe:
Hey Remi, I just looked into this a bit, and I don't think it works currently. It just looks like it does.

Code:

``` php
<?php

require_once('./init.php');

\Stripe\Stripe::setApiKey("...");

echo "Running create\n";
$cu = \Stripe\Customer::create(array(
  "shipping" => array(
    "name" => "pre update name",
    "address" => array(
        "line1" => "An address"
    )
  )
));
echo "Name post-create is: " . $cu->shipping->name . "\n\n";

echo "Running update\n";
$cu->shipping->name = "POST update name";
$cu= $cu->save();
echo "Name post-update is: " . $cu->shipping->name . "\n\n";

echo "Running retrieve\n";
$cu = \Stripe\Customer::retrieve($cu->id);
echo "Name post-retrieve is: " . $cu->shipping->name . "\n\n";
```

No error is thrown, but here's the output:

```
$ php customer_update_test.php
Running create
array(1) {
  ["shipping"]=>
  array(2) {
    ["name"]=>
    string(15) "pre update name"
    ["address"]=>
    array(1) {
      ["line1"]=>
      string(10) "An address"
    }
  }
}
Name post-create is: pre update name

Running update
array(0) {
}
Name post-update is: POST update name

Running retrieve
Name post-retrieve is: pre update name
```

As you can see, if we fetch the object back again, the name hasn't actually changed, meaning that the call to save did nothing.

It's not great, but I think what you did here adding `shipping` is probably the right thing to do, and it looks like the API will want an entire `shipping` hash to be sent up. Does that sound right to you?

remi-stripe:
@brandur-stripe gotcha, that makes sense then and explains why it felt like it worked. Thanks for digging into this!

brandur-stripe:
Something weird is going on with this branch's build — PHP 5.3 cannot seem to unpack even after a few retries, but for some reason it seems to be working on `master`.

I suspect that Travis is caching a bad result or something (it seems unlikely that this branch is causing the problem), so I'm bringing this in.

brandur-stripe:
Released as 5.2.1.

PR #374, opened 2017-09-20
Set correct mapping for FileUpload

frankrrr:
Just found a bug while working with strict return types. Also, according to [the docs](https://stripe.com/docs/api/curl#file_uploads), `"object": "file_upload"` gets returned:

```
{
  "id": "file_1B44w82IJJWDZPe7TAatUjAc",
  "object": "file_upload",
  "created": 1505901368,
  "purpose": "identity_document",
  "size": 158860,
  "type": "png",
  "url": "https://stripe-upload-api.s3.amazonaws.com/uploads/file_1B44w82IJJWDZPe7TAatUjAc?AWSAccessKeyId=KEY_ID\u0026Expires=TIMESTAMP\u0026Signature=SIGNATURE"
}
```

Therefore changed the mapping from `file` to `file_upload`.

### DIFF
--- a/lib/Util/Util.php
+++ b/lib/Util/Util.php
@@ -82,7 +82,7 @@ abstract class Util
             'invoice' => 'Stripe\\Invoice',
             'invoiceitem' => 'Stripe\\InvoiceItem',
             'event' => 'Stripe\\Event',
-            'file' => 'Stripe\\FileUpload',
+            'file_upload' => 'Stripe\\FileUpload',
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Yes that looks right! Test failures are unrelated and fixed in #375, so I'm pulling this in as is.

Thanks! And slick job adding the extra test too!

brandur-stripe:
Master's green. Fix released in 5.2.2.

PR #375, opened 2017-09-20
Stop testing Bitcoin APIs

brandur:
The Bitcoin APIs have been deprecated in favor of Stripe's new generic
sources endpoint that handles all APMs.

Bitcoin-related tests are no longer working because they don't have the
endpoints they need to run anymore. For the time being to avoid a major
release, let's just remove tests instead of the whole API. We can
piggyback on another major release somewhere down the road to strip the
rest of Bitcoin from the library.

r? @remi-stripe


brandur-stripe:
Thanks Remi.

PR #376, opened 2017-09-27
Add PHPDoc for Card.php

Samistine:
Document Card.php exactly like how https://github.com/stripe/stripe-php/blob/master/lib/Charge.php is documented.


brandur-stripe:
Looks good to me. Thanks!

brandur-stripe:
Released a version as 5.2.3.

PR #378, opened 2017-10-11
detach method for detaching sources from customers

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @stan-stripe @will-stripe

- Renames the `Source::delete` method to `Source::detach`. `detach` is more descriptive of what the method actually does (i.e. detaching the source object from its customer object, not actually deleting the source object).

- Alias `detach` to `delete` to avoid a breaking change. We should remove the `delete` method in next major version. (No tests here as deprecated methods are only marked via the `@deprecated` PHPDoc tag in the library.)

### DIFF
--- a/lib/Source.php
+++ b/lib/Source.php
@@ -67,11 +67,11 @@ class Source extends ApiResource
 
     /**
      * @param array|null $params
-     * @param array|string|null $opts
+     * @param array|string|null $options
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Released as 5.3.0.

PR #383, opened 2017-10-24
Static methods for accessing nested resources

### LINKED ISSUE #363: Create a Card
Hello

Is there a way to create a card from this library ?
I want to translate [this call](https://stripe.com/docs/api#create_card) but the Card Class doesn't contain a `create` method.

I could do a PR to add these features but the use of static method is a problem, this class would have to overwrite the _create, _list, _update methods from ApiResource (since the URL expects a parameter).

```php

class Card extends ApiResource
{

    public static function create(string $customer, ?array $params = null, ?array $options = null): StripeObject
    {
        self::_validateParams($params);
        $url = "/v1/customers/" . $customer . "/sources";
        [$response, $opts] = static::_staticRequest('post', $url, $params, $options);
        $obj = Util::convertToStripeObject($response->json, $opts);
        $obj->setLastResponse($response);
        return $obj;
    }

}
```

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Fixes #363.

Basically the same thing as https://github.com/stripe/stripe-ruby/pull/597 and https://github.com/stripe/stripe-python/pull/350, but with less fancy metaprogramming because PHP 😢

### DIFF
--- a/lib/Account.php
+++ b/lib/Account.php
@@ -37,6 +37,9 @@ namespace Stripe;
  */
 class Account extends ApiResource
 {
+    const PATH_EXTERNAL_ACCOUNTS = '/external_accounts';
+    const PATH_LOGIN_LINKS = '/login_links';
+
     public function instanceUrl()
     {
         if ($this['id'] === null) {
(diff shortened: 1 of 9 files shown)


brandur-stripe:
Wow @ob-stripe, what are you doing to fit 36 hours into a day? Whatever it is, I want it ;)

Nice work! There were a couple build failures in the matrix, but I think they were intermittent problems. Going to go ahead and merge this.

PR #386, opened 2017-10-30
Add support for exchange_rates API requests

ob-stripe:
r? @brandur-stripe
cc @stripe/api-libraries @alixander-stripe

Adds support for the new `/v1/exchange_rates` endpoints.

### DIFF
--- a/tests/ExchangeRateTest.php
+++ b/tests/ExchangeRateTest.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace Stripe;
+
+class ExchangeRateTest extends TestCase
+{
+    public function testRetrieve()
+    {
+        $this->mockRequest(
+            'GET',
+            '/v1/exchange_rates/usd',
+            array(),
+            array(
+                'id' => 'usd',
+                'object' => 'exchange_rate',
+                'rates' => array('eur' => 0.845876),
+            )
+        );
+
+        $currency = "usd";
+        $rates = ExchangeRate::retrieve($currency);
+        $this->assertEquals('exchange_rate', $rates->object);
+    }
+
+    public function testList()
+    {
+        $this->mockRequest(
+            'GET',
+            '/v1/exchange_rates',
+            array(),
+            array(
+                'object' => 'list',
+                'data' => array(
+                    array(
+                        'id' => 'eur',
+                        'object' => 'exchange_rate',
+                        'rates' => array('usd' => 1.18221),
+                    ),
+                    array(
+                        'id' => 'usd',
+                        'object' => 'exchange_rate',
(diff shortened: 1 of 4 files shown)


brandur-stripe:
I hate to ask, but mind bumping this one so `ExchangeRates` becomes singular?

ob-stripe:
Done, ptal @brandur-stripe

tests/ExchangeRateTest.php
```diff
@@ -0,0 +1,52 @@
...
+{
+    public function testRetrieve()
+    {
+        $this->mockRequest(
+            'GET',
+            '/v1/exchange_rates/usd',
+            array(),
+            array(
+                'id' => 'usd',
+                'object' => 'exchange_rates',
```
brandur-stripe: Maybe switch this over to `exchange_rate` (and there are three other instances below to replace in this file as well).

brandur-stripe:
Minor comment, but otherwise looks good. Thanks OB!

ptal @ob-stripe

ob-stripe:
Oops! Should be fixed now.

tests/ExchangeRateTest.php
```diff
@@ -0,0 +1,52 @@
...
+            array(
+                'id' => 'usd',
+                'object' => 'exchange_rate',
+                'rates' => array('eur' => 0.845876),
+            )
+        );
+
+        $currency = "usd";
+        $rates = ExchangeRate::retrieve($currency);
+        $this->assertEquals('exchange_rate', $rates->object);
```
ob-stripe: I really wish we could do `$this->assertInstanceOf(ExchangeRate::class, $rates);` here instead, but unfortunately that doesn't fly with PHP <= 5.4 😢

brandur-stripe: Ugh ...

Well, it's worth noting that [5.4 has been EOLed for more than two years](http://php.net/supported-versions.php) (since mid-2015) and hasn't even been receiving security fixes for more than a year, which means that it's not safe to use. This approach works fine for now, but it's totally plausible to drop support for it at this point. (If you're not updating off of PHP 5.4, you're also probably not bothering to update your version of stripe-php.)

brandur-stripe: Actually, I misread that. It's _5.5_ has been EOLed for two years and hasn't been receiving updates for a year.

5.4 has been EOLed for three years (mid-2014) and hasn't been receiving updates for two years!

arcanedev-maroc: http://php.net/supported-versions.php

brandur-stripe:
Released as 5.6.0.

PR #387, opened 2017-10-30
Add personal_address_kana and personal_address_kanji for jp

kannokanno:
Hi.

When country is jp, update account does not work.
Because personal_address_kana and personal_address_kanji are necessary.

Thanks.


brandur-stripe:
Thank you @kannokanno! LGTM.

brandur-stripe:
Released as 5.5.1.

PR #388, opened 2017-10-30
Clean up + sort `$nestedUpdateableAttributes` array

brandur:
Here we cleanup `$nestedUpdatableAttributes` in `StripeObject` by moving
every field name to its own line and then alphabetizing them. This makes
it a little easier to see what's going on when people add new ones, and
gives them a deterministic place in the array with which to make the
addition.

r? @ob-stripe

### DIFF
--- a/lib/StripeObject.php
+++ b/lib/StripeObject.php
@@ -27,13 +27,29 @@ class StripeObject implements ArrayAccess, JsonSerializable
     {
         self::$permanentAttributes = new Util\Set(array('_opts', 'id'));
         self::$nestedUpdatableAttributes = new Util\Set(array(
-            'metadata', 'legal_entity', 'address', 'dob', 'payout_schedule', 'transfer_schedule', 'verification',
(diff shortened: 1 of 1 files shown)


lib/StripeObject.php
```diff
@@ -27,13 +27,29 @@ public static function init()
...
+            'additional_owners',
+            'address',
+            'address_kana',
+            'address_kanji',
+            'dob',
             'inventory',
+            'legal_entity',
+            'metadata',
+            'payout_schedule',
             'owner',
```
remi-stripe: this should be a line above

brandur-stripe: Oops. Absolutely right! Amended in bdbfe87.

Good to go otherwise?

brandur-stripe:
> We likely want to find a more reliable solution than this hardcoded list though in the near future since it has caused so many issues the past 12 months

Totally agree.

Thanks!

brandur-stripe:
Thanks guys. Had one intermittent failure in the test matrix, but I'm 99% sure this is fine. Merging.

PR #389, opened 2017-11-13
Start using stripe-mock for tests

ob-stripe:
@remi-stripe Okay, here's a first attempt at using stripe-mock in the PHP library.

It's still pretty rough, and has some limitations/quirks when compared to the ongoing work in Python:

- instead of calling `assert_requested`, you must call `$this->expectsRequest` immediately _before_ the request is sent

- if you need to stub a request, you can call `$this->stubRequest`

There are probably some cases I have not anticipated that will reveal themselves as we port the tests.


tests/ChargeTest.php
```diff
@@ -2,141 +2,90 @@
 
 namespace Stripe;
 
-class ChargeTest extends TestCase
-{
-    public function testUrls()
```
remi-stripe: @ob-stripe Am I correct to remove this as it's mostly tested elsewhere? Wasn't sure we tested URL-encoding properly in the lib.

tests/ApplicationFeeTest.php
```diff
@@ -2,92 +2,70 @@
 
 namespace Stripe;
 
-class ApplicationFeeTest extends TestCase
+define('TEST_RESOURCE_ID', 'fee_123');
+define('TEST_FEEREFUND_ID', 'fr_123');
+
+class ApplicationFeeTest extends StripeMockTestCase
 {
-    public function testUrls()
```
remi-stripe: @ob-stripe Flagging that `ApplicationFee::update()` exists even though it's unsupported in the API. Likely want to remove from the code.

tests/CustomerTest.php
```diff
@@ -2,337 +2,220 @@
...
-        $updatedCustomer = Customer::retrieve($customer->id);
-        $updatedSources = $updatedCustomer->sources->all();
-        $this->assertSame(count($updatedSources["data"]), 2);
+        $customer = Customer::retrieve(TEST_RESOURCE_ID);
+        $this->stubRequest(
+            'post',
+            '/v1/customers/' . $customer->id . '/subscription',
+            array("plan" => "plan")
+        );
+        $customer->updateSubscription(array("plan" => "plan"));
```
remi-stripe: @ob-stripe This test (and the next one) don't test the returned value because... nothing is returned? Didn't seem to matter but I wanted to flag

tests/CustomerTest.php
```diff
@@ -2,337 +2,220 @@
...
-        );
-
-        $sources = Customer::allsources(
-            'cus_123'
+        $this->expectsRequest(
+            'get',
+            '/v1/customers/' . TEST_RESOURCE_ID . '/sources'
         );
+        $resources = Customer::allSources(TEST_RESOURCE_ID);
+        $this->assertTrue(is_array($resources->data));
```
remi-stripe: this one doesn't test the class because stripe-mock currently returns a list of `Account` which is incorrect/impossible. Flagged internally

tests/AccountOAuthTest.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace Stripe;
```
remi-stripe: @ob-stripe  I had to create a new file because PHPUnit seems to require one class per file otherwise it just ignores the other ones.

tests/AccountTest.php
```diff
@@ -2,469 +2,156 @@
...
-        $account = Account::create(array('managed' => true));
-        $account->legal_entity->first_name = 'Bob';
-        $account->save();
-
-        $this->assertSame('Bob', $account->legal_entity->first_name);
+        $resource->save();
+        $this->assertSame("Stripe\\Account", get_class($resource));
     }
 
-    public function testUpdateLegalEntity()
```
remi-stripe: @ob-stripe I removed those tests but I am worried that we're not testing this properly anymore. Should this live here or in a separate test that properly handles sub-fields/nested resources updates?

tests/SourceTest.php
```diff
@@ -2,211 +2,98 @@
 
 namespace Stripe;
 
-class SourceTest extends TestCase
+define('TEST_RESOURCE_ID', 'src_123');
+
+class SourceTest extends StripeMockTestCase
```
remi-stripe: @ob-stripe Flagging that `Source.php` has the `all` method even though the API never supported it. Did not add a test since this should not exist in the first place I think

tests/SourceTest.php
```diff
@@ -2,211 +2,98 @@
...
+        $resource->delete();
+        $this->assertSame("Stripe\\Source", get_class($resource));
     }
 
-    /**
-     * @expectedException Stripe\Error\Api
-     */
-    public function testDetachUnattached()
+    /*
+    This does not yet work with our version of PHP Unit?
```
remi-stripe: @ob-stripe Tried to catch the exception but got errors because `expectException` is not defined. Seems to be part of a different class than the one you inherit from for StripeMockCase. Didn't want to break everything so deferring to you

tests/SourceTest.php
```diff
@@ -2,211 +2,98 @@
...
-        $this->assertSame($source->owner['name'], "Stripey McStripe");
-        $this->assertSame($source->owner['address']['line1'], "Test Address");
-        $this->assertSame($source->owner['address']['city'], "Test City");
-        $this->assertSame($source->owner['address']['postal_code'], "12345");
-        $this->assertSame($source->owner['address']['state'], "Test State");
-        $this->assertSame($source->owner['address']['country'], "Test Country");
+        $resource = Source::update(TEST_RESOURCE_ID, array(
+            "metadata" => array("key" => "value"),
+        ));
+        $this->assertSame("Stripe\\Source", get_class($resource));
```
remi-stripe: @ob-stripe  Flagging that we don't test owner updated anymore and might want to.

remi-stripe:
@ob-stripe  Okay did almost all resources. I think we have EphemeralKey and Card/BankAccount left!

I left quite a few comments about things broken or that I need help with.

Also I see ```
Notice: Constant TEST_RESOURCE_ID already defined in stripe-php/tests/SubscriptionTest.php on line 5
PHP Notice:  Constant TEST_RESOURCE_ID already defined in stripe-php/tests/ThreeDSecureTest.php on line 5

Notice: Constant TEST_RESOURCE_ID already defined in stripe-php/tests/ThreeDSecureTest.php on line 5
PHP Notice:  Constant TEST_RESOURCE_ID already defined in stripe-php/tests/TokenTest.php on line 5
```

I think the constant was the wrong approach and I should have used a class constant instead though it's more verbose.

Let me know what you think!

File: tests/ChargeTest.php (hunk 1 above)
Comment (ob-stripe) [reply tests/ChargeTest.php]: It doesn't seem like we do atm. That said I agree that `ChargeTest` is not the correct place for this. We should add an `ApiResourceTest`.

File: tests/SourceTest.php (hunk 7 above)
Comment (ob-stripe) [reply tests/SourceTest.php]: 👍  Let's fix this in a separate PR.

File: tests/AccountOAuthTest.php (hunk 5 above)
Comment (ob-stripe) [reply tests/AccountOAuthTest.php]: Ack. I'll extend `stubRequest` to be able to handle non-`api.stripe.com` requests and we'll be able to move the test back in `AccountTest`.

File: tests/ApplicationFeeTest.php (hunk 2 above)
Comment (ob-stripe) [reply tests/ApplicationFeeTest.php]: 👍

File: tests/CustomerTest.php (hunk 3 above)
Comment (ob-stripe) [reply tests/CustomerTest.php]: You can add a fourth parameter to `stubRequest` with the array that you want to return.

File: tests/CustomerTest.php (hunk 4 above)
Comment (ob-stripe) [reply tests/CustomerTest.php]: Ack. Can you add a comment with a TODO so that hopefully we remember to add the proper assertion once stripe-mock does the right thing?

File: tests/CustomerTest.php (hunk 3 above)
Comment (remi-stripe) [reply tests/CustomerTest.php]: @ob-stripe I know that, but I am testing the real function and that function did not return a resource which seems like a bug in the library itself (or potentially stripe-mock)

File: tests/CustomerTest.php (hunk 4 above)
Comment (remi-stripe) [reply tests/CustomerTest.php]: Not sure we need an assertion here to be honest and it does not really matter since everything is dynamic and any issue would come from stripe-mock or the API and not the library

File: tests/SourceTest.php (hunk 8 above)
Comment (ob-stripe) [reply tests/SourceTest.php]: Yes, we're on a stupidly old version of PHPUnit because of PHP 5.3 :yellatcloud:

You can still expect exceptions via a comment with the `@expectedException` tag. Take a look at `WebhookTest.php` for a few examples.

Comment (remi-stripe) [reply tests/SourceTest.php]: Ah cool, I noticed this one but assumed it was unrelated. Fixed in https://github.com/stripe/stripe-php/pull/389/commits/9ff4172531127459320aef1945e69cdc5eec5d7f

File: tests/CustomerTest.php [hunk 10]
```diff
@@ -137,30 +137,43 @@ public function testCanUpdateSubscription()
...
         $customer = Customer::retrieve(TEST_RESOURCE_ID);
         $this->stubRequest(
             'delete',
-            '/v1/customers/' . $customer->id . '/discount'
+            '/v1/customers/' . $customer->id . '/discount',
         );
-        $resource = $customer->deleteDiscount();
-        $this->assertSame($resource, null);
+        $customer->deleteDiscount();
+        $this->assertSame($customer->discount, null);
```
remi-stripe: While this works, this is IMO incorrect. stripe-mock does not handle discount in the first place and it's null mostly by luck no?

tests/CustomerTest.php
```diff
@@ -137,30 +137,43 @@ public function testCanUpdateSubscription()
...
+            array("plan" => "plan"),
+            array(
+                "object" => "subscription",
+                "id" => "sub_foo"
+            )
         );
-        $customer->updateSubscription(array("plan" => "plan"));
+        $resource = $customer->updateSubscription(array("plan" => "plan"));
+        $this->assertSame("Stripe\\Subscription", get_class($resource));
+        $this->assertSame("sub_foo", $customer->subscription->id);
```
remi-stripe: This works but it should not be necessary if the lib or stripe-mock did the right thing in the first place right?

tests/CustomerTest.php
remi-stripe: nevermind this is stubbed hence why it did not do what I expected it to do

remi-stripe: nevermind this is stubbed hence why it did not do what I expected it to do

tests/CustomerTest.php
remi-stripe: nevermind this is stubbed hence why it did not do what I expected it to do

tests/CustomerTest.php
ob-stripe: Mostly to be consistent with other list tests, but fair enough, we don't really need to assert the contents of every list returned by stripe-mock -- we're already checking that a list is returned which is probably enough to ensure that we're hitting the correct URL.

remi-stripe:
> I'd say open a separate PR, and leave the tests for now, we can update them once the other PR is merged in and we rebase this one.

do you mean I should undo this change in the middle of the PR? If it's not a breaking change it might be easier to just keep it there as something we caught?

tests/AccountOAuthTest.php
ob-stripe: I pushed another commit to move this test directly in `AccountTest`.

tests/AccountTest.php
ob-stripe: Good question. I suppose since this is all handled in `StripeObject` directly, we should test this in `StripeObjectTest`.

ob-stripe:
I think I've addressed all your comments for now, but let me know if I missed anything!

ob-stripe:
Test post, please ignore.

brandur-stripe:
Amazing OB! You're making a dent in the universe here.

How much further does this one have to go?

ob-stripe:
Remi helped a ton too :)

I think this one is ready for review -- I removed the WIP tag.

tests/Stripe/SubscriptionItemTest.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+namespace Stripe;
+
+class SubscriptionItemItemTest extends TestCase
```
brandur-stripe: `ItemItem` should just be `Item`.

tests/Stripe/TransferReversalTest.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Stripe;
+
+class TransferreversalTest extends TestCase
```
brandur-stripe: `Transferreversal` should be `TransferReversal`.

tests/bootstrap.php
```diff
@@ -1,3 +1,43 @@
...
+         "instance?";
+    exit(1);
+}
+
+if (version_compare($version, MOCK_MINIMUM_VERSION) == -1) {
+    echo "Your version of stripe-mock (" . $version . ") is too old. The minimum " .
+         "version to run this test suite is " . MOCK_MINIMUM_VERSION . ". " .
+         "Please see its repository for upgrade instructions.\n";
+    exit(1);
+}
```
brandur-stripe: Thanks a lot for porting all this stripe-mock setup code :) I love how consistent it is across languages!

brandur-stripe:
@ob-stripe @remi-stripe Incredible cleanup here guys! It's really nice to see that you fixed a few things during the port.

So I must admit that I didn't cover every single line in depth on this one, but I took a basic pass through the entire diff, and the general abstractions look good. I probably missed something somewhere, but IMO we should just go with the merge and put in any minor fixups as we find them.

I left a couple minor comments above with regards to naming, but as soon as that's fixed, +1 for merge!

brandur-stripe:
Oh and more more thing: I just checked the build times and this takes total build time (i.e. the entire version matrix) from ~85 minutes to ~25 minutes, which is good, but on the surface wouldn't seem like as much of a speedup as you might expect. However, if you dig a little deeper, the time to just run tests (minus setup overhead) went from ~3 minutes to ~10 seconds, which is awesome! (Also, this should fix build intermittency which is an even bigger deal.)

After this is in, we should see if we can spend a little time making sure that we're following best practices around Composer caching in Travis — we might be able to make these numbers even smaller.

tests/Stripe/SubscriptionItemTest.php
ob-stripe: Fixed!

tests/Stripe/TransferReversalTest.php
ob-stripe: Fixed too

ob-stripe:
I pushed another commit to fix the typos you noticed. Let me know if you'd like me to squash the commits before merging.

And yeah, on my laptop the entire test suite now runs under 1 second! Huge improvement over the O(minutes) it took before :)

brandur-stripe:
> And yeah, on my laptop the entire test suite now runs under 1 second! Huge improvement over the O(minutes) it took before :)

Wow, that's amazing!

> I pushed another commit to fix the typos you noticed. Let me know if you'd like me to squash the commits before merging.

Hm, I guess I have a slight preference for a squash here given that the current set of commits is a little unruly, but it doesn't make too much of a difference. Another possibility might be to try and rewrite the commit history into one that's still multiple commits, but shows a clean and clear progression (but it may not be worth the effort).

ob-stripe:
> Hm, I guess I have a slight preference for a squash here given that the current set of commits is a little unruly, but it doesn't make too much of a difference. Another possibility might be to try and rewrite the commit history into one that's still multiple commits, but shows a clean and clear progression (but it may not be worth the effort).

Yeah, I should have been more rigorous and done this as I worked on the PR. At this point, rewriting the history into multiple logical commits seems like more of a hassle than it's worth, so I just squashed into a single commit. Sorry @remi-stripe for stealing your work :)

ob-stripe:
@brandur-stripe I think this PR is ready to be merged unless you have any other comments?

brandur-stripe:
@ob-stripe Do it!

ob-stripe:
Wooooo \o/

PR #390, opened 2017-11-14
Remove unsupported API methods

ob-stripe:
r? @dpetrovics-stripe (DP run, feel free to re-assign!)

Removes two methods that exist in the PHP library but are not (and never were) supported by the API:
- `ApplicationFee::update()`
- `Source::all()`


grey-stripe:
LGTM

PR #391, opened 2017-11-14
Alphabetize list and add missing objects

ob-stripe:
r? @dpetrovics-stripe (DP run, feel free to re-assign!)

Alphabetize the list of API resources in `Util::convertToStripeObject()` and add missing resources:
- `application_fee`
- `balance`

(This was not a big issue as the PHP library does not rely on this method to instantiate resource classes when retrieving a resource, but we might as well be comprehensive.)

### DIFF
--- a/lib/Util/Util.php
+++ b/lib/Util/Util.php
@@ -65,11 +65,19 @@ abstract class Util
     public static function convertToStripeObject($resp, $opts)
     {
         $types = array(
+            // data structures
+            'list' => 'Stripe\\Collection',
+
+            // business objects
             'account' => 'Stripe\\Account',
             'alipay_account' => 'Stripe\\AlipayAccount',
             'apple_pay_domain' => 'Stripe\\ApplePayDomain',
-            'bank_account' => 'Stripe\\BankAccount',
+            'application_fee' => 'Stripe\\ApplicationFee',
+            'balance' => 'Stripe\\Balance',
             'balance_transaction' => 'Stripe\\BalanceTransaction',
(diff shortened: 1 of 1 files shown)


remi-stripe:
> This was not a big issue as the PHP library does not rely on this method to instantiate resource classes when retrieving a resource

Why didn't this properly work for ApplicationFee in that case? See my related commit https://github.com/stripe/stripe-php/pull/389/commits/64e909b5b8414ff054d1e8d2a40ce6ca52e23b4d which had to fix this to ensure we did get an `ApplicationFee` instance instead of `StripeObject`

ob-stripe:
> Why didn't this properly work for ApplicationFee in that case? See my related commit 64e909b which had to fix this to ensure we did get an ApplicationFee instance instead of StripeObject

Right, so the test that failed is `testIsListable`. The `all()` method _does_ rely on `Util::convertToStringObject()` to instantiate list elements as the proper resource class.

Before the fix, objects would be returned as `StripeObject`s. You were still able to access any attribute, e.g. this worked:

```php
$fees = \Stripe\ApplicationFee::all();
$fee = $fees->data[0];
$amount = $fee->amount;
```

However, you would not be able to use `ApplicationFee` methods, e.g. this would have failed:
```php
$fee->metadata["key"] = "value";
$fee->save();
```

Because `$fee` was a `StripeObject` and not an `ApplicationFee`.

remi-stripe:
Understood! I assumed that your PR description implied it was not an issue though but I see how limited the impact is ultimately.

grey-stripe:
LGTM

PR #393, opened 2017-11-20
Fix card source expiry date update

ob-stripe:
r? @dpetrovics-stripe (dev-platform run)
cc @stripe/api-libraries 

Fixes an issue where the library failed to update the expiry date when updating a card source.

### DIFF
--- a/lib/StripeObject.php
+++ b/lib/StripeObject.php
@@ -37,6 +37,7 @@ class StripeObject implements ArrayAccess, JsonSerializable
             'address',
             'address_kana',
             'address_kanji',
+            'card',
             'dob',
             'inventory',
             'legal_entity',
--- a/tests/SourceTest.php
+++ b/tests/SourceTest.php
@@ -137,6 +137,44 @@ class SourceTest extends TestCase
         $this->assertSame($source->owner['address']['country'], "Test Country");
     }
 
+
+    public function testSaveCardExpiryDate()
+    {
+        $response = array(
+            'id' => 'src_foo',
+            'object' => 'source',
+            'card' => array(
+                'exp_month' => 8,
+                'exp_year' => 2019,
+            ),
+        );
+        $source = Source::constructFrom(
+            $response,
+            new Util\RequestOptions()
+        );
+
+        $response['card']['exp_month'] = 12;
+        $response['card']['exp_year'] = 2022;
+        $this->mockRequest(
+            'POST',
+            '/v1/sources/src_foo',
+            array(
+                'card' => array(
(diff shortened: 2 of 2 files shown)


grey-stripe:
LGTM

lib/StripeObject.php
```diff
@@ -37,6 +37,7 @@ public static function init()
             'address',
             'address_kana',
             'address_kanji',
+            'card',
```
arcanedev-maroc: I think this PR is a breaking change because it replace `Stripe\Card` to `Stripe\AttachedObject` object.

Skipped: https://github.com/stripe/stripe-php/blob/master/lib/Util/Util.php#L81

Try to `var_dump($source->card)` to see the result.

ob-stripe: The `card` hash under source objects does not have an `"object" => "card"` entry (because it's not a card object), so `convertToStripeObject()` would not convert those hashes to `\Stripe\Card` objects but just to a generic `\Stripe\Object`.

You're correct that this patch did cause the type to be changed from `\Stripe\Object` to `\Stripe\AttachedObject`, which might be seen as a breaking change, but I have a hard time thinking of a practical case where that would break anything. What do you think?

arcanedev-maroc: Try to add this test in `TokenTest` class to see the issue:

```php

/** @test */
public function it_can_create_card_token()
{
    $token = self::createCardToken();

    $this->assertInstanceOf(Token::class, $token);
    $this->assertSame('card', $token->type);
    $this->assertInstanceOf(Card::class, $token->card);
    $this->assertSame('Visa', $token->card->brand);
}

/**
  * Create a card Token for tests
  *
  * @return Token
  */
private static function createCardToken()
{
    return Token::create([
        'card' => [
            'number'    => '4242424242424242',
            'exp_month' => date('n'),
            'exp_year'  => date('Y') + 1,
            'cvc'       => '314',
        ]
    ]);
}
```

ob-stripe: Ah, indeed, the `card` hash nested in _token_ objects is an actual card object, and this patch does cause the regression you described.

I'll try to push a fix shortly.

arcanedev-maroc: Can you try to remove/revert only [this line](https://github.com/stripe/stripe-php/blob/master/lib/StripeObject.php#L40) and see if the tests passes ?

ob-stripe: That would make your new test pass, but cause the test I added in this PR to fail (i.e. it would make it impossible to update expiry dates on card sources).

PR #396, opened 2017-11-30
[WIP] Ensure that `card` is only treated as a nested updatable attribute on Source objects

ob-stripe:
r? @deontologician-stripe (dev-platform run, feel free to reassign)
cc @stripe/api-libraries @arcanedev-maroc 

This fixes the issue described by @arcanedev-maroc [here](https://github.com/stripe/stripe-php/pull/393#discussion_r153878518).

Basically, in #393 I added `card` to the list of nested updatable attributes so that users can update the expiry date on card sources:

```php
$source = \Stripe\Source::create(array(
  "type" => "card",
  "token" => "tok_visa",
));
$source->card->exp_month = 12;
$source->card->exp_month = 2022;
$source->save();
```

This worked, but had the side-effect of turning all nested `card` hashes into `\Stripe\AttachedObject`. That's fine for sources since the `card` hash is just a hash, but not for token objects since the `card` hash nested in token objects is an actual card object, and so the instance should be a `\Stripe\Card` (even though I have trouble imagining a practical case where that would be an issue -- there's no reason I can think of to manipulate a card object via token rather than via a customer/recipient/account).

This PR changes the way the `$nestedUpdatableAttributes` works so that each class can define its own nested updatable attributes. `StripeObject` now uses [late static bindings](http://uk1.php.net/manual/en/language.oop5.late-static-bindings.php) so that derived classes can override the `getNestedUpdatableAttributes()` static function (which replaces the `$nestedUpdatableAttributes` static variable) to define its own list of nested updatable attributes.

### DIFF
--- a/lib/Source.php
+++ b/lib/Source.php
@@ -10,6 +10,20 @@ namespace Stripe;
 class Source extends ApiResource
 {
     /**
+     * @return Util\Set Attributes that are nested but still updatable from
+     *    the parent class's URL (e.g. metadata).
+     */
+    public static function getNestedUpdatableAttributes()
+    {
+        static $nestedUpdatableAttributes = null;
+        if ($nestedUpdatableAttributes === null) {
+            $nestedUpdatableAttributes = StripeObject::getNestedUpdatableAttributes();
+            $nestedUpdatableAttributes->add('card');
+        }
+        return $nestedUpdatableAttributes;
+    }
+
+    /**
      * @param array|string $id The ID of the source to retrieve, or an options
      *     array containing an `id` key.
      * @param array|string|null $opts
--- a/lib/StripeObject.php
+++ b/lib/StripeObject.php
@@ -13,45 +13,53 @@ use InvalidArgumentException;
 class StripeObject implements ArrayAccess, JsonSerializable
 {
     /**
-     * @var Util\Set Attributes that should not be sent to the API because
+     * @return Util\Set Attributes that should not be sent to the API because
      *    they're not updatable (e.g. API key, ID).
      */
-    public static $permanentAttributes;
+    public static function getPermanentAttributes()
+    {
+        static $permanentAttributes = null;
+        if ($permanentAttributes === null) {
+            $permanentAttributes = new Util\Set(array('_opts', 'id'));
+        }
+        return $permanentAttributes;
+    }
+
     /**
-     * @var Util\Set Attributes that are nested but still updatable from
+     * @return Util\Set Attributes that are nested but still updatable from
      *    the parent class's URL (e.g. metadata).
      */
-    public static $nestedUpdatableAttributes;
-
-    public static function init()
+    public static function getNestedUpdatableAttributes()
     {
-        self::$permanentAttributes = new Util\Set(array('_opts', 'id'));
-        self::$nestedUpdatableAttributes = new Util\Set(array(
-            // Numbers are in place for indexes in an `additional_owners` array.
-            //
-            // There's a maximum allowed additional owners of 3, but leave the
-            // 4th so errors work properly.
-            0, 1, 2, 3, 4,
+        static $nestedUpdatableAttributes = null;
+        if ($nestedUpdatableAttributes === null) {
+            $nestedUpdatableAttributes = new Util\Set(array(
+                // Numbers are in place for indexes in an `additional_owners` array.
+                //
+                // There's a maximum allowed additional owners of 3, but leave the
+                // 4th so errors work properly.
+                0, 1, 2, 3, 4,
 
-            'additional_owners',
-            'address',
-            'address_kana',
-            'address_kanji',
-            'card',
-            'dob',
-            'inventory',
-            'legal_entity',
-            'metadata',
-            'owner',
-            'payout_schedule',
-            'personal_address',
-            'personal_address_kana',
-            'personal_address_kanji',
-            'shipping',
-            'tos_acceptance',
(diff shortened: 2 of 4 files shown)


lib/StripeObject.php
```diff
@@ -13,45 +13,54 @@
...
+                'address_kana',
+                'address_kanji',
+                //'card',
+                'dob',
+                'inventory',
+                'legal_entity',
+                'metadata',
+                'owner',
+                'payout_schedule',
+                'personal_address',
```
remi-stripe: This lives under `legal_entity` on the `Account` object and `legal_entity` is not a real class/object in the SDK (nor the API). This means you can't define a custom `getNestedUpdatableAttributes` and make it work for some of those params. Would the plan be to "leave" it on `StripeObject` directly? If so, won't most people pushing PR just default to putting it here anyway? Seems hard to realize there are edge-cases when you do.

lib/Source.php
```diff
@@ -9,6 +9,20 @@
...
+    /**
+     * @return Util\Set Attributes that are nested but still updatable from
+     *    the parent class's URL (e.g. metadata).
+     */
+    public static function getNestedUpdatableAttributes()
+    {
+        static $nestedUpdatableAttributes = null;
+        if ($nestedUpdatableAttributes === null) {
+            $nestedUpdatableAttributes = StripeObject::getNestedUpdatableAttributes();
+            $nestedUpdatableAttributes->add('card');
```
remi-stripe: I worry that this will make it even harder to not miss edge-case. We already "break" the API often by forgetting to add those new API params to the whitelist. Now we need to add those but to potentially multiple objects now instead of just one canonical (though hidden) place in the library

ob-stripe: What this PR does is give the option to override the global whitelist in each class. If we're afraid that splitting up the whitelist will make it harder to maintain, we can simply keep the PR in its current form, i.e. keep the global whitelist on `StripeObject` and keep the one special case of `card` under `Source`.

remi-stripe: I understand but it _feels_ wrong/confusing.

lib/StripeObject.php
ob-stripe: The attributes defined here don't have to be top-level attributes in a given object, they can themselves be nested under other attributes (else the current code would not work either). So my intent was to move all of the identity verification attributes under `Account`, but as said above we can also keep the attributes in `StripeObject` by default and only override the global whitelist when needed.

deontologician-stripe:
I'm not familiar with this binding and my php is pretty rusty, is there someone better who could review this?

ob-stripe:
@brandur-stripe Can you take a look at this and share your thoughts?

lib/StripeObject.php
```diff
@@ -13,45 +13,53 @@
...
+     * @return Util\Set Attributes that should not be sent to the API because
      *    they're not updatable (e.g. API key, ID).
      */
-    public static $permanentAttributes;
+    public static function getPermanentAttributes()
+    {
+        static $permanentAttributes = null;
+        if ($permanentAttributes === null) {
+            $permanentAttributes = new Util\Set(array('_opts', 'id'));
+        }
```
brandur-stripe: Oh man. I was going to say that it would be nice to just move this up to a statically initialized expression so we don't have to check the `null` everytime, but just reading through the manual it seems like this only became supported in 5.6 for non-trivial expressions :/

lib/StripeObject.php
brandur-stripe: I'd be bullish in moving these under something like `Account` now that class-specific rules are supported. It's always seemed pretty janky to me that things like `0, 1, 2, 3, 4` and `additional_owners` sit at the top level.

lib/Source.php
```diff
@@ -9,6 +9,20 @@
  */
 class Source extends ApiResource
 {
+    /**
+     * @return Util\Set Attributes that are nested but still updatable from
+     *    the parent class's URL (e.g. metadata).
+     */
+    public static function getNestedUpdatableAttributes()
```
brandur-stripe: Could we add a comment somewhere in here that gives a breakdown similar to the one in this PR's description? I feel like this is a pretty difficult-to-understand complexity of the PHP library and I worry that the reasons we have `card` in here will be forgotten pretty quickly.

ob-stripe: 👍

lib/StripeObject.php
ob-stripe: Yep. I amended the commit so you can't see it, but my first submission used static initializations 😭

ob-stripe:
So I swore I posted an update on this issue, but it apparently got lost during last week's madness.

My approach does not work as well as I had hoped. What I said here is incorrect:

> The attributes defined here don't have to be top-level attributes in a given object, they can themselves be nested under other attributes (else the current code would not work either). So my intent was to move all of the identity verification attributes under Account, but as said above we can also keep the attributes in StripeObject by default and only override the global whitelist when needed.

This only works for top-level attributes. E.g. for `Account`, this would work for `legal_entity` but not for `additional_owners` and the `0`, `1`, etc. indexes. This is because the `legal_entity` hash will be cast as an `AttachedObject`, and its own nested attributes will then use `AttachedObject`'s whitelist rather than `Account`'s.

We could merge the PR as-is in order to fix the `card` issue, but I'm not too enthusiastic about it. I'm going to try rewriting the deserialization logic to see if we can get rid of the whitelist altogether.

ob-stripe:
Closing this one in favor of #404.

PR #397, opened 2017-12-02
Start testing PHP 7.2

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

PHP 7.2.0 has just been [released](http://php.net/releases/7_2_0.php), so we should start testing it in Travis.

The upgrade revealed a few bad tests where we were using `count()` on `StripeObject`s, which is [no longer supported](https://wiki.php.net/rfc/counting_non_countables) in PHP 7.2 (and returned `1` in previous versions, which likely wasn't the expected behavior when we wrote those tests). I've verified that the updated tests in the `ob-stripe-mock` branch all pass with PHP 7.2, and pushed a commit to fix the current tests here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,10 @@ matrix:
       env: AUTOLOAD=1
     - php: 7.1
       env: AUTOLOAD=0
+    - php: 7.2
+      env: AUTOLOAD=1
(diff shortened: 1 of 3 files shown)


brandur-stripe:
LGTM.

PR #401, opened 2017-12-13
Add total_count property to Collection

thePanz:


remi-stripe:
@thePanz `total_count` is a property that has been deprecated for a long time so I don't think we should merge this into the official repo to avoid having people rely on it.

thePanz:
@remi-stripe can you please be more specific?
The current API is returning such property, as an example:
`curl https://api.stripe.com/v1/customers/cus_Xxxxxxxx -u sk_xxxxxxxxxxxxxxxxxxxxx:`

Which is the correct property to use to fetch the total amount of items?

remi-stripe:
The `total_count` property is present on sub-collections but likely going away in the future. I would recommend not relying on this value.

Instead, you should rely on `has_more` on the collection to know if there are more elements than the ones in your current page and if so paginate through to get the full list and exact count.

PR #404, opened 2017-12-20
Fix parameter serialization logic

### LINKED ISSUE #402: Always return Stripe\Card when calling Stripe\Token->card
Currently, it returns an `AttachedObject`, while I think it should return a `Card`

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

This is a pretty beefy PR, sorry about that! I can probably split it up in a few smaller PRs.

Basically, this replaces the entire parameter serialization logic to make it as similar as possible to stripe-ruby's. This lets us get rid of the nested parameters whitelist (with a minor exception for `metadata`), so this PR fixes #402 and replaces #396.

While I expect most users will be able to upgrade without issues, this PR does introduce some breaking changes:
- `AttachedObject` no longer exists. All nested objects that are not a specific API resource are now `StripeObject`s (as is the case in stripe-ruby and stripe-python).
- `Collection` now derives from `StripeObject` instead of `ApiResource`. This is also the case in stripe-ruby, and is necessary to solve a serialization quirk when replacing a list object (e.g. when replacing `items` on a subscription object)

Tagged as WIP to avoid merging for now. Since this will require a major version bump, we might want to bundle some other breaking changes with it.

Let me know if you have any questions or if you want me to split the PR into smaller ones.


brandur-stripe:
Sorry for the delay on this one OB. I've done a partial review, and will get to the rest of it today (I think).

ob-stripe:
Since this is a breaking change, I've updated the base branch to be `integration-v6` so that we can solve potential conflicts once #410 is merged in.

ob-stripe:
I rebased the branch on `integration-v6` and set that as the merge target.

lib/Account.php
```diff
@@ -173,4 +185,45 @@ public static function createLoginLink($id, $params = null, $opts = null)
     {
         return self::_createNestedResource($id, static::PATH_LOGIN_LINKS, $params, $opts);
     }
+
+    public function serializeParameters($force = false)
```
brandur-stripe: Just OOC, is this mostly copied from the Ruby implementation? The test suite looks like it's pretty good in terms of vetting the new code, but just wondering how much we should expect to have to worry about it.

lib/ApiResource.php
```diff
@@ -9,11 +9,47 @@
...
+    ];
+
+    /**
+     * @var boolean A flag that can be set a behavior that will cause this
+     * resource to be encoded and sent up along with an update of its parent
+     * resource. This is usually not desirable because resources are updated
+     * individually on their own endpoints, but there are certain cases,
+     * replacing a customer's source for example, where this is allowed.
+     */
+    public $saveWithParent = false;
```
brandur-stripe: Somewhat too bad this is getting copied across — it felt a little janky as it was coming into Ruby and I was secretly hoping that there was a better way that I hadn't thought of :) It's okay though. At least similar implementations will make easier to reason about it if you're familiar with the other.

lib/ApiResource.php
```diff
@@ -93,6 +137,11 @@ public function instanceUrl()
         return static::resourceUrl($this['id']);
     }
 
+    /**
+     * @param array|null|mixed $params The list of parameters to validate
+     *
+     * @throws Stripe\Error\Api if $params exists and is not an array
+     */
```
brandur-stripe: Thanks for filling in this documentation!

lib/ApiResource.php
```diff
@@ -206,7 +301,7 @@ protected function _delete($params = null, $options = null)
      * @param array|null $params
      * @param array|string|null $options
      *
-     * @return StripeObject
+     * @return Stripe\StripeObject
```
brandur-stripe: Nice fixes too ...

lib/Collection.php
```diff
@@ -12,10 +12,47 @@
...
+     * @param array|string|null $options
+     *
+     * @return array tuple containing (the JSON response, $options)
+     */
+    protected function _request($method, $url, $params = [], $options = null)
+    {
+        $opts = $this->_opts->merge($options);
+        list($resp, $options) = static::_staticRequest($method, $url, $params, $opts);
+        $this->setLastResponse($resp);
+        return [$resp->json, $options];
```
brandur-stripe: Wow, you rebuilt for 5.3 already? Nice haha.

lib/StripeObject.php
```diff
@@ -205,81 +166,170 @@ public static function constructFrom($values, $opts)
...
+        // There are _some_ known exceptions though.
+        //
+        // For example, if the value is unsaved (meaning the user has set it), and
+        // it looks like the API resource is persisted with an ID, then we include
+        // the object so that parameters are serialized with a reference to its
+        // ID.
+        //
+        // Another example is that on save API calls it's sometimes desirable to
+        // update a customer's default source by setting a new card (or other)
+        // object with `#source=` and then saving the customer. The
```
brandur-stripe: The `#` designation for instance variables is fairly Ruby specific. Is there a more conventional way for PHP? (Maybe just `.`.)

lib/StripeObject.php
```diff
@@ -205,81 +166,170 @@ public static function constructFrom($values, $opts)
...
+        // `#save_with_parent` flag to override the default behavior allows us to
```
brandur-stripe: `save_with_parent` -> `saveWithParent`

tests/Stripe/AccountTest.php
```diff
@@ -179,4 +179,92 @@ public function testCanCreateLoginLink()
         $resource = Account::createLoginLink(self::TEST_RESOURCE_ID);
         $this->assertInstanceOf("Stripe\\LoginLink", $resource);
     }
+
+    public function testSerializeNewAdditionalOwners()
```
brandur-stripe: Thanks for adding all these tests. Account's serialization is problematic enough that you really can't have enough test coverage of it.

tests/Stripe/StripeObjectTest.php
```diff
@@ -147,38 +176,234 @@ public function testSerializeParametersOnArrayThatLengthens()
...
+            $optsReflector->getValue($copyValues["map"]["0"])
+        );
+        $this->assertNotSame(
+            $optsReflector->getValue($values["map"]["0"]),
+            $optsReflector->getValue($copyValues["map"]["0"])
+        );
+        $this->assertEquals($values["map"]["1"], $copyValues["map"]["1"]);
+        $this->assertEquals($values["map"]["2"], $copyValues["map"]["2"]);
+    }
+    public function testDeepCopyMaintainClass()
```
brandur-stripe: Maybe space out these function definitions a bit (should have a newline in there).

tests/Stripe/StripeObjectTest.php
```diff
@@ -147,38 +176,234 @@ public function testSerializeParametersOnArrayThatLengthens()
...
+        $this->assertEquals(
+            $optsReflector->getValue($values["map"]["0"]),
+            $optsReflector->getValue($copyValues["map"]["0"])
+        );
+        $this->assertNotSame(
+            $optsReflector->getValue($values["map"]["0"]),
+            $optsReflector->getValue($copyValues["map"]["0"])
+        );
+        $this->assertEquals($values["map"]["1"], $copyValues["map"]["1"]);
+        $this->assertEquals($values["map"]["2"], $copyValues["map"]["2"]);
```
brandur-stripe: Nit, but the whole section above is pretty dense. I like that we have good test coverage, but it could probably use some more whitespace for readability.

lib/Account.php
ob-stripe: Yep, it's basically a straight port from the Ruby implementation.

lib/ApiResource.php
ob-stripe: Yeah... I might have been cargo culting a little bit too much in this PR but I couldn't come up with anything better and I like that the implementations are consistent across languages for the reason you exposed.

lib/StripeObject.php
ob-stripe: I used `->source =`. I'm not sure how conventional it is but I think it's fairly explicit.

brandur-stripe:
LGTM.

ob-stripe:
@brandur-stripe I've squashed the existing commits and added a new one. As part of the changes to the serialization logic, `Collection` no longer derives from `ApiResource` but rather from `StripeObject`. This is necessary so `serializeParamsValue` works as expected, and is similar to stripe-ruby (https://github.com/stripe/stripe-ruby/blob/master/lib/stripe/list_object.rb#L2).

I initially duplicated the `_request` and `_staticRequest` methods, but now that we can use traits, this is no longer necessary.

Can you take a look at the latest commit? Thanks :)

init.php
```diff
@@ -36,10 +36,11 @@
 require(dirname(__FILE__) . '/lib/Error/OAuth/UnsupportedResponseType.php');
 
 // API operations
-echo "bouh";
```
ob-stripe: Leftover from manual debugging 😨

lib/Account.php
```diff
@@ -40,6 +40,10 @@ class Account extends ApiResource
     use ApiOperations\All;
     use ApiOperations\Create;
     use ApiOperations\Delete;
+    use ApiOperations\NestedResource;
+    use ApiOperations\Retrieve {
+        retrieve as protected _retrieve;
```
ob-stripe: Resources that have their own custom implementation of API methods but need to call the standard implementation (usually because they have some additional parameter checking) can use this syntax to "inherit" the trait methods under a different name.

init.php
brandur-stripe: Oops. Totally missed that apparently, haha.

lib/Account.php
brandur-stripe: That's pretty cool. Thanks for the ntoe!

brandur-stripe:
> I've squashed the existing commits and added a new one. As part of the changes to the serialization logic, Collection no longer derives from ApiResource but rather from StripeObject. This is necessary so serializeParamsValue works as expected, and is similar to stripe-ruby (https://github.com/stripe/stripe-ruby/blob/master/lib/stripe/list_object.rb#L2).
>
> I initially duplicated the _request and _staticRequest methods, but now that we can use traits, this is no longer necessary.

Makes sense. Thanks!

Alright, I had one more look over, and I think it's pretty solid. LGTM.

PR #410, opened 2018-01-04
Drop support for PHP 5.3

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

This PR drops support for PHP 5.3, and gets rid of some now unneeded stuff.

Note that the destination branch is `integration-v6` instead of `master` so that we can merge all PRs going into stripe-php 6.0.0 and solve conflicts before merging into `master`.

Fixes #406, although the fix won't be effective until we merge into `master` and actually release a new major version.

PR #412, opened 2018-01-04
stripe-php 6.0.0

### LINKED ISSUE #394: Request retries on failure
Hello,

I was wondering if a request is retried in case of failure. I read about [idempotency recommendations](https://stripe.com/blog/idempotency) :

> The Stripe Ruby library retries on failure automatically with an idempotency key using increasing backoff times and jitter

Does stripe-php has a similar feature ? If not, would it be interesting to implement it ?

### LINKED ISSUE #402: Always return Stripe\Card when calling Stripe\Token->card

### LINKED ISSUE #406: Deprecate `Stripe\JsonSerializable` and bump minimum version to PHP 5.4 (or later)
PHP version 5.3 is at End of Life cycle for at least 3 years now. The file `./lib/JsonSerializable.php` as a bridge for the `\JsonSerializable` interface prevents execution of the AspectMock library for testing, apart from not being PSR-2 compliant.

[AspectMock](https://github.com/Codeception/AspectMock) library from Codeception is a powerful mocking library we use to test static methods and mock objects but it parses PHP files on their own and doesn't recognise the special syntax used (`interface` declarations must be at root).

This issue is a suggestion to make the current version usable much like the v1.18.0 version for PHP 5.2, as specified in [Legacy Version Support](https://github.com/stripe/stripe-php/#legacy-version-support).

ob-stripe:
- [x] Drop support for PHP 5.3 (#410)
- [x] Use traits for common API operations (#411)
- [x] Use short array syntax (#414)
- [x] Fix serialization logic (#404)
- [x] Get rid of ExternalAccount (#417)
- [x] Add more tests (#418)
- [x] Update CA bundle and add script for future updates (#421)
- [x] Use vendored CA bundle for all requests (#422)
- [x] Automatic request retries (#428)

Fixes #394, fixes #402, fixes #406.


ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Alright, I think we're ready to release all these updates! \o/

You've already reviewed each PR individually, so maybe just take a quick look to make sure we've not missed anything.

brandur-stripe:
Woot! Really excellent coordination for getting this release out the door Olivier!

I didn't do an in-depth review, but did take a quick scan of the entire diff — nothing jumped out. Go for it!

PR #414, opened 2018-01-04
Use short array syntax everywhere

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @arcanedev-maroc 

Now that we're dropping support for <5.4, we can start using the short array syntax.

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -33,7 +33,7 @@ class ApiRequestor
         } elseif ($d === false) {
             return 'false';
         } elseif (is_array($d)) {
-            $res = array();
+            $res = [];
             foreach ($d as $k => $v) {
                 $res[$k] = self::_encodeObjects($v);
             }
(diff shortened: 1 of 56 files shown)


lib/ApiRequestor.php
```diff
@@ -54,17 +54,13 @@ private static function _encodeObjects($d)
...
     public function request($method, $url, $params = null, $headers = null)
     {
-        if (!$params) {
-            $params = array();
-        }
-        if (!$headers) {
-            $headers = array();
-        }
+        $params = $params ?: [];
+        $headers = $headers ?: [];
```
ob-stripe: I opportunistically did a few of these "fixes" which is why the PR has a negative net line count.

brandur-stripe: +1.

brandur-stripe:
Soooooo much cleaner! LGTM. (Feel free to go for a merge when this is ready.)

I restarted a couple of failed builds in the matrix. They look like intermittent problems talking to httpbin:

> (Network error [errno 7]: Failed to connect to httpbin.org port 443: Connection timed out)

PR #415, opened 2018-01-05
Use assertInstanceOf where possible

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Just a small improvement to make the tests slightly cleaner: use `assertInstanceOf($class_name, $instance)` instead of `assertSame($class_name, get_class($instance))`.


brandur-stripe:
LGTM.

PR #417, opened 2018-01-09
Get rid of ExternalAccount

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Okay, I'm not 100% sure we actually want this change, but since we're going to release a new major version I thought I'd open the PR to gather feedback.

This PR removes the `ExternalAccount` class entirely. The existing implementation has a number of flaws:
- generic `instanceUrl()` method that assumes every external account can be attached to a customer, an account or a recipient
- generic `verify()` method that assumes every external account can be verified
    - in practice, only bank accounts can be verified, and `BankAccount` has its own `verify()` method
- `Source` does not derive from `ExternalAccount`

This change makes the implementation closer to what we have in stripe-ruby and stripe-python: all "external accounts" (namely: cards, bank accounts, sources, Alipay accounts and Bitcoin receivers) derive directly from `APIResource` and have their own `instanceUrl()` method that knows which type of owner each resource can have (e.g. bank accounts cannot be attached to recipients, etc.), plus custom `retrieve()` and `update()` methods that return an helpful error message.


brandur-stripe:
Great implementation (with nice test additions!).

I think others would have a better idea as to whether this is a desirable change or not, but the reduction in special casing seems like a good idea to me in concept at least. +1.

lib/Account.php
```diff
@@ -120,7 +120,7 @@ public function deauthorize($clientId = null, $opts = null)
      * @param array|null $params
      * @param array|string|null $opts
      *
-     * @return ExternalAccount
+     * @return ApiResource
      */
     public static function createExternalAccount($id, $params = null, $opts = null)
```
remi-stripe: Should re remove those methods explicitly in favor of `$account->external_accounts->create()` instead? This avoids the confusion and we would automatically return the right class based on the type of external accounts

lib/AlipayAccount.php
```diff
@@ -7,7 +7,60 @@
  *
  * @package Stripe
  */
-class AlipayAccount extends ExternalAccount
+class AlipayAccount extends ApiResource
```
remi-stripe: Where are we on the idea that we should remove those classes entirely since we don't support them anymore and have not for over a year? Since PHP is dynamic and would return an `ApiResource` anyway when you get one of those on a Charge for example, is it time to get rid of it at the same time?

lib/Card.php
```diff
@@ -31,7 +31,62 @@
...
+    /**
+     * @param array|string $_id
+     * @param array|string|null $_opts
+     *
+     * @throws \Stripe\Error\InvalidRequest
+     */
+    public static function retrieve($_id, $_opts = null)
+    {
+        $msg = "Cards cannot be accessed without a customer, recipient or account ID. " .
+               "Retrieve a bank account using \$customer->sources->retrieve('card_id'), " .
```
remi-stripe: s/bank account/card/

lib/Account.php
ob-stripe: No, these methods let users manipulate external accounts without having to retrieve the account first (like other nested resource methods). Also, to clarify, these methods _will_ return the correct type -- though of course we cannot know what this type will be in the PHPDoc.

lib/Card.php
ob-stripe: 👍 Thanks!

lib/AlipayAccount.php
ob-stripe: I think we can keep the classes around -- the definitions do still exist in the OpenAPI spec, and in the case of Bitcoin some users do still send requests to the top-level endpoint `/v1/bitcoin/receivers`.

I've removed the `refund()` method from `BitcoinReceiver` though, and updated the PHPDoc to include links to the sources API documentation.

ob-stripe:
Okay, since the consensus seems to be in favor, I'm going to pull this in the integration branch for v6.

PR #418, opened 2018-01-11
Add more tests

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Adds a bunch of tests for previously untested codepaths.

Most of the remaining untested parts are overly defensive programming that we could arguably get rid of (e.g. handling the absence of an `id` attribute in non-singleton resources).

There are also some untested parts in `CurlClient`, mostly error conditions (connection failures, certificate verification failures, etc.). Those are not so easy to test unfortunately.


brandur-stripe:
Hard to argue with this PR!

LGTM.

Issue #124, opened 2015-01-12
Refactor to not use static method calls


atrauzzi:
In many applications (especially where dependency injection is used), it's preferable to not have libraries that rely on global state.

The Stripe PHP library currently takes the approach that every call is done via a static method.  While it's fine to retain this as an option, it would be nice if actual objects could be leveraged.

GrahamCampbell:
:+1:

joshhornby:
(Y)

Rican7:
:+1:

skyzyx:
:+1:

browner12:
would it be possible to maintain both as an option on the same code? I feel like that would become unmanageable quickly. would it be better to break it off into another repo, or would that be just as difficult to maintain?

maybe the better question is, *should* the static approach be kept, or maybe deprecated? does someone have an example where it would be difficult to instantiate actual objects?

either way :+1:

ruudk:
:+1:

yurytolochko:
:+1:

phillipsnick:
:+1:  

Having fun trying to unit test my payment service because I can't mock static methods!

armetiz:
For people trying to unit-test payment, you should make a wrapper around Stripe.

```php
use Acme\Payment\Stripe

class Stripe
{
  static private $instance;

  public function __construct($stripeSecretKey)
  {
        if(self::$instance) {
            throw new \RuntimeException('Stripe could not be instantiate more than once. Check PHP implementation : https://github.com/stripe/stripe-php');
        }

        self::$instance = $this;

        \Stripe\Stripe::setApiKey($stripeSecretKey);
  }

  public function payAmountWithToken($token, $amount)
  {
    \Stripe\Charge::create([]); // use correct values
  }
}
```

I made a singleton to avoid confusion. This is a limitation but in many case you will not need to have more than one instance.

You will be able to inject your custom wrapper / mock and make your unit-test ;)

Regards.

bakura10:
As always, for people that do not want to use static methods, you could have a look at my library [ZfrStripe](https://github.com/zf-fr/zfr-stripe). It is maintained, is based on Guzzle and is definitely easier to mock and test. Having said that it may be a bit harder to use as you definitely need to inject the client when you need it instead of relying on the staticness.

arcanedev-maroc:
:+1:

adrienbrault:
:cry:

KostyaChoporov:
[...]

PR #419, opened 2018-01-16
Use AspectMock in tests

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Adds support for the [AspectMock](https://github.com/Codeception/AspectMock) library in tests.

AspectMock allows for some things that PHPUnit itself cannot handle, most notable mocking/stubbing static methods. Since we use a lot of static methods throughout the library, we will be able to use AspectMock to write better tests.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,6 +22,7 @@
   },
   "require-dev": {
     "phpunit/phpunit": "~4.0",
+    "codeception/aspect-mock": "~0.5.5",
     "satooshi/php-coveralls": "~0.6.1",
     "squizlabs/php_codesniffer": "~2.0"
   },
--- a/phpunit.no_autoload.xml
+++ b/phpunit.no_autoload.xml
@@ -1,4 +1,4 @@
-<phpunit bootstrap="tests/bootstrap.no_autoload.php" colors="true">
+<phpunit bootstrap="tests/bootstrap.no_autoload.php" backupGlobals="false" colors="true">
     <testsuites>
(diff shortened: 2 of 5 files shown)


ob-stripe:
Argh. Looks like AspectMock (or rather, the Go! AOP framework dependency) does not support HHVM: https://github.com/goaop/framework/issues/143#issuecomment-306840301.

brandur-stripe:
Doh!

@ob-stripe I left some messages on Slack, but it does seem like unfortunately cross compatibility between PHP and HHVM is not long for this world, even if it's the libraries that are more intertwined with internals (i.e., like a mocking library) that need to drop compatibility first.

Let me know what you think, but I have a slight preference for keeping HHVM support around until it starts to cause more acute pain (and hopefully until after most other projects have dropped it). This would unfortunately mean foregoing more advanced mocking libraries for the time being.

The other alternative is that instead of trying to improve our testing, we could start refactoring in the direction of non-static methods. This would be a little too much API churn probably, but if we were starting from zero, it'd still probably be the superior design.

ob-stripe:
> The other alternative is that instead of trying to improve our testing, we could start refactoring in the direction of non-static methods. This would be a little too much API churn probably, but if we were starting from zero, it'd probably be the superior design.

Definitely agreed, though it will very likely require some breaking changes. (Though if we ever are to tackle #124, there's probably no way around it anyway.)

Anyway, closing for now. As much as I'd have liked to use AspectMock for testing some new features, I'll manage without it.

PR #422, opened 2018-01-17
Use vendored CA bundle for all requests

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Currently, the library sends requests using the environment's default CA bundle. If the request fails due to an SSL verification error, the library will automatically retry the request using the vendored bundle.

This PR changes the behavior to use the vendored bundle by default for all requests, and adds an option to configure the path to a custom bundle too. This makes the PHP implementation consistent with the Ruby implementation.

While this should not have any user impact, I'm tagging it as WIP because I'm slightly nervous about this change given the number of TLS issues we've observed in the past. Part of me is definitely like "if it ain't broke don't fix it" 😐 

For context, I'm working on adding request retries as requested in #394 and wanted to simplify the logic in order to avoid having two different retry mechanisms (one for cert verification issues and one for other connection/409 issues).

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,15 @@ $curl = new \Stripe\HttpClient\CurlClient([CURLOPT_SSLVERSION => CURL_SSLVERSION
 \Stripe\ApiRequestor::setHttpClient($curl);
 ```
 
+### Configuring CA Bundles
+
+By default, the library will use its own internal bundle of known CA
+certificates, but it's possible to configure your own:
+
+```php
+\Stripe\Stripe::setCABundlePath("path/to/ca/bundle");
+```
+
 ## Development
 
 Install dependencies:
--- a/lib/HttpClient/CurlClient.php
+++ b/lib/HttpClient/CurlClient.php
@@ -180,31 +180,13 @@ class CurlClient implements ClientInterface
         $opts[CURLOPT_TIMEOUT] = $this->timeout;
         $opts[CURLOPT_HEADERFUNCTION] = $headerCallback;
         $opts[CURLOPT_HTTPHEADER] = $headers;
-        if (!Stripe::$verifySslCerts) {
+        $opts[CURLOPT_CAINFO] = Stripe::getCABundlePath();
+        if (!Stripe::getVerifySslCerts()) {
             $opts[CURLOPT_SSL_VERIFYPEER] = false;
         }
 
         curl_setopt_array($curl, $opts);
         $rbody = curl_exec($curl);
-
-        if (!defined('CURLE_SSL_CACERT_BADFILE')) {
-            define('CURLE_SSL_CACERT_BADFILE', 77);  // constant not defined in PHP
-        }
-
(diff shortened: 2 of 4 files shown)


brandur-stripe:
@ob-stripe 

I think what you have here makes sense. There are some pretty complex tradeoffs that are a little bit hazy — for example, if a root CA were to be revoked, is it more likely that we'd get the update out to our libraries first or that a user's system bundle would get updated first? The answer is very unclear because we have so little control. Even if we automated system that would rebundle certificates periodically, people wouldn't necessarily upgrade.

All else being equal, I'd probably lean toward _not_ bundling because at least some systems will have good package managers to help keep the certificate bundle more up-to-date than we likely would (e.g., Archlinux), and users with a bundle so outdated that they can't connect to Stripe could be instructed on how to download and set a fresh one. That said, given that we're already bundling in Ruby, I think it makes sense to try and built out a consistent approach, so I'd say go for it.

I like that we give security-minded users who want to specify their own constrained bundle a way to do so through configuration.

> While this should not have any user impact, I'm tagging it as WIP because I'm slightly nervous about this change given the number of TLS issues we've observed in the past. Part of me is definitely like "if it ain't broke don't fix it" 😐

I can't personally think of a way that this could break people, but yeah, it's something we should keep in mind.

ob-stripe:
Alright, let's go for it. Thanks @brandur-stripe !

EatonZ:
Out of curiosity, did you ever give any thought to checking if the user has specified a CA file via php.ini, and use that?

Right now I'm using this code:
`\Stripe\Stripe::setCABundlePath(ini_get('openssl.cafile'));`

It would be cool if you detected that ini setting in this library and use that by default instead.

If you decide to go for that, please note there's actually 2 ini settings you should check. [The php source shows how this works.](https://github.com/php/php-src/blob/47699a24e6fe9b9f2654d4f2537cfdb7c089b27f/ext/curl/interface.c#L1865)

ob-stripe:
@EatonZ Interesting! I didn't know about those php.ini settings.

I think we should still use the vendored CA bundle by default though, as we know it to be compatible with Stripe's API.

We could add a helper method to make it slightly easier to use the CA bundle specified in php.ini settings, but it's fairly trivial to do anyway so I don't think that would provide much value.

tobiastom:
@ob-stripe We encountered some problems because of the usage of `realpath` inside the default bundle file method `getDefaultCABundlePath`. We deploy our software as a phar archive, so the realpath call fails.

Any changes this will be changed in a future version, or should we use the `setCABundlePath` function with our custom path?

tobiastom:
@ob-stripe After writing it I realised myself that the path inside the phar would not work. 

We now configure the global CA bundle before we use any of the Stripe functionality. This means it's not updated with the Stripe dependency, but with the one from the operating system. 

It seems to work quite well for us and I don't think there is anything to improve this. Maybe this reference helps someone else in the future. :)

Thanks for your help @ob-stripe!

PR #423, opened 2018-01-17
Escape unsanitized input in OAuth example

brandur:
Here we run unsanitized input through `htmlspecialchars` before echoing
it to screen. Most of these fields will come from our own processes, but
because they're taken from `$_GET`, they could potentially contain
content set by a malicious user.

I think `htmlspecialchars` is the right thing to do here. It just
escapes HTML-related characters instead of anything that has an HTML
equivalent like `htmlentities` (which is probably overkill here).

r? @ob-stripe


brandur-stripe:
Thanks OB.

PR #430, opened 2018-02-07
Updated @property tags to reflect API version 2018-02-05

nickdnk:
Guess who's back (already).

Changes were made in API 2018-02-05 to Product and Plan objects. I've updated the phpdocs to reflect this.

I also cleaned up a few cases of array vs mixed, no biggie though.


nickdnk:
Oh I didn't know about this tab/space indent rule. I'll fix this.

PR #435, opened 2018-02-11
Fix header persistence on Collection objects

### LINKED ISSUE #434: PHP Error:  Access to undeclared static property: Stripe\Collection::$HEADERS_TO_PERSIST in /app/vendor/stripe/stripe-php/lib/ApiOperations/Request.php on line 58
php: PHP 7.1.14-1+ubuntu16.04.1+deb.sury.org+1 (cli)

stripe: v6.0.0

the error happened when trying to add a card to an existing customer with 

`$src = $customer->sources->create(['source' => 'tok_visa']);`

if it makes any difference i am using connect and the customer in question is part of a "custom" account

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @rs-sliske

Fixes #434.

In 6.0.0, `Collection` is no longer a subclass of `ApiResource` (cf. #404). Because the list of persistent headers was stored on `ApiResource`, this broke API requests on collection objects that used headers.

Since we have a dedicated class in stripe-php to store request options (`Util\RequestOptions`), I've moved the header persistence logic there.

### DIFF
--- a/lib/ApiOperations/Request.php
+++ b/lib/ApiOperations/Request.php
@@ -54,11 +54,7 @@ trait Request
         $opts = \Stripe\Util\RequestOptions::parse($options);
         $requestor = new \Stripe\ApiRequestor($opts->apiKey, static::baseUrl());
         list($response, $opts->apiKey) = $requestor->request($method, $url, $params, $opts->headers);
-        foreach ($opts->headers as $k => $v) {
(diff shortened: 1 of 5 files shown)


brandur-stripe:
Nice fix, and great tests as usual!

LGTM.

PR #436, opened 2018-02-12
Introduce new `Idempotency` error class

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Adds support for the `idempotency_error` error type. Cf. https://github.com/stripe/stripe-ruby/pull/613.

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -70,6 +70,7 @@ class ApiRequestor
      * @param array $resp
      *
      * @throws Error\InvalidRequest if the error is caused by the user.
+     * @throws Error\Idempotency if the error is caused by an idempotency key.
      * @throws Error\Authentication if the error is caused by a lack of
      *    permissions.
      * @throws Error\Permission if the error is caused by insufficient
@@ -106,6 +107,7 @@ class ApiRequestor
         $msg = isset($errorData['message']) ? $errorData['message'] : null;
         $param = isset($errorData['param']) ? $errorData['param'] : null;
         $code = isset($errorData['code']) ? $errorData['code'] : null;
+        $type = isset($errorData['type']) ? $errorData['type'] : null;
 
         switch ($rcode) {
             case 400:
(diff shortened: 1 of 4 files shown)


lib/ApiRequestor.php
```diff
@@ -114,6 +116,9 @@ private static function _specificAPIError($rbody, $rcode, $rheaders, $resp, $err
                 if ($code == 'rate_limit') {
                     return new Error\RateLimit($msg, $param, $rcode, $rbody, $resp, $rheaders);
                 }
+                if ($type == 'idempotency_error') {
```
brandur-stripe: Sorry about the usual pedantry, but it kind of looks like the other conditional ladders for error interpretation in this file are alphabetized, and it might be worth giving this the same treatment. (Also, might want to change this to a `switch` like the others? It doesn't matter that much though.)

brandur-stripe:
Nice one @ob-stripe, and thanks for taking! My bad for not going through and adding these in when the server changes appeared.

lib/ApiRequestor.php
ob-stripe: Do you mean grouping `rate_limit` and `idempotency_error` in a `switch` statement? The problem with that is that the two values are in different variables (`$code` vs `$type`).

I've taken a look at how we handle this in other libraries: stripe-ruby completely dropped support for 400 `rate_limit` errors, which maybe we should do in other libraries as well, but that would be a breaking change and we _just_ released 6.0.0 :/

stripe-python uses a series of semi-complex if statements: https://github.com/stripe/stripe-python/blob/master/stripe/api_requestor.py#L190-L214, but then again Python doesn't have `switch`.

I'm tempted to update the logic to something like stripe-python anyway, wdyt?

brandur-stripe: Oops, my mistake — I wasn't paying enough attention to notice the `$code` vs. `$type` distinction there. Feel free to leave it.

> I'm tempted to update the logic to something like stripe-python anyway, wdyt?

IMO the logic ladders are at least pretty similar, and honestly PHP's is quite a bit more readable, so it might be okay to just leave as is. Up to you :)

ob-stripe: Yeah, let's leave it as is then. Thanks!

PR #438, opened 2018-02-14
Update changelog with list of BC breaks

### LINKED ISSUE #437: Release notes
Thanks for making this library available, it's really nice to not have to deal with the API directly.

With the release of 6.0.0 I was having some unanticipated troubles. There were a couple of backward incompatible changes introduced, which were not documented in either the release notes or the changelog. There's a list of relevant PRs, but it's a lot of manual work to go through all of them hunting for changes that might affect you.

Also, the release notes point to the changelog for more information, but it has the exact same information as the release notes, so it's not helpful point people there at all.

I would propose the following in future releases:
1. On a major release (like 6.0.0), the release notes state why there was a major version jump
2. Release notes stipulate if a BC break is introduced or not, along with a general description of the changes in that specific release
3. The changelog has details about what exactly changed, and how to upgrade in the event of a BC break
4. In the event of a large BC break, and probably with the release of a major version, a separate UPGRADE.md file is updated, where people can find how to upgrade from each (major) version to the next. This also helps people who are still an extra major version behind.

What do you think?

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @pascaldevink

Adds a list of backwards incompatible changes to watch out for following the 6.0 release.

Fixes #437.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@
 * [#436](https://github.com/stripe/stripe-php/pull/436) Introduce new `Idempotency` error class
 
 ## 6.0.0 - 2018-02-07
+Major version release. List of backwards incompatible changes to watch out for:
++ The minimum PHP version is now 5.4.0. If you're using PHP 5.3 or older, consider upgrading to a more recent version.
+* `\Stripe\AttachedObject` no longer exists. Attributes that used to be instances of `\Stripe\AttachedObject` (such as `metadata`) are now instances of `\Stripe\StripeObject`.
++ Attributes that used to be PHP arrays (such as `legal_entity->additional_owners` on `\Stripe\Account` instances) are now instances of `\Stripe\StripeObject`, except when they are empty. `\Stripe\StripeObject` has array semantics so this should not be an issue unless you are actively checking types.
+* `\Stripe\Collection` now derives from `\Stripe\StripeObject` rather than from `\Stripe\ApiResource`.
+
+Pull requests included in this release:
 * [#410](https://github.com/stripe/stripe-php/pull/410) Drop support for PHP 5.3
 * [#411](https://github.com/stripe/stripe-php/pull/411) Use traits for common API operations
 * [#414](https://github.com/stripe/stripe-php/pull/414) Use short array syntax


CHANGELOG.md
```diff
@@ -5,6 +5,13 @@
 * [#436](https://github.com/stripe/stripe-php/pull/436) Introduce new `Idempotency` error class
 
 ## 6.0.0 - 2018-02-07
+Major version release. List of backwards incompatible changes to watch out for:
++ The minimum PHP version is now 5.4.0. If you're using PHP 5.3 or older, consider upgrading to a more recent version.
+* `\Stripe\AttachedObject` no longer exists. Attributes that used to be instanced as `\Stripe\AttachedObject` (such as `metadata`) are now instanced as `\Stripe\StripeObject`.
```
brandur-stripe: I might use "used to be instances of `\Stripe\AttachedObject`" instead of "instanced" — the latter is a little unusual and I found it a little hard to parse.

CHANGELOG.md
```diff
@@ -5,6 +5,13 @@
...
++ Attributes that used to instanced as PHP arrays (such as `legal_entity->additional_owners` on `\Stripe\Account` instances) are now instanced as `\Stripe\StripeObject`, except when they are empty. `\Stripe\StripeObject` has array semantics so this should not be an issue unless you are actively checking types.
```
brandur-stripe: And I might do the same here. There's also a missing "be" in "Attributes that used to be instanced".

brandur-stripe:
Great idea Olivier! Added a couple minor nits above, but otherwise LGTM.

CHANGELOG.md
ob-stripe: I can't English :coneofshame: I think "instantiated" was what I was looking for, but your version is better 👍

ob-stripe:
Made the updates, ptal @brandur-stripe

CHANGELOG.md
```diff
@@ -5,6 +5,13 @@
 * [#436](https://github.com/stripe/stripe-php/pull/436) Introduce new `Idempotency` error class
 
 ## 6.0.0 - 2018-02-07
+Major version release. List of backwards incompatible changes to watch out for:
++ The minimum PHP version is now 5.4.0. If you're using PHP 5.3 or older, consider upgrading to a more recent version.
+* `\Stripe\AttachedObject` no longer exists. Attributes that used to be instances if `\Stripe\AttachedObject` (such as `metadata`) are now instances of `\Stripe\StripeObject`.
```
brandur-stripe: One more small one: "that used to be instance if" should be "of" instead of "if" :)

ob-stripe: 🤦‍♂️  Fixed.

brandur-stripe:
LGTM. Thanks OB.

PR #439, opened 2018-02-15
Document automatic retries

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Features are nice, documented features are nicer!


brandur-stripe:
Good call!

LGTM.

PR #440, opened 2018-02-16
Add support for /v1/topups endpoints

jkakar-stripe:
This add standard retrieve, create and update client support for the new `/v1/topups` endpoint.

r? @apakulov-stripe @ccontinanza-stripe @chellman-stripe @kenneth-stripe @miguel-stripe
r? @stripe/api-libraries

### DIFF
--- a/lib/Topup.php
+++ b/lib/Topup.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Stripe;
+
+/**
+ * Class Topup
+ *
+ * @property string $id
+ * @property string $object
+ * @property int $amount
+ * @property string $balance_transaction
+ * @property int $created
+ * @property string $currency
+ * @property string $description
+ * @property int $expected_availability_date
+ * @property string $failure_code
+ * @property string $failure_message
+ * @property bool $livemode
+ * @property StripeObject $metadata
+ * @property mixed $source
+ * @property string $statement_descriptor
+ * @property string $status
+ *
+ * @package Stripe
+ */
+class Topup extends ApiResource
+{
+    use ApiOperations\All;
+    use ApiOperations\Create;
+    use ApiOperations\Retrieve;
+    use ApiOperations\Update;
+}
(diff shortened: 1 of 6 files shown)


lib/Topup.php
```diff
@@ -0,0 +1,31 @@
...
+ * @property int $created
+ * @property string $currency
+ * @property string $description
+ * @property string $failure_code
+ * @property string $failure_message
+ * @property bool $livemode
+ * @property mixed $source
+ * @property string $statement_descriptor
+ * @property string $status
+ * @property string $balance_transaction
```
brandur-stripe: Mind alphabetizing this list (although keep `$id` and `$object` on top) and confirming that you went over it and looks correct? It looks okay to me, but I happened to notice that `$description` is still marked as undocumented so maybe we don't intend to expose that right away? Thanks!

jkakar-stripe: Hmm, I've updated the properties to reflect what should be exposed. I'm not sure I understand what you mean by `$description` being undocumented? All the fields listed now are public and should be known to users of the client.

brandur-stripe: Thanks!

If you check the API resource in the server-side implementation, it still seems that `description` is marked as undocumented:

``` ruby
undocumented :description
```

Maybe just an oversight?

brandur-stripe: Just to close the loop here, it looks like `description` is now an officially documented property.

brandur-stripe:
LGTM.

brandur-stripe:
Released as 6.2.0.

EatonZ:
What is a topup? Nothing in [the API docs](https://stripe.com/docs/api) about it besides a few event types.

remi-stripe:
@EatonZ It's not documented yet and we just planned the feature in the libraries for now.

EatonZ:
@remi-stripe Just a heads up the documentation is still missing.

remi-stripe:
@EatonZ This is totally expected. This feature is not public for now and only available in private beta for specific users. It will be released in the next few months.

PR #441, opened 2018-02-16
Incorrect phpdoc for SignatureVerification error

nickdnk:
I am back with more phpdoc edits :)

### DIFF
--- a/lib/WebhookSignature.php
+++ b/lib/WebhookSignature.php
@@ -17,8 +17,8 @@ abstract class WebhookSignature
      * @param string $secret secret used to generate the signature.
      * @param int $tolerance maximum difference allowed between the header's
      *  timestamp and the current time
-     * @throws \Stripe\Error\SignatureVerification if the verification fails.
-	 * @return bool
+     * @throws \Stripe\Error\SignatureVerification if the verification fails
+     * @return bool
      */
     public static function verifyHeader($payload, $header, $secret, $tolerance = null)
     {
--- a/lib/WebhookSignature.php
+++ b/lib/WebhookSignature.php
@@ -17,7 +17,8 @@ abstract class WebhookSignature
      * @param string $secret secret used to generate the signature.
      * @param int $tolerance maximum difference allowed between the header's
      *  timestamp and the current time
-     * @throws SignatureVerification if the verification fails.
+     * @throws \Stripe\Error\SignatureVerification if the verification fails.
+	 * @return bool
      */
     public static function verifyHeader($payload, $header, $secret, $tolerance = null)
     {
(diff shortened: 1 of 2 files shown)


lib/WebhookSignature.php
```diff
@@ -17,7 +17,8 @@
      * @param string $secret secret used to generate the signature.
      * @param int $tolerance maximum difference allowed between the header's
      *  timestamp and the current time
-     * @throws SignatureVerification if the verification fails.
+     * @throws \Stripe\Error\SignatureVerification if the verification fails.
+	 * @return bool
```
brandur-stripe: Mind correcting the tabs versus spaces problem here?

brandur-stripe:
Nice fix. Thank you.

I left a minor comment. After that's fixed, would you mind rebasing just so we have a clean commit to bring in. Thanks!

lib/WebhookSignature.php
nickdnk: I just tried. I don't know why that happened. it looked right in my IDE.

nickdnk:
I don't know how to do that. I am not really great with git :(

lib/WebhookSignature.php
brandur-stripe: It's usually a tabwidth problem (i.e., how many spaces your tabs are set to display as like 4 versus 8). Your width is shorter in your IDE so it looks okay, but the difference is visible once you get to GitHub where tabs are rendered differently.

nickdnk:
I'm gonna delete this and start over :D

brandur-stripe:
> I don't know how to do that. I am not really great with git :(

It's fairly easy so want to give it a quick shot?

Run this from your branch: `git rebase -i master`

Your `$EDITOR` should open (like Vim, Nano, or whatever), and you'll be presented with a list of commits that you've made on top of master like:

```
pick 6d3f940ddb8 Commit1
pick 796ed2b3d68 Commit2
pick 4bbd65b63a8 Commit3
```

All you need to do is change all the commits except the first one from `pick` to `f` so that they get just get wrapped up with the first one like this (`f` is for `fixup` which just brings it into its parent):

```
pick 6d3f940ddb8 Commit1
f 796ed2b3d68 Commit2
f 4bbd65b63a8 Commit3
```

If you want, you can also change the first `pick` to `edit` and it'll give you a chance to change the commit message. After making those changes, save and exit your editor, and the rebase will happen.

brandur-stripe:
> I'm gonna delete this and start over :D

Ah, k. Without being to prescriptive, I'd recommend try to develop at least a passing ability to do a basic rebase like this one. It's not as hard as it sounds, and I think you'll find that quite a few open source projects will request it.

nickdnk:
Thanks for the guide. I'll keep this in mind for next time. I opened a new clean PR with 1 commit.

PR #444, opened 2018-02-19
More phpdocs updates and OBJECT_NAME constant for ApiResources

nickdnk:
As requested I split my PR onto two, with this one having the phpdocs and OBJECT constant.

The idea with the OBJECT constant is to prevent cases where a dev might have an invalid value in the object-string, such as `application_fee_refund` instead of `fee_refund` or `invoice_item` instead of `invoiceitem`. Also it helps with typos in general.


ob-stripe:
Thanks for the contribution @nickdnk!

- I like the constants with the resource names! However, could you rename the constants to `OBJECT_NAME`? This would be consistent with our other libraries (e.g. [stripe-ruby](https://github.com/stripe/stripe-ruby/blob/master/lib/stripe/application_fee_refund.rb#L6)), is more explicit, and might help with some syntax highlighters (GitHub seems to think that `OBJECT` is a reserved keyword)

- I'm not thrilled about importing the exceptions as global names with `use` just to satisfy the PHPDoc declarations. I would rather not import the exceptions and use the absolute paths with PHPDoc (e.g. `@throws \Stripe\Error\Api`)

(Also, for future reference, these two changes don't seem too related and so should have been submitted in different PRs.)

nickdnk:
Hello @ob-stripe 

I changed it to not using `use` for errors and added all `@throws` tags. I started from the superclasses and inherited these throws down to all objects. I reviewed every object as thoroughly I could to make sure I didn't modify return types to something incorrect (phpstorm wants to do this some places because of traits).

And then I renamed to `OBJECT_NAME` as requested.

nickdnk:
Hello again

Do I need to change something to get this accepted? I have many places where I suppress "exception not thrown" warnings in PHPStorm because these docs are missing :(

lib/BankAccount.php
```diff
@@ -80,17 +83,25 @@ public static function update($_id, $_params = null, $_options = null)
...
     public function verify($params = null, $options = null)
     {
-        $url = $this->instanceUrl() . '/verify';
-        list($response, $opts) = $this->_request('post', $url, $params, $options);
-        $this->refreshFrom($response, $opts);
-        return $this;
+         $url = $this->instanceUrl() . '/verify';
+         list($response, $opts) = $this->_request('post', $url, $params, $options);
+         $this->refreshFrom($response, $opts);
+         return $this;
```
brandur-stripe: I think the indentation here was correct before. Can you dedent this block by one space?

lib/HttpClient/CurlClient.php
```diff
@@ -117,39 +124,39 @@ public function request($method, $absUrl, $headers, $params, $hasFile)
         if (is_callable($this->defaultOptions)) { // call defaultOptions callback, set options to return value
             $opts = call_user_func_array($this->defaultOptions, func_get_args());
             if (!is_array($opts)) {
-                throw new Error\Api("Non-array value returned by defaultOptions CurlClient callback");
+                throw new \Stripe\Error\Api("Non-array value returned by defaultOptions CurlClient callback");
```
brandur-stripe: I like that a lot of these symbols are now fully qualified! Much easier to follow now.

brandur-stripe:
@nickdnk Just for clarity: `OBJECT_NAME` isn't actually used anywhere I guess right? I guess my only concern here is that it'll be be pretty easy to omit it or get it wrong when a new model is added in the future.

@ob-stripe I'll leave this one up to you for final approval, but it's looking pretty good to me. The one piece that scares me a little bit is that there's so many places where we explicitly list every possible exception as a `@throws`. The detail of that approach is nice, but it'll be a bit painful anytime we want to add a new exception type.

ob-stripe:
Hi @nickdnk, sorry I dropped the ball here! Totally my bad.

> The one piece that scares me a little bit is that there's so many places where we explicitly list every possible exception as a @throws.

Yeah, I feel the same. We're backpedaling on that approach in stripe-java, so I'd rather avoid introducing this here. @nickdnk, what do you think of replacing the multiple `@throws` statements with just `@throws Stripe\Error\Base`?

nickdnk:
@ob-stripe I see your point. But adding Base defeats the purpose a little, as what I'm trying to avoid is this all over PHPStorm:
https://i.imgur.com/FxatxwN.png

But I suppose that adding `@throws Base` will at least let you know where you didn't handle all Stripe error cases, and will of course rid you of that warning when you do catch `Base`, which I assume is what most people do. I only rarely catch `InvalidRequest` or `Card` in addition to `Base`.

With regards to `OBJECT_NAME` - I do believe keeping that property updated on all models is a task that comes with the library, just as we did with event types on `Event` and as we did with `@property` tags. Looking at each API release note and checking if something has changed or is added should be all it takes - and again, if these do become outdated it won't break anything, but may just confuse people a bit (which may trigger a PR just as it did for me).

nickdnk:
I'm gonna close this and create a new one only for OBJECT_NAME changes.

PR #447, opened 2018-02-21
Automatic retry for rate-limited requests

lushc:
Hello everyone, very happy to see #428 land, it's a great feature! I noticed that [the docs](https://stripe.com/docs/testing#rate-limits) mention that it's safe to retry rate-limited requests with exponential backoff, so it'd be nice to also handle the 429 status code.

### DIFF
--- a/lib/HttpClient/CurlClient.php
+++ b/lib/HttpClient/CurlClient.php
@@ -304,8 +304,8 @@ class CurlClient implements ClientInterface
             return true;
         }
 
-        // 409 conflict
-        if ($rcode === 409) {
+        // 409 conflict or 429 too many requests
+        if ($rcode === 409 || $rcode === 429) {
             return true;
         }
 
--- a/tests/Stripe/HttpClient/CurlClientTest.php
+++ b/tests/Stripe/HttpClient/CurlClientTest.php
@@ -151,6 +151,15 @@ class CurlClientTest extends TestCase
         $this->assertTrue($this->shouldRetryMethod->invoke($curlClient, 0, 409, 0));
     }
 
+    public function testShouldRetryOnTooManyRequests()
+    {
+        Stripe::setMaxNetworkRetries(2);
+
+        $curlClient = new CurlClient();
+
+        $this->assertTrue($this->shouldRetryMethod->invoke($curlClient, 0, 429, 0));
+    }
+
     public function testShouldNotRetryAtMaximumCount()
     {
         Stripe::setMaxNetworkRetries(2);


ob-stripe:
Thanks for the contribution! This looks safe to me, but I'd prefer a second opinion.

r? @brandur-stripe

brandur-stripe:
@ob-stripe Yeah, I think it's okay. It'd be a problem if there was no quiesce period because it'd just keep on hammering our stack, but I'm +1 given the backoff that's built in here.

remi-stripe:
When using Idempotent Requests, we'd just retry the same request over and over and get the original rate limit error though. We likely need to special-case this.

lushc:
I'd like to revisit this if possible. Looking at the progress made in https://github.com/stripe/stripe-node/pull/559 it seems there are two other useful conditions for retrying the request:

* 503 status code
* 5xx status code for non-POST requests

I'll try to keep my eye on that PR and once it's approved/merged I'll align this one when I can (unless someone beats me to it!) Also, was there consensus on whether it was safe to retry rate-limited requests using idempotency keys?

ob-stripe:
@lushc Wonderful! I'm very much +1 to having the same behavior for automatic retries in all our client libraries.

Really sorry that we forgot about this PR btw 😞 We'll try not to let that happen again once you update it (but feel free to ping me if we do!).

ob-stripe:
> Also, was there consensus on whether it was safe to retry rate-limited requests using idempotency keys?

I think the consensus is that this is safe because automatic retries use an exponential backoff (otherwise the retries would risk compounding the rate limiting issue). @brandur-stripe, do you confirm?

brandur-stripe:
> I think the consensus is that this is safe because automatic retries use an exponential backoff (otherwise the retries would risk compounding the rate limiting issue). @brandur-stripe, do you confirm?

Yep, it's safe from the client-side in that we've got the exponential backoff and safe from the server side in that all rate limiting happens before an idempotency key is locked, so a 429ed or 503ed request will not have persisted an idempotency key at all.

One thing that I just thought of that I'd be a little concerned about — the trouble with automatically retrying rate limited requests is that it hides the obscures that you're at your rate limit. Ideally, a client should explicitly handle a rate limit error and use that as a signal to slow down their rate of requests. If we retried automatically that would all become opaque, so you could easily imagine the integration pouring in more requests on and having large numbers of them failed from a 429ed and sitting in retry queues, with the pressure continually building as more of them accumulate.

An alternative might be to suggest/implement an easy way to handle a rate limited error and retry the request.

brandur-stripe:
> 429ed or 503ed

Small correction: Remi reminded me that although what I said about 429s is true, it's only true about certain types of 503s. A 503 from a load shedder or top-level lock timeouts is safe to retry, but there are other types of 503s caused by more explicit locks which are not.

lushc:
No worries about forgetting the PR, these things happen 😄 Reading @brandur-stripe's reasoning both here and in https://github.com/stripe/stripe-node/pull/559#issuecomment-463374335 I agree that this isn't the correct way to go about retrying rate-limited requests. I thought it would be nice to hook into the existing exponential backoff logic, but didn't consider the side-effect that it'd mask an error you'd want clients to be aware of. I think aligning with the work done in stripe-node should be in a separate PR, so I'll leave it up to you whether to keep this one open or closed for the time being (since a few server-side improvements to make this possible may be pending) 👍

ob-stripe:
Okay, I'm going to close this one. We'll update the implementation to be consistent with stripe-node's and retry on 503's and non-POST 5xx's, but not on 429's.

Thanks @lushc!

PR #450, opened 2018-02-23
Add support for code attribute on all Stripe exceptions

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

PHP does not support overloading constructors, so I've had to resort to a slightly hackish solution to avoid modifying the constructor's signature and preserve BC. There was already a precedent for it though, for the `decline_code` attribute on card errors: https://github.com/stripe/stripe-php/blob/6791706360c98c9cb99b7ad5e1f1dc1c54f962cc/lib/Error/Card.php#L23-L28

### DIFF
--- a/lib/Error/Base.php
+++ b/lib/Error/Base.php
@@ -20,9 +20,18 @@ abstract class Base extends Exception
         $this->httpHeaders = $httpHeaders;
         $this->requestId = null;
 
+        // TODO: make this a proper constructor argument in the next major
+        //       release.
+        $this->stripeCode = isset($jsonBody["error"]["code"]) ? $jsonBody["error"]["code"] : null;
+
         if ($httpHeaders && isset($httpHeaders['Request-Id'])) {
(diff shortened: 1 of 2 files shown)


brandur-stripe:
LGTM.

I think for the next breaking release we should possibly instead consider just changing the constructor to take one associative array parameter ... real arguments are usually better, but when you have this many and so little typing, I'm not sure if there's any real advantage over simulating named parameters with an array.

It's too bad — Googling around a bit it seems like there's a fair bit of demand for named parameters in the language, but the only format RFC for it seems to have stalled out: https://wiki.php.net/rfc/named_params

ob-stripe:
Yeah, good point. We could also simply pass the HTTP status code and the JSON body, and deduce all the other attributes from that. It's probably a little unconventional to do anything but store some values in an exception constructor, but since our exceptions are basically models for our errors, it kind of makes sense that they'd know how the JSON is structured.

PR #452, opened 2018-02-28
Invoice missing amount_paid and amount_remaining properties

nickdnk:
For some reason I missed these in my phpdoc frenzy. Probably because they're also not listed on https://stripe.com/docs/api#invoice_object in the left section, only in the example on the right.

Just a minor edit.

### DIFF
--- a/lib/Invoice.php
+++ b/lib/Invoice.php
@@ -8,6 +8,8 @@ namespace Stripe;
  * @property string $id
  * @property string $object
  * @property int $amount_due
+ * @property int $amount_paid
+ * @property int $amount_remaining
  * @property int $application_fee
  * @property int $attempt_count
  * @property bool $attempted


brandur-stripe:
Looks good. Thank you!

brandur-stripe:
Released as 6.3.2.

nickdnk:
Cool

However, it seems these properties don't exist on all invoices, which I don't really understand. I emailed support about it. I'll update this thread when they reply.

ob-stripe:
Just a guess, but I think these properties are specific to invoices paid via [ACH Credit Transfer](https://stripe.com/docs/subscriptions/ach-credit).

nickdnk:
I received this answer from support:

> Thanks for writing in, and for contributing to our libraries! You're totally right: `amount_paid` and `amount_remaining` were added last week and haven't yet been documented.  
> 
> There is a bug with these fields that our Subscriptions team is actively working on fixing -- I'll let you know when a fix is out.

So I suppose the properties are valid for all invoices but just bug out a little bit currently.

ob-stripe:
Yep, seems likely. Thanks for closing the loop!

PR #455, opened 2018-03-01
Fixed namespaces in docblocks

weirdan:
Fixed a number of unqualified/wrongly qualified class references in
docblocks identified by static analysis.

Also fixed one param docblock that used `number` as a type (which is not
according to phpdoc docs).


ob-stripe:
Thanks @weirdan!

PR #456, opened 2018-03-01
Exception namespace fixes

weirdan:
Fixed a number of `throw`s that used non-imported exception classes.


ob-stripe:
Released in 6.4.1. Thanks for the contribution @weirdan!

PR #458, opened 2018-03-17
Added $account property for connect webhook events

nickdnk:
Hello

According to https://stripe.com/docs/connect/webhooks, the Event class has a property called `account` when the event comes from a connected account. I've just added this to phpdocs.

### DIFF
--- a/lib/Event.php
+++ b/lib/Event.php
@@ -7,6 +7,7 @@ namespace Stripe;
  *
  * @property string $id
  * @property string $object
+ * @property string $account
  * @property string $api_version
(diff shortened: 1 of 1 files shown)


ob-stripe:
Hi @nickdnk, do you mind just moving the attribute after `$object` and before `$api_version`? Aside from `$id` and `$object` which are always at the top, the attributes should be kept in alphabetical order. Thanks!

nickdnk:
Done.

ob-stripe:
Released in 6.4.2.

PR #461, opened 2018-04-05
Fix replacement of non-`metadata` embedded `StripeObjects`

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

This should look familiar :)

https://github.com/stripe/stripe-ruby/pull/632 for context.


brandur-stripe:
It makes me so happy that we're finally being a little proactive in getting these bugs fixed before they're reported :)

LGTM.

Thanks for taking the initiative here @ob-stripe!

PR #468, opened 2018-05-12
Updated phpdocs

nickdnk:
Hey guys

I made some updates to phpdocs for missing/new properties :)

Cheers

### DIFF
--- a/lib/Recipient.php
+++ b/lib/Recipient.php
@@ -6,6 +6,21 @@ namespace Stripe;
  * Class Recipient
  *
  * @package Stripe
+ *
+ * @property string $id
+ * @property string $object
+ * @property mixed $active_account
+ * @property Collection $cards
+ * @property int $created
+ * @property string $default_card
+ * @property string $description
+ * @property string $email
+ * @property bool $livemode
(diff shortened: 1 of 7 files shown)


lib/Recipient.php
```diff
@@ -6,6 +6,22 @@
...
+ * @property string $default_card
+ * @property string $description
+ * @property string $email
+ * @property bool $livemode
+ * @property StripeObject $metadata
+ * @property string $migrated_to
+ * @property string $name
+ * @property string $rolled_back_from
+ * @property string $type
+ * @property bool $verified
```
brandur-stripe: I wonder if we should pull `verified`? I think it's technically supposed to be undocumented.

brandur-stripe:
Excellent work as usual @nickdnk! Thanks so much for the corrections.

I left one minor comment above, but this looks good to me.

nickdnk:
@brahn-stripe I've removed the verified property as requested and also added $support_address to Account.

brandur-stripe:
@nickdnk Excellent. Thank you!

brandur-stripe:
Released as 6.7.1.

PR #471, opened 2018-05-13
Added OBJECT_NAME constant to all API objects and added missing object

nickdnk:
Hello guys

As in a previous PR that was a bit more of a mess I added the OBJECT_NAME as we discussed once before. This time the PR is cleaner with only this as focus. To refresh: This is to avoid situations where one may look for `invoice_item` instead of `invoiceitem` or `application_fee_refund` instead of `fee_refund`.

I also added the InvoiceLineItem object, which according to the docs is almost the same as Invoice Item. I am not sure why they both exist, but their object property is different ("invoiceitem" vs "line_item") and one has a few properties the other does not, so there must be a case for both of them. If you don't like this let me know and I'll remove it from the PR.


brandur-stripe:
Hey @nickdnk, thanks the patch! This is quite similar to the setup that we already have in stripe-ruby, so I think this is a reasonable approach.

Would you mind seeing if you could update [the map in `Util.php`](https://github.com/stripe/stripe-php/blob/master/lib/Util/Util.php#L73) to take advance of these new constants? I believe that a few of our classes are missing them, so this might also help find the places where `OBJECT_NAME` is forgotten.

> I also added the InvoiceLineItem object, which according to the docs is almost the same as Invoice Item. I am not sure why they both exist, but their object property is different ("invoiceitem" vs "line_item") and one has a few properties the other does not, so there must be a case for both of them. If you don't like this let me know and I'll remove it from the PR.

Thanks! I think it's fine to include invoice line item in here.

nickdnk:
I don't understand why the build is failing with autoload 0. It says it cannot find the InvoiceLineItem class, but I've used it the exact same way as all the other classes in the Util map.

nickdnk:
@brandur-stripe That did the trick.

lib/Util/Util.php
```diff
@@ -2,7 +2,52 @@
...
+use Stripe\SourceTransaction;
 use Stripe\StripeObject;
+use Stripe\Subscription;
+use Stripe\SubscriptionItem;
+use Stripe\ThreeDSecure;
+use Stripe\Token;
+use Stripe\Topup;
+use Stripe\Transfer;
+use Stripe\TransferReversal;
+use Stripe\UsageRecord;
```
brandur-stripe: @nickdnk Pretty minor, but given the size of this list, what do you think about just using the fully qualified path directly on the lines in the map below so we can avoid two of these big lists?

i.e., Something like this:

``` php
\Stripe\Account::OBJECT_NAME => 'Stripe\\Account',
```

brandur-stripe:
Left one more comment, but awesome work here. Thanks.

nickdnk:
Hey guys
It seems we stalled here. Want me to change anything, @brandur-stripe?

lib/Util/Util.php
ob-stripe: Hi @nickdnk! Thanks a lot for the contribution. Would you mind addressing @brandur-stripe's comment above? Once that's done, I think we'll be ready to merge :)

ob-stripe:
@nickdnk Brandur is not available today (US bank holiday) but I took a look and I think this looks good, modulo Brandur's comment re. using fully qualified paths so that we have one less list to maintain when a new resource is added.

nickdnk:
@ob-stripe Done :)

ob-stripe:
Released in 6.7.2. Thanks again @nickdnk :)

nickdnk:
@ob-stripe Cool. Did you remember to check the use traits on InvoiceLineItem? I was not sure which were applicable to that object.

ob-stripe:
Aw, I missed that. So `InvoiceLineItem` should not have any traits -- the only operation it supports is listing, and that's done via the `lines` list object automatically returned when fetching an invoice. I'll push a fix.

nickdnk:
Okay cool. Just wanted to make sure you were on top of it. I don't want to introduce issues :)

PR #472, opened 2018-05-14
Added phpDoc for ApiRequestor, especially regarding thrown errors

rtm-ctrlz:


lib/ApiRequestor.php
```diff
@@ -9,12 +9,27 @@
...
+     */
     private $_apiKey;
 
+    /**
+     * @var string
+     */
     private $_apiBase;
 
+    /**
+     * @var HttpClient\CurlClient
```
brandur-stripe: I think this is more like a `ClientInterface` rather than necessarily a `CurlClient`.

lib/ApiRequestor.php
```diff
@@ -158,6 +223,13 @@ private static function _specificOAuthError($rbody, $rcode, $rheaders, $resp, $e
         return null;
     }
 
+    /**
+     * @static
+     *
+     * @param $appInfo
```
brandur-stripe: Should we specify a type for this parameter?

rtm-ctrlz:
ptal @brandur-stripe

lib/ApiRequestor.php
```diff
@@ -312,11 +430,19 @@ private function _interpretResponse($rbody, $rcode, $rheaders)
         return $resp;
     }
 
+    /**
+     * @static
+     *
+     * @param HttpClient\CurlClient $client
```
brandur-stripe: I missed this before, but could you also set this one as `HttpClient\ClientInterface`?

lib/ApiRequestor.php
```diff
@@ -312,11 +430,19 @@ private function _interpretResponse($rbody, $rcode, $rheaders)
...
+     */
     public static function setHttpClient($client)
     {
         self::$_httpClient = $client;
     }
 
+    /**
+     * @return HttpClient\CurlClient
```
brandur-stripe: And same here.

brandur-stripe:
Thanks for circling back on this @rtm-ctrlz. I'd missed a couple other instances where `HttpClient\ClientInterface` should be used and added comments on the lines above. Otherwise, LGTM.

rtm-ctrlz:
Fixed CurlClient -> ClientInterface
Also I had to fix phpdoc for ClientInterface interface since @throws and @return sections was broken

ptal @brandur-stripe

brandur-stripe:
> Also I had to fix phpdoc for ClientInterface interface since @throws and @return sections was broken

Oh nice.

LGTM. Thanks!

brandur-stripe:
Released as 6.8.1.

PR #475, opened 2018-05-17
Allow stripe-mock version identifying as `master`

brandur:
As in other API libraries testing against stripe-mock, allow `master` as
a catch all version so that we can test the library against the `master`
branch without getting caught by the minimum version check.


brandur-stripe:
(Again, very minor so going to self-approve.)

PR #480, opened 2018-05-29
API 2018-05-21 changes and new charge.expired event

nickdnk:
Hello guys

Just a few minor edits here.

According to the 2018-05-21 update, the `skus` collection has been removed entirely from the `Product` object. I've fixed that in phpdocs.

I've also added the `unit_label` property which was missing from `Product` as well.

I've also added the `charge.expired` event that's now possible to receive from webhooks (https://stripe.com/blog/changelog - https://stripe.com/docs/api#event_types-charge.expired).

Cheers


brandur-stripe:
Nice work here as usual @nickdnk. Thanks for the continuing updates!

brandur-stripe:
Released as 6.7.4.

PR #481, opened 2018-05-30
Added more constants, fixed more docs, added missing objects

nickdnk:
Hello

1. Added Discount, DisputeEvidence and OrderItem objects - none of these extend ApiResource since they don't have an `id` property which I assume is necessary for that. Anywhere these objects are referenced in phpdocs I've update those from the previously used `mixed` type.

2. Added charge decline code constants to Charge - following the logic elsewhere in the library. I was not sure if this should be on Card or on Charge, but I think Charge, no?

3. Added payout failure code constants to Payout - following the logic elsewhere in the library.

4. Indented subscription constants like elsewhere - minor edit.

Some things I noticed:

The Topup object is missing from the documentation, even though it links to it itself: https://stripe.com/docs/api#topup_object - it doesn't exist on the page. For this reason I can't check if the phpdocs match.

The SourceTransaction object - what is that? Why do we have it? I can't find it anywhere in the documentation. Isn't this just a normal Charge object used as a source for transactions? See https://stripe.com/docs/connect/charges-transfers#transfer-availability

And; is there somehow we can follow changes made to the API? It doesn't seem to be documented anywhere when changes to properties take place, which makes it quite tedious to go through all of the objects again and again to keep them updated. I assume this is the reason the phpdocs were outdated in the first place.

Cheers


nickdnk:
@ob-stripe 

Hey

I removed the `DisputeEvidence` class and made the other changes you mentioned. I'll leave it up to you guys to decide if you want `OrderItem` and `Discount` as classes in the library, since they don't "do" anything. I think it's fair to have them because of the `OBJECT_NAME` constant - for consistency and to support phpdocs. They now extend `StripeObject`.

I added those classes to `Util` and `init.php`.

I'll rebase to one commit once we've agreed on everything.

nickdnk:
Does this mean that all variables currently defined as `mixed` could/should be `StripeObject` ?

nickdnk:
I've just added the `invoice_pdf` and `hosted_invoice_url` properties to `Invoice` - these are new :)

nickdnk:
Added the `active` property to `Plan`

brandur-stripe:
Thanks for keeping this one alive and updated @nickdnk!

If you get a chance, could you squash your commits? Otherwise we'll do it for you when we bring it in (and the commit message might not be as good).

> Overall though, I have some concerns about adding those new classes purely for the sake of describing the attributes with phpdoc. @brandur-stripe wdyt?

Sorry @ob-stripe, I totally missed this @ to me last time. I guess I could see it either way, but IMO since these are considered first class API objects (i.e. they have their own sections in the API reference), it's okay to have them as separate classes, so I'd +1 this as currently implemented. I'll defer to you though.

I did a quick pass on the changes otherwise too, and LGTM.

ptal @ob-stripe

nickdnk:
@brandur-stripe Squashed as requested :)

ob-stripe:
Released as 6.8.0. Thanks again @nickdnk :)

PR #487, opened 2018-06-27
Add support for the PaymentIntent resource

remi-stripe:
This feature is currently gated so the tests are stubbed for now.

r? @ob-stripe 
cc @stripe/api-libraries 
cc @jenan-stripe @michelle-stripe

### DIFF
--- a/lib/PaymentIntent.php
+++ b/lib/PaymentIntent.php
@@ -1,0 +1,101 @@
+<?php
+
+namespace Stripe;
+
+/**
+ * Class PaymentIntent
+ *
+ * @property string $id
+ * @property string $object
+ * @property string[] $allowed_source_types
+ * @property int $amount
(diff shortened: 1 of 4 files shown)


lib/PaymentIntent.php
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * This is a special case because the payment intents endpoint has an
+     *    underscore in it. The parent `className` function strips underscores.
+     *
+     * @return string The name of the class.
+     */
+    public static function className()
+    {
+        return 'payment_intent';
```
remi-stripe: Minor but I spent 20 minutes on figuring out that I was missing this again. I wonder if we can/should use `OBJECT_NAME` instead.

ob-stripe:
Released in 6.9.0.

PR #488, opened 2018-06-28
Adds support for 'partner_id' to 'setAppInfo'

zachwick:
This adds support for an optional partner_id attribute in the setAppInfo hash.

### DIFF
--- a/lib/Stripe.php
+++ b/lib/Stripe.php
@@ -199,12 +199,13 @@ class Stripe
      * @param string $appVersion The application's version
      * @param string $appUrl The application's URL
      */
-    public static function setAppInfo($appName, $appVersion = null, $appUrl = null)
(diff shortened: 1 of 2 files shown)


lib/Stripe.php
```diff
@@ -199,12 +199,13 @@ public static function getAppInfo()
      * @param string $appVersion The application's version
      * @param string $appUrl The application's URL
      */
-    public static function setAppInfo($appName, $appVersion = null, $appUrl = null)
+    public static function setAppInfo($appName, $appPartnerId = null, $appUrl = null, $appVersion = null)
```
ob-stripe: Unfortunately PHP doesn't support keyword arguments, so this is a breaking change. Mind setting `$appPartnerId` as the last argument?

zachwick:
Reordered the positional arguments.

ptal @ob-stripe

PR #497, opened 2018-07-14
Use HTTP/2 only for HTTPS requests

franzliedke:
The previously used flag tried to upgrade *any* request to HTTP/2, by sending an `Upgrade: h2c` header. Some HTTP servers (such as PHP's built-in one, which I use for mocking HTTP requests) apparently cannot deal with it. Using the `CURL_HTTP_VERSION_2TLS` flag instead makes Curl attempt an upgrade for HTTPS requests (e.g. to Stripe's API endpoint), but not for HTTP ones.

See https://ec.haxx.se/http-http2.html and https://curl.haxx.se/libcurl/c/CURLOPT_HTTP_VERSION.html.

(Plus, see my previous comments in PR #494 which introduced this feature.)

### DIFF
--- a/lib/HttpClient/CurlClient.php
+++ b/lib/HttpClient/CurlClient.php
@@ -22,8 +22,8 @@ if (!defined('CURL_SSLVERSION_TLSv1_2')) {
 }
 // @codingStandardsIgnoreEnd
 
-if (!defined('CURL_HTTP_VERSION_2_0')) {
-    define('CURL_HTTP_VERSION_2_0', 3);
+if (!defined('CURL_HTTP_VERSION_2TLS')) {
+    define('CURL_HTTP_VERSION_2TLS', 4);
 }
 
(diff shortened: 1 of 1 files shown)


franzliedke:
Obligatory XKCD reference:

XKCD: This update broke my workflow

*(from https://xkcd.com/1172/)*

ob-stripe:
Aw, that's really unfortunate that PHP's builtin server does not handle this well :/

Thanks for looking into this and submitting a PR!

ob-stripe:
Released as 6.10.3.

franzliedke:
Thanks for the swift response. Works like a charm. :+1:

franzliedke:
Another option, by the way, would be to make the Curl flag overridable. Right now, any of the [user-providable default options](https://github.com/stripe/stripe-php#custom-curl-options-eg-proxies) will just be ignored - at least for this flag.

PR #498, opened 2018-07-18
Better `classUrl()`

ob-stripe:
cc @stripe/api-libraries 

Simplifies `classUrl()` by using the recently-added `OBJECT_NAME` class constants. Removes `className()` entirely.

### DIFF
--- a/lib/ApiResource.php
+++ b/lib/ApiResource.php
@@ -67,30 +67,6 @@ abstract class ApiResource extends StripeObject
     }
 
     /**
-     * @return string The name of the class, with namespacing and underscores
-     *    stripped.
-     */
-    public static function className()
-    {
-        $class = get_called_class();
-        // Useful for namespaces: Foo\Charge
-        if ($postfixNamespaces = strrchr($class, '\\')) {
-            $class = substr($postfixNamespaces, 1);
-        }
-        // Useful for underscored 'namespaces': Foo_Charge
-        if ($postfixFakeNamespaces = strrchr($class, '')) {
-            $class = $postfixFakeNamespaces;
-        }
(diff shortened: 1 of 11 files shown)


lib/ApiResource.php
```diff
@@ -66,30 +66,6 @@ public function refresh()
         return $this;
     }
 
-    /**
-     * @return string The name of the class, with namespacing and underscores
-     *    stripped.
-     */
-    public static function className()
```
remi-stripe: I'm so glad to see this one go

lib/ApiResource.php
```diff
@@ -103,7 +79,7 @@ public static function baseUrl()
      */
     public static function classUrl()
     {
-        $base = static::className();
+        $base = str_replace('.', '/', static::OBJECT_NAME);
```
remi-stripe: we likely should add a comment about that one to be clearer/more discoverable in the future

lib/SingletonApiResource.php
```diff
@@ -22,7 +22,7 @@ protected static function _singletonRetrieve($options = null)
      */
     public static function classUrl()
     {
-        $base = static::className();
+        $base = static::OBJECT_NAME;
```
remi-stripe: OOC what is that file about? I don't remember seeing it before

brandur-stripe:
LGTM. Very nice cleanup!

lib/SingletonApiResource.php
ob-stripe: The only singleton resource is `Balance`. I guess that we expected this to be a more common pattern in the API than it ended up being.

ob-stripe:
Released as 6.10.4.

PR #499, opened 2018-07-22
Add Charge::sanitizeStatementDescriptor(...) method

imliam:
Charge statement descriptors have some [hard rules](https://stripe.com/docs/connect/statement-descriptors) about what they are allowed to be. This PR offers a method to take any string and sanitize it to be an acceptable statement descriptor.

### DIFF
--- a/tests/Stripe/ChargeTest.php
+++ b/tests/Stripe/ChargeTest.php
@@ -137,4 +137,41 @@ class ChargeTest extends TestCase
         $this->assertInstanceOf("Stripe\\Charge", $resource);
         $this->assertSame($resource, $charge);
     }
+
+    public function testStatementDescriptorSanitization()
+    {
+        // Strings that are fine will pass.
+        $this->assertEquals(
+            'This is fine.',
+            Charge::sanitizeStatementDescriptor('This is fine.')
+        );
+
+        // Long strings are truncated.
+        $this->assertEquals(
+            'This string is longer ',
+            Charge::sanitizeStatementDescriptor('This string is longer than 22 characters.')
+        );
+
+        // Disallowed characters are stripped out.
+        $this->assertEquals(
+            '<b>"Disallowed\' bits</b>',
+            Charge::sanitizeStatementDescriptor('bDisallowed bits/b')
+        );
+
+        // A prefix is not applied if the descriptor is not numeric.
+        $this->assertEquals(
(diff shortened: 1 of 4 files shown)


tests/Stripe/ChargeTest.php
```diff
@@ -137,4 +137,41 @@ public function testCanMarkAsSafe()
...
+        // Long strings are truncated.
+        $this->assertEquals(
+            'This string is longer ',
+            Charge::sanitizeStatementDescriptor('This string is longer than 22 characters.')
+        );
+
+        // Disallowed characters are stripped out.
+        $this->assertEquals(
+            '<b>"Disallowed\' bits</b>',
+            Charge::sanitizeStatementDescriptor('bDisallowed bits/b')
```
ob-stripe: I think you swapped the expected string and the input string here :)

tests/Stripe/ChargeTest.php
```diff
@@ -137,4 +137,41 @@ public function testCanMarkAsSafe()
...
+        );
+
+        // A prefix is applied if the descriptor is numeric.
+        $this->assertEquals(
+            'Example: 123',
+            Charge::sanitizeStatementDescriptor('123', 'Example: ')
+        );
+
+        // A prefix must be supplied if the descriptor is numeric.
+        $this->expectsException(\InvalidArgumentException::class);
```
ob-stripe: Unfortunately, the library still needs to support PHP 5.4 so we cannot use `::class`.

Also, I don't think `expectsException` is a valid PHPUnit method -- afaik you have to use the [`@expectedException`](https://phpunit.de/manual/4.8/en/appendixes.annotations.html#appendixes.annotations.expectedException) annotation, which would require splitting this test into multiple methods.

ImLiam:
Sorry about that, maybe I rushed this PR without too much attention to it on the spot, hopefully, these changes should sort it.

- I think `Util` is a fine place for it since it can be used in multiple places, there's not really any other place currently that's better suited for it.

- I've removed the `$prefix` parameter

- I've re-ordered the sanitization rules so that the special characters are removed and the string is truncated first, so now the exceptions are thrown based on the already sanitized string.

- Perhaps there is room for a similar method `isValidStatementDescriptor` that simply returns a boolean without sanitizing it?

ob-stripe:
Closing due to staleness.

PR #502, opened 2018-07-31
Add `isDeleted()` method to `StripeObject`

ob-stripe:
cc @stripe/api-libraries 

Adds a `isDeleted()` method to `StripeObject` that returns `true` if the resource has been deleted and `false` otherwise.

This is similar to what we have in stripe-ruby: https://github.com/stripe/stripe-ruby/blob/ec66c3f0f44274f885de8d13de5dce2657932121/lib/stripe/stripe_object.rb#L99-L104


ob-stripe:
Released as 6.13.0.

Issue #507, opened 2018-08-10
satooshi/php-coveralls listed as abandoned


chris-rempel-wag:
Currently, `satooshi/php-coveralls` is listed as a `require-dev` in `composer.json`. However, this package is listed as abandoned and has been replaced with `php-coveralls/php-coveralls`. Is this something that should be fixed? Why or why not?

PR #508, opened 2018-08-13
Update reference to php-coveralls

remi-stripe:
Fixes https://github.com/stripe/stripe-php/issues/507

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
   },
   "require-dev": {
     "phpunit/phpunit": "~4.0",
-    "satooshi/php-coveralls": "~0.6.1",
+    "php-coveralls/php-coveralls": "1.*",
     "squizlabs/php_codesniffer": "~2.0"
   },
   "autoload": {
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,4 +1,3 @@
 service_name: travis-ci
-src_dir: .
 coverage_clover: clover.xml
 json_path: coveralls-upload.json


composer.json
```diff
@@ -22,7 +22,7 @@
   },
   "require-dev": {
     "phpunit/phpunit": "~4.0",
-    "satooshi/php-coveralls": "~0.6.1",
+    "php-coveralls/php-coveralls": "1.*",
```
remi-stripe: using `1.*` as we are still supporting PHP 5.4

brandur-stripe:
@remi-stripe Thanks!

If you pull up one of the builds that looked like they succeeded ([example](https://travis-ci.org/stripe/stripe-php/jobs/415408238)), you can can see that they actually failed during the 
Coveralls step:

image

(You need to expand the step to see it though.)

Travis is hiding the failure because its running in the `after_script` command (see [docs on that here](https://blog.travis-ci.com/after_script_behavior_changes/)).

I think to get this working you'll need to tweak `.coveralls.yml` a bit:

``` patch
diff --git a/.coveralls.yml b/.coveralls.yml
index a343c9a..4eecff5 100644
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,4 +1,3 @@
 service_name: travis-ci
-src_dir: .
 coverage_clover: clover.xml
 json_path: coveralls-upload.json
```

It also seems a little unreasonably easy to accidentally break Coveralls due to the `after_script` thing — we should consider moving that into the `script` block instead.

ptal @remi-stripe

remi-stripe:
wow nice catch @brandur-stripe I would never have thought to look at the output. I just fixed it with your advice

```
Accepted Job #2134.1
You can see the build on https://coveralls.io/jobs/39267727
elapsed time: 1.033 sec memory: 7.75 MB
Done. Your build exited with 0.
```

brandur-stripe:
Nice! LGTM.

PR #516, opened 2018-09-13
Register autoload instead of requiring all classes

julienbourdeau:
Hi 👋 ,

I work at Algolia and I'm currently writing our PHP API Client. I was looking at yours because I believe we have similar requirements: maintaining old verison of PHP (I even maintain 5.3), powerful enough for Enterprise customer but as easy as possible for self-serve users and such.

I was looking at your `init.php`, I maintain something similar for WordPress or Drupal users. For the v2 I went for an autoloader instead of requiring all the classes. As far as I tested, it works great.

I created this PR to open the conversation, what do you think of this solution? Do you know a reason why it wouldn't work that I'm missing?

### DIFF
--- a/init.php
+++ b/init.php
@@ -1,122 +1,32 @@
 <?php
 
-// Stripe singleton
-require(dirname(__FILE__) . '/lib/Stripe.php');
+/**
+ * Based on https://www.php-fig.org/psr/psr-4/examples/.
+ */
+spl_autoload_register(function ($class) {
+    // project-specific namespace prefix
+    $prefix = 'Stripe\\';
 
-// Utilities
-require(dirname(__FILE__) . '/lib/Util/AutoPagingIterator.php');
-require(dirname(__FILE__) . '/lib/Util/LoggerInterface.php');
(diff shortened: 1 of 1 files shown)


ob-stripe:
Hi @julienbourdeau! ❤️ Algolia :)

This looks super promising. I don't know much about PHP autoload mechanics, but if this works as advertised and saves us from manually having to add every new file, I'd be very much in favor of pulling this in.

cc @brandur-stripe @remi-stripe for visibility

brandur-stripe:
Cool! Personally, I kind of like the explicitness of big lists of files, but I can see how this would help make things a little more obvious for new contributors at least, so @ob-stripe feel free to do what you think is best.

julienbourdeau:
I think I'm gonna go with it for our v2. I like not having to maintain the list.

Feel free to close this PR if you don't want to merge it, I won't be offended 😄

ob-stripe:
On second thought, while I think maintaining a long list of `require`s is a bit tedious, it has the advantage of being very explicit. The autoloader code is more obscure. Better the devil you know...

Going to close without merging, but thanks again for the contribution :)

PR #517, opened 2018-09-14
Integer-index encode all sequential arrays

ob-stripe:
cc @stripe/api-libraries 

Okay, this goes a bit beyond just changing the encoding of arrays to use integer-indexing. I updated all the encoding logic to be ~identical to stripe-ruby's.


brandur-stripe:
This is soooo much better than what we had before! Thanks for being the one to get their hands dirty.

LGTM.

ob-stripe:
Haha, thanks for writing good code that's easy to port :)

ob-stripe:
Released as 6.17.1.

PR #522, opened 2018-09-17
Handle case when adding a new additional owner to the array

### LINKED ISSUE #521: E_NOTICE level error when updating additional owners.
I was following the instructions at https://stripe.com/docs/connect/identity-verification-api#additional-owners to update a connected account with a second additional owner. I noticed that the code actually raises a `Undefined offset: 1` 'error'. With default PHP settings, this is just a notice, and it isn't even logged. 

The error seems to come from `serializeAdditionalOwners`:

https://github.com/stripe/stripe-php/blob/b07887090c13674912a9a1a8fed8e42f174eeab8/lib/Account.php#L229-L237

`$additionalOwners` is the new value being passed in(which may be larger than the existing `$originalValue`), so we basically rely on being able to set an array property at an index that doesn't currently exist(`$originalValue[$i]`). This works fine generally since it only throws an E_NOTICE level error.

However, if you have set up your PHP environment to convert errors into fatal exceptions(http://php.net/manual/en/class.errorexception.php#errorexception.examples), this can be a problem. I'm not sure how to proceed here since I'm not very familiar with PHP best practises — is it a problem that we generate an E_NOTICE level error and something that we should fix?

I have a repro script here : https://gist.github.com/karlr-stripe/a9447ef36d688afd84a5c1f3f9b294cf 

You can workaround this by updating the account object this way instead : 

```
$count = count($stripeAccount->legal_entity->additional_owners);
$newAdditionalOwner = [
    $count => [
      'first_name' => 'Jane',
      'last_name' => 'Smith',
    ]
  ];

  \Stripe\Account::update($stripeAccount->id, array(
    "legal_entity" => array(
      "additional_owners" => $newAdditionalOwner
    )
  ));
```

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @karlr-stripe 

Fixes #521.

I've verified that without the fix, the new test does trigger the issue reported by @karlr-stripe in #521:
```
There was 1 error:

1) Stripe\AccountTest::testSerializeAddAdditionalOwners
Undefined offset: 2

/Users/ob/workspace/php/stripe-php/lib/Account.php:235
/Users/ob/workspace/php/stripe-php/lib/Account.php:208
/Users/ob/workspace/php/stripe-php/tests/Stripe/AccountTest.php:225
```

### DIFF
--- a/lib/Account.php
+++ b/lib/Account.php
@@ -230,7 +230,9 @@ class Account extends ApiResource
             $update = ($v instanceof StripeObject) ? $v->serializeParameters() : $v;
 
             if ($update !== []) {
-                if (!$originalValue || ($update != $legalEntity->serializeParamsValue($originalValue[$i], null, false, true))) {
+                if (!$originalValue ||
+                    !array_key_exists($i, $originalValue) ||
+                    ($update != $legalEntity->serializeParamsValue($originalValue[$i], null, false, true))) {
                     $updateArr[$i] = $update;
                 }
             }
--- a/tests/Stripe/AccountTest.php
+++ b/tests/Stripe/AccountTest.php
@@ -202,6 +202,29 @@ class AccountTest extends TestCase
         $this->assertSame($expected, $obj->serializeParameters());
     }
 
+    public function testSerializeAddAdditionalOwners()
+    {
+        $obj = Util\Util::convertToStripeObject([
+            'object' => 'account',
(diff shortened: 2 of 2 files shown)


brandur-stripe:
LGTM.

ob-stripe:
Released as 6.17.2.

PR #526, opened 2018-09-25
Ignore null values in params

### LINKED ISSUE #525: Null array value regression
Hi, we updated the Stripe PHP library on our server the other day to 6.17.2 and our order process has started failing. After looking into it, it appears the problem may be with the Stripe PHP library.

Our code:
`$customer_id = \Stripe\Customer::create(array('email' => $email_address, 'metadata' => $wp_user_id == 0 ? array() : array('wp_user_id' => $wp_user_id), 'source' => $source))->id;`

The catch here is that `source` may be null. This was done to make passing a source optional. In previous Stripe PHP versions, setting an array value to null meant it would not be sent at all. It appears the behavior has recently changed to instead interpret null as an empty string, so our order process has blown up with this Stripe error response: Invalid source object: must be a dictionary or a non-empty string.

I looked in our Stripe dashboard and see there is a source in the request body as an empty string, which should not be there.

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries @EatonZ

This fixes a regression I caused in #517 that caused `null` values in parameters to be encoded as empty strings and sent in requests.

stripe-ruby handles this via the [`objects_to_ids`](https://github.com/stripe/stripe-ruby/blob/7b939cb7d80f2f5f9082475c2cee8cc6c37918e3/lib/stripe/util.rb#L27-L40) helper method, whose main purpose is to replace instances of Stripe resources with their IDs, but that also clears null values. Since the parameter encoding logic in stripe-php is already very similar to stripe-ruby's, I simply ported the method.

Fixes #525.


lib/Util/Util.php
```diff
@@ -206,6 +206,38 @@ public static function secureCompare($a, $b)
...
+     * ApiResource, then it is replaced by the resource's ID.
+     * Also clears out null values.
+     *
+     * @param mixed $h
+     * @return mixed
+     */
+    public static function objectsToIds($h)
+    {
+        if ($h instanceof \Stripe\ApiResource) {
+            return $h->id;
```
remi-stripe: I wonder if there will be some cases where some resources will not have an id and break this logic and we will regret this.

tests/Stripe/Util/UtilTest.php
```diff
@@ -46,6 +46,26 @@ public function testUtf8()
...
     }
 
+    public function testObjectsToIds()
+    {
+        $params = [
+            'foo' => 'bar',
+            'customer' => Util\Util::convertToStripeObject([
+                'id' => 'cus_123',
+                'object' => 'customer',
+            ], null),
```
remi-stripe: indentation on this kind of threw me off

lib/Util/Util.php
ob-stripe: Yeah, I considered that. This would only happen if a user passed a resource without an ID inside a parameter hash though. We could check if the resource has an `id` property and output an error message if not, but I don't think it's worth the hassle.

(stripe-ruby has the same issue fwiw)

PR #527, opened 2018-09-28
Updated phpdocs

nickdnk:
Hey guys. Long time no see.

I did a once-over of the php docs and found the following changes.

### DIFF
--- a/lib/Charge.php
+++ b/lib/Charge.php
@@ -29,6 +29,7 @@ namespace Stripe;
  * @property string $order
  * @property mixed $outcome
  * @property bool $paid
+ * @property string $payment_intent
  * @property string $receipt_email
  * @property string $receipt_number
  * @property bool $refunded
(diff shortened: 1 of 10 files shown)


lib/BankAccount.php
```diff
@@ -14,7 +14,6 @@
  * @property string $country
  * @property string $currency
  * @property string $customer
- * @property bool $default_for_currency
```
remi-stripe: This property should stay as it still exists: https://stripe.com/docs/api#account_bank_account_object-default_for_currency

nickdnk: I thought this object was this: https://stripe.com/docs/api#customer_bank_account_object

What's the difference between these two?

remi-stripe: The resource represents a Bank Account in the API. This is shared by Customers and Accounts (for External Accounts) since the resource itself has the same shape (except a small number of properties). You can confirm this in the `instance_url` method [here](https://github.com/stripe/stripe-php/blob/master/lib/BankAccount.php#L38-L55) which shows how we calculate the endpoint to use when accessing/updating a bank account

nickdnk:
Removed the default for currency change

remi-stripe:
Thanks again for making those changes @nickdnk

PR #530, opened 2018-10-08
Supply constants for Source objects

lode:
I'd like to compare status and usage of a source I fetch from the api. I added flow as it might be handy for others. I didn't add type as it felt to fluid and less so something you depend flow upon, more something you might store as is.


ob-stripe:
Released as 6.19.3.

lode:
Thanks!

PR #531, opened 2018-10-08
Store HTTP response headers in case-insensitive array

### LINKED ISSUE #529: Request-Id is now lowercase
After upgrading from v6.10.0 to v6.19.1 I noticed that the `Request-Id` header does not exist anymore. Instead, you have to use `request-id`. 

Why was this breaking change not documented and included in a new major?

ob-stripe:
cc @stripe/api-libraries 

Store HTTP response headers in case-insensitive array. HTTP headers are supposed to be case-insensitive, but since we stored them in a regular PHP array, they were actually case-sensitive.

This is particularly problematic since the API will return headers with different cases depending on whether HTTP/2 is used or not (since HTTP/2 requires that all headers be in lowercase).

Fixes #529.

### DIFF
--- a/lib/Util/CaseInsensitiveArray.php
+++ b/lib/Util/CaseInsensitiveArray.php
@@ -1,0 +1,62 @@
+<?php
+
+namespace Stripe\Util;
+
+use ArrayAccess;
+
+/**
+ * CaseInsensitiveArray is an array-like class that ignores case for keys.
+ *
+ * It is used to store HTTP headers. Per RFC 2616, section 4.2:
+ * Each header field consists of a name followed by a colon (":") and the field value. Field names
+ * are case-insensitive.
+ *
+ * In the context of stripe-php, this is useful because the API will return headers with different
+ * case depending on whether HTTP/2 is used or not (with HTTP/2, headers are always in lowercase).
+ */
+class CaseInsensitiveArray implements ArrayAccess
+{
+    private $container = array();
+
+    public function __construct($initial_array = array())
+    {
+        $this->container = array_map("strtolower", $initial_array);
+    }
+
+    public function offsetSet($offset, $value)
+    {
+        $offset = static::maybeLowercase($offset);
+        if (is_null($offset)) {
+            $this->container[] = $value;
+        } else {
+            $this->container[$offset] = $value;
+        }
+    }
+
+    public function offsetExists($offset)
+    {
+        $offset = static::maybeLowercase($offset);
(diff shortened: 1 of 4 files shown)


ob-stripe:
@brandur-stripe Also, could you check if stripe-go is affected by this issue (since it's the only other library that uses HTTP/2)? (My guess is that it's not because the `http` package canonicalizes header names.)

lib/Util/CaseInsensitiveArray.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+namespace Stripe\Util;
+
+use ArrayAccess;
+
+class CaseInsensitiveArray implements ArrayAccess
```
brandur-stripe: Do you think we could put a comment block in here that mentions that this exists to provide case insensitive access to response headers because their case may vary depending on whether the request came back over HTTP or HTTP/2 (which mandates lowercase)? I think it may be helpful in reasoning about why it's here in the future.

brandur-stripe:
Nice implementation!

> Also, could you check if stripe-go is affected by this issue (since it's the only other library that uses HTTP/2)? (My guess is that it's not because the http package canonicalizes header names.)

Yeah, it looks like you found it already, but the `Response` struct's `Headers` struct canonicalizes naming:

``` go
        // Header maps header keys to values. If the response had multiple
        // headers with the same key, they may be concatenated, with comma
        // delimiters.  (RFC 7230, section 3.2.2 requires that multiple headers
        // be semantically equivalent to a comma-delimited sequence.) When
        // Header values are duplicated by other fields in this struct (e.g.,
        // ContentLength, TransferEncoding, Trailer), the field values are
        // authoritative.
        //
        // Keys in the map are canonicalized (see CanonicalHeaderKey).
        Header Header
```

And `CanonicalHeaderKey`:

``` go
CanonicalHeaderKey returns the canonical format of the header key s. The canonicalization converts the first letter and any letter following a hyphen to upper case; the rest are converted to lowercase. For example, the canonical key for "accept-encoding" is "Accept-Encoding". If s contains a space or invalid header field bytes, it is returned without modifications.
```

So we're okay there.

r? @ob-stripe

ob-stripe:
I added a comment as requested, ptal @brandur-stripe. (And thanks for confirming that stripe-go is unaffected!)

brandur-stripe:
Thanks OB!

LGTM.

ob-stripe:
Released as 6.19.2.

PR #534, opened 2018-10-11
Add FileUpload.php file to fix autoloading

ob-stripe:
cc @stripe/api-libraries 

Adds a `FileUpload.php` file so that PSR-4 autoloading for the `\Stripe\FileUpload` class works correctly.

Fixes the issue reported by @ruudk in https://github.com/stripe/stripe-php/pull/520#issuecomment-428934556.

I can't think of a way to write automated tests for this, but I've verified manually that it works:

- before the fix:
```
php > require("vendor/autoload.php");
php > echo class_exists('Stripe\FileUpload');
php > echo class_exists('Stripe\File');
1
php > echo class_exists('Stripe\FileUpload');
1
```

- after the fix:
```
php > require("vendor/autoload.php");
php > echo class_exists('Stripe\FileUpload');
1
```

### DIFF
--- a/init.php
+++ b/init.php
@@ -78,6 +78,7 @@ require(dirname(__FILE__) . '/lib/Event.php');
 require(dirname(__FILE__) . '/lib/ExchangeRate.php');
 require(dirname(__FILE__) . '/lib/File.php');
 require(dirname(__FILE__) . '/lib/FileLink.php');
+require(dirname(__FILE__) . '/lib/FileUpload.php');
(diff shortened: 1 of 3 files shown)


brandur-stripe:
Nice! I was following along on the thread, and it's great to have found a solution so quickly.

LGTM.

ob-stripe:
Thanks @brandur-stripe! All credit goes to @ruudk here :)

ob-stripe:
Released as 6.19.4.

PR #542, opened 2018-10-31
Add support for Radar List and ListItem resources

remi-stripe:
cc @stripe/api-libraries 

cc @ob-stripe as `List` is forbidden as a class name in PHP so `List` as an object in Radar was likely a bad idea :p

### DIFF
--- a/init.php
+++ b/init.php
@@ -98,6 +98,8 @@ require(dirname(__FILE__) . '/lib/Payout.php');
 require(dirname(__FILE__) . '/lib/Person.php');
 require(dirname(__FILE__) . '/lib/Plan.php');
 require(dirname(__FILE__) . '/lib/Product.php');
+require(dirname(__FILE__) . '/lib/Radar/List.php');
+require(dirname(__FILE__) . '/lib/Radar/ListItem.php');
 require(dirname(__FILE__) . '/lib/Recipient.php');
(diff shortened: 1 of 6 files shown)


ob-stripe:
Yep. We'll probably have to rename it to `RadarList` or something.

tests/Stripe/Radar/ValueListItemTest.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+namespace Stripe\Radar;
+
+class ValueListItemTest extends \Stripe\TestCase
+{
+    const TEST_RESOURCE_ID = 'rsli_123';
+
+    public function testIsValueListItemable()
```
ob-stripe:
```suggestion
    public function testIsListable()
```

tests/Stripe/Radar/ValueListTest.php
```diff
@@ -0,0 +1,77 @@
+<?php
+
+namespace Stripe\Radar;
+
+class ValueListTest extends \Stripe\TestCase
+{
+    const TEST_RESOURCE_ID = 'rsl_123';
+
+    public function testIsValueListable()
```
ob-stripe:
```suggestion
    public function testIsListable()
```

remi-stripe:
Damn search and replace. Fixed, PTAL @ob-stripe

PR #544, opened 2018-11-03
Phpdoc updates and new Event constants

nickdnk:
Hello guys

I found the following minor changes to phpdocs and added a bunch of missing Event class constants, namely those related to payment intent, subscription schedule, reporting, issuing/issuer and subscription schedule. Unfortunately I had to indent the events because of a longer-than-usual name, which means you'll have to review the list without white-space changes to see what's happened.

The only thing I'm not sure about is that `on_behalf_of` doesn't seem to exist on `PaymentIntent` on the newest version of the API. Has this been a mistake from the beginning or has it been removed?

### DIFF
--- a/lib/Event.php
+++ b/lib/Event.php
@@ -23,11 +23,12 @@ class Event extends ApiResource
 
     const OBJECT_NAME = "event";
 
-     /**
+    /**
      * Possible string representations of event types.
      * @link https://stripe.com/docs/api#event_types
      */
     const ACCOUNT_UPDATED                      = 'account.updated';
+    const ACCOUNT_APPLICATION_AUTHORIZED       = 'account.application.authorized';
     const ACCOUNT_APPLICATION_DEAUTHORIZED     = 'account.application.deauthorized';
     const ACCOUNT_EXTERNAL_ACCOUNT_CREATED     = 'account.external_account.created';
     const ACCOUNT_EXTERNAL_ACCOUNT_DELETED     = 'account.external_account.deleted';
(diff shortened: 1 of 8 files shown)


lib/Event.php
```diff
@@ -23,96 +23,128 @@ class Event extends ApiResource
...
+    const ISSUING_TRANSACTION_UPDATED              = 'issuing_transaction.updated';
+    const ORDER_CREATED                            = 'order.created';
+    const ORDER_PAYMENT_FAILED                     = 'order.payment_failed';
+    const ORDER_PAYMENT_SUCCEEDED                  = 'order.payment_succeeded';
+    const ORDER_UPDATED                            = 'order.updated';
+    const ORDER_RETURN_CREATED                     = 'order_return.created';
+    const PAYMENT_INTENT_AMOUNT_CAPTURABLE_UPDATED = 'payment_intent.amount_capturable_updated';
+    const PAYMENT_INTENT_CREATED                   = 'payment_intent.created';
+    const PAYMENT_INTENT_PAYMENT_FAILED            = 'payment_intent.payment_failed';
+    const PAYMENT_INTENT_REQUIRES_CAPTURE          = 'payment_intent.requires_capture';
```
remi-stripe: Can you remove this one? We're actively deprecating it.

lib/PaymentIntent.php
```diff
@@ -22,10 +23,10 @@
  * @property string $currency
  * @property string $customer
  * @property string $description
+ * @property mixed $last_payment_error
  * @property bool $livemode
  * @property StripeObject $metadata
  * @property mixed $next_source_action
- * @property string $on_behalf_of
```
remi-stripe: can you leave that one? It's gated but supported.

lib/Person.php
```diff
@@ -35,6 +35,7 @@
  */
 class Person extends ApiResource
 {
+
```
remi-stripe: Can you remove this empty line?

lib/Event.php
nickdnk: Roger.

lib/Person.php
nickdnk: Hey. I only added it because all the other classes have it. Should we remove it from all classes to be consistent? I know it's really nitpicking, but still.

lib/PaymentIntent.php
nickdnk: Done

lib/Person.php
remi-stripe: huh that is a weird one. I guess it does not hurt to leave as is on this one after all

remi-stripe:
Thanks for the quick and thorough work as usual @nickdnk!

PR #546, opened 2018-11-05
Loose package weight

ankurk91:
Hi,

Let's not publish all those file.
`.gitattributes` can be used to exclude files from final archive that is downloaded by composer.

Read more
- https://madewithlove.be/gitattributes/
- https://alexbilbie.com/2012/11/exclude-objects-with-gitattributes/
- http://www.pixelite.co.nz/article/using-git-attributes-exclude-files-your-release/

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,12 @@
+# https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes
+
+# A list of files and folders those will be excluded from release archive
+/.coveralls.yml export-ignore
+/examples export-ignore
+/.github export-ignore
+/.gitattributes export-ignore
+/phpunit.xml export-ignore
+/phpunit.no_autoload.xml export-ignore
+/tests export-ignore
+/.travis.yml export-ignore
+/.vscode export-ignore


brandur-stripe:
@ankurk91 Thanks! Do you know if it'd be possible to make this a whitelist instead of a blacklist like you can do with [`.gitignore`](https://jasonstitt.com/gitignore-whitelisting-patterns)?

I generally find this strategy to be more effective because you can just whitelist the most important stuff like `lib/`, and then don't have to remember to add new things to `.gitattributes` every time someone adds a new top level file.

ankurk91:
`.gitignore` file prevent files and folders from being published in git repository.

While `.gitattributes` file tells git to ignore files and folders during export.
So `.gitattribues` will not prevent listed files from being pushed in git repo. It will only affect downloads by composer.

https://git-scm.com/docs/git-archive#ATTRIBUTES

Most of the packages already following same pattern see
https://github.com/guzzle/guzzle/blob/master/.gitattributes

brandur-stripe:
@ankurk91 Ah, yes I understand the difference betweens `.gitignore` and `.gitattributes`. What I was asking is whether `.gitattributes` could support a whitelist like you can do with `.gitignore`. Check that link in my comment above for details, but the whitelist-style `.gitignore` looks a little like this (for example):

```
# Ignore everything
*
# But descend into directories
!*/
# Recursively allow files under subtree
!/subtree/**
# You can be specific with these rules
!/some/other/deep/path/**
!.gitignore
```

You basically start with a coarse rule that says "ignore all", then you pull individual files or directories back in for consideration.

ankurk91:
I got your point. Unfortunately I couldn't find option to whitelist files in .gitattributes.

We can test it locally by creating the . gitattributes files and running `git archive` command.

Now check the contents of the zip file

ankurk91:
@brandur-stripe 
I just tested the whitelisting pattern in `.gitattributes` file and it didn't worked.

git throws an error saying
```
warning: Negative patterns are ignored in git attributes
Use '\!' for literal leading exclamation.
```

.gitattributes
```diff
@@ -0,0 +1,9 @@
+/.github export-ignore
+/.vscode export-ignore
+/tests export-ignore
+/examples export-ignore
+.coveralls.yml export-ignore
```
brandur-stripe: A couple minor asks:

1. Since you're using the prefixing `/` for the first set of these, would you mind just using it for all of them?
2. Can you alphabetize the list?
3. Can you add a short comment to the top of the file indicating that the purpose of this list is to make sure that these files are excluded from the final Composer package.
4. Can you squash down your commits?

brandur-stripe:
> I just tested the whitelisting pattern in `.gitattributes` file and it didn't worked.

Cool. Thanks for taking a look at that anyway!

Added a few asks above, but otherwise this looks fine. Thanks!

.gitattributes
ankurk91: @brandur-stripe 
Done

brandur-stripe:
Looks good. Thanks!

PR #550, opened 2018-11-14
Enable persistent connections

ob-stripe:
cc @stripe/api-libraries 

Instead of creating and closing a new curl handle for each request, create a curl handle once and reuse it for all requests. This should offer a significant performance improvement, since curl will reuse the connection.

Unfortunately, it's difficult to write automated tests for this given the current shape of `CurlClient` (not really possible to test private static methods). I've ran some manual tests though, and things look pretty good.

This should also be thread-safe. From https://secure.php.net/manual/en/intro.pthreads.php:
> Static Members: When a new context is created ( Thread or Worker ), they are generally copied, but resources and objects with internal state are nullified (for safety reasons).

Because `CurlClient` is instantiated as a singleton via `CurlClient::instance()`, each thread gets its own client instance (and thus its own curl handle). I've verified that running multiple threads in parallel doesn't cause issues when persistent connections are enabled.

### DIFF
--- a/lib/HttpClient/CurlClient.php
+++ b/lib/HttpClient/CurlClient.php
@@ -40,6 +40,10 @@ class CurlClient implements ClientInterface
 
     protected $userAgentInfo;
 
+    protected $enablePersistentConnections = true;
+
+    protected $curlHandle = null;
+
     /**
      * CurlClient constructor.
      *
@@ -60,6 +64,11 @@ class CurlClient implements ClientInterface
         $this->initUserAgentInfo();
     }
 
+    public function __destruct()
+    {
+        $this->closeCurlHandle();
+    }
+
     public function initUserAgentInfo()
     {
         $curlVersion = curl_version();
@@ -77,6 +86,22 @@ class CurlClient implements ClientInterface
     public function getUserAgentInfo()
     {
         return $this->userAgentInfo;
+    }
+
+    /**
+     * @return boolean
+     */
+    public function getEnablePersistentConnections()
+    {
+        return $this->enablePersistentConnections;
+    }
+
+    /**
+     * @param boolean $enable
+     */
+    public function setEnablePersistentConnections($enable)
+    {
+        $this->enablePersistentConnections = $enable;
     }
 
     // USER DEFINED TIMEOUTS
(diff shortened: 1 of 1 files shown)


lib/HttpClient/CurlClient.php
```diff
@@ -218,17 +229,24 @@ private function executeRequestWithRetries($opts, $absUrl)
             $rcode = 0;
             $errno = 0;
 
-            $curl = curl_init();
-            curl_setopt_array($curl, $opts);
-            $rbody = curl_exec($curl);
+            if (!is_null($this->curlHandle) && Stripe::getEnablePersistentConnections()) {
+                curl_reset($this->curlHandle);
+            } else {
+                $this->curlHandle = curl_init();
```
remi-stripe: Shouldn't we close the old handle before creating a new one? Your first if means that you can have it initialized and not cleaned up (for example with an exception on the previous request) right?

lib/Stripe.php
```diff
@@ -152,6 +155,22 @@ public static function setCABundlePath($caBundlePath)
...
+    /**
+     * @return boolean
+     */
+    public static function getEnablePersistentConnections()
+    {
+        return self::$enablePersistentConnections;
+    }
+
+    /**
+     * @param boolean $verify
```
remi-stripe: wrong comment

ob-stripe: 👍

lib/HttpClient/CurlClient.php
ob-stripe: 👍

brandur-stripe:
I don't 100% understand the static thread safety thing — it kind of sounds like we're sort of hoping that a cURL connection is correctly identified as stateful and therefore copied safely. That said, if testing yielded good results I guess it's fine.

LGTM.

ob-stripe:
> it kind of sounds like we're sort of hoping that a cURL connection is correctly identified as stateful and therefore copied safely.

More precisely, it's the static `CurlClient` instance that's identified as stateful and nullified, so the first time `CurlClient::instance()` is called in a new thread, a new `CurlClient` instance will be created.

PR #552, opened 2018-11-15
stripe-php v7.0.0

### LINKED ISSUE #587: PHP 7.2 Warnings
Hi, I when I use WP-Enginge's php 7.2 Compatibility Checker (https://wordpress.org/plugins/php-compatibility-checker/), get the following warnings:

```
FILE: /nas/content/staging/lxbplatform/wp-content/plugins/lxb-account/lib/stripe/lib/StripeObject.php\
---------------------------------------------------------------------------------------------------------------------------------------------------------\
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 2 LINES\
---------------------------------------------------------------------------------------------------------------------------------------------------------\
 293 | WARNING | Method name "StripeObject::__toJSON" is discouraged; PHP has reserved all method names with a double underscore prefix for future use.\
 308 | WARNING | Method name "StripeObject::__toArray" is discouraged; PHP has reserved all method names with a double underscore prefix for future use.\
---------------------------------------------------------------------------------------------------------------------------------------------------------\
```

### LINKED ISSUE #702: Util::convertToStripeObject and Util::convertStripeObjectToArray don't seem to round-trip
Stripe-PHP version: v6.43.0
PHP version: 7.2.21
PHPUnit version: 8.3.4

It'd probably be easier to express what I've tripped over as executable test cases (if these help chasing it down, then please use them)

```php
    public function testRoundTripFromObject()
    {
        $array = ['foo' => 'bar'];

        $stripeObject = \Stripe\Util\Util::convertToStripeObject($array, []);

        $final = \Stripe\Util\Util::convertStripeObjectToArray($stripeObject);

        $this->assertEquals($array, $final);
    }

    public function testRoundTripFromArray()
    {
        $array = ['foo' => 'bar'];

        $stripeObject = \Stripe\StripeObject::constructFrom($array, []);

        $final = \Stripe\Util\Util::convertStripeObjectToArray($stripeObject);

        $this->assertEquals($array, $final);
    }
```

On my environment (referenced above), both of those tests are failing, with convertStripeObjectToArray returning an empty array in both cases.  The second test tries to narrow down the scope of what could go wrong when the StripeObject is being created.

Upon digging a bit further into convertStripeObjectToArray, the foreach loop on line 43 of \Stripe\Util\Util does not seem to be firing in either of these tests.

### LINKED ISSUE #723: Array Functions on Objects Soon to be Deprecated
Objects currently behave fairly close to Arrays, however PHP 7.4 will be adding a notice of deprecation for many Array-purpose-functions on Objects. It would probably be a good idea to make a few preemptive changes before 7.4 is released.
So far, I've only noticed something that would be a problem on [line 402](https://github.com/stripe/stripe-php/blob/master/lib/ApiRequestor.php#L402) of `lib/ApiRequestor.php`, though there may be other instances. `array_key_exists` will throw a warning because `$rheaders` is an Object, not an Array.

ob-stripe:
cc @stripe/api-libraries 

WIP PR for the next major release.

⚠️ = breaking change

- [x] ⚠️ Drop support for PHP 5.4 (#551) 
- [x] ⚠️ Drop support for PHP 5.5 (#554) 
- [x] Bump dependencies (#553)
- [x] Remove `CURLFile` check (#555)
- [x] Update constant definitions for PHP >= 5.6 (#556)
- [x] ⚠️ Remove `FileUpload` alias (#557)
- [x] Remove `curl_reset` check (#570)
- [x] Use `\Stripe\<class>::class` constant instead of strings (#643)
- [x] Use `array_column` to flatten params (#686)
- [x] ⚠️ Remove deprecated methods (#692)
- [x] ⚠️ Remove `IssuerFraudRecord` (#696)
- [x] Update constructors of Stripe exception classes (#559)
- [x] Fix remaining TODOs (#700)
- [x] Use yield for autopagination (#703)
- [x] ⚠️ Rename fake magic methods and rewrite array conversion (#704)
- [x] Add `ErrorObject` to Stripe exceptions (#705)
- [x] Start using PHP CS Fixer (#706)
- [x] Update error messages for nested resource operations (#708)
- [x] Upgrade retry logic (#707)
- [x] ⚠️ `Collection` improvements / fixes (#715)
- [x] ⚠️ Modernize exceptions (#709)
- [x] Add constants for error codes (#716)
- [x] Update certificate bundle (#717)
- [x] Retry requests on a 429 that's a lock timeout (#718)
- [x] Fix `toArray()` calls (#719)
- [x] Couple of fixes for PHP 7.4 (#725)

Fixes #587 (via #704), fixes #702 (via #704), fixes #723 (via #725).


browner12:
I'm not sure if this is planned already, but since this is targeting a new major version, I would suggest dropping support for PHP 5.6 and PHP 7.0 while you're at it.  Both of these version will be losing security support by the end of the year.

http://php.net/supported-versions.php

Would be great to see a new major version more forward with only supported PHP versions, and allow users who require older PHP versions to stay on v6.

ob-stripe:
Hey @browner12, thanks for the feedback!

I'm aware that security support for 5.6 and 7.0 will be ending soon, but our usage metrics show that usage for these versions is still very high: half of our PHP users that are currently on the 6.x branch are using one of these (25% on 5.6 and 25% on 7.0).

As it is, dropping support for 5.4 and 5.5 will prevent 6.5% of users (again, counting only users on the 6.x branch) from upgrading to 7.0 until they upgrade to a more recent PHP version. Bumping that number to 56.5% seems a little bit too steep for now.

That said, we will reevaluate in 2019 once support for 5.6 and 7.0 has actually ended, and might revisit this decision then.

browner12:
Thanks for the response. Are you saying you guys would consider another major version (v8) after the new year?

I guess I don't see the downside of dropping the additional PHP versions. Even though the percentage of users on unsupported PHP versions is high, they will still be able to use this package by staying on v6. Users on supported PHP versions would be able to upgrade to v7.

I would say in the package community this is a pretty common practice of branching to a new major version and dropping old PHP support, which also helps drive adoption of newer versions of PHP.

ob-stripe:
@browner12 Replying a bit late, sorry! As a developer, I agree with you entirely. However, many of Stripe's PHP users are not developers or even very technical persons and rely on e.g. Wordpress plugins for their Stripe integrations. It's an unfortunate reality that many hosting providers are not very diligent about keeping their PHP versions up to date, so dropping support for all but officially supported PHP versions would be a pretty harsh move for those users and would go against Stripe's philosophy of minimizing friction in accepting online payments.

As said above, we'll certainly reconsider this decision in a few months, and might release a new major version then.

browner12:
Definitely agree, and understand where you're coming from.

I thought since this PR was for a Major release, it would be a good time to drop unsupported (or soon to be unsupported) versions of PHP. 

People on older versions of PHP would still be able to use Stripe by sticking to v6, while other users could hop up to v7.  Because Stripe is so popular, this would also add some pressure for people to start upgrading to a supported version of PHP.

You guys are also in a unique position because you accept payments meaning security is paramount to you (and should be to your users). When I see a website that accepts payments via Stripe, it gives me an extra bit of confidence that my information is secure. But knowing the website could be using Stripe, but also still be on an older version of PHP takes that confidence away.

We could discuss this forever, so I'll just say keep up the good work building this great product.

rvitaliy:
@ob-stripe 
Since PHP dropped the `5.6` & `7.0` support, I would suggest upgrading to `7.1` [see supported versions](http://php.net/supported-versions.php).
So you can finally start using the types 🚀 and make this lib more secure 🔒

ob-stripe:
@rvitaliy Thanks! However, we're going to keep supporting 5.6 and 7.0 for a bit longer despite these versions having reached end-of-life. See the discussion above for some justification.

ob-stripe:
@brandur-stripe @remi-stripe I think this is more or less ready! I just need to finish writing up the migration guide: https://github.com/stripe/stripe-php/wiki/%5BWIP%5D-Migration-guide-for-v7.

Is there anything else you think we need to include in this release?

remi-stripe:
Nothing comes to mind. I'm sure we'll think of something the day after release as usual :)

brandur-stripe:
Haha, same Olivier. You already blew past any ideas that I had by about 10x.

Very exciting that we're getting close though!! Lemme know if there's anything else I can do to help get it across the finish line.

ob-stripe:
Pulling this in!

ob-stripe:
Released as v7.0.0 (obviously). Thanks everyone!

PR #555, opened 2018-11-15
Remove CURLFile check

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

The [`CURLFile`](https://secure.php.net/manual/en/class.curlfile.php) class is available as of PHP 5.5, so we can remove the check.


remi-stripe:
LGTM

PR #559, opened 2018-11-16
Update constructors of Stripe exception classes

ob-stripe:
r? @remi-stripe @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -193,31 +193,32 @@ class ApiRequestor
         $param = isset($errorData['param']) ? $errorData['param'] : null;
         $code = isset($errorData['code']) ? $errorData['code'] : null;
         $type = isset($errorData['type']) ? $errorData['type'] : null;
+        $declineCode = isset($errorData['decline_code']) ? $errorData['decline_code'] : null;
 
         switch ($rcode) {
             case 400:
                 // 'rate_limit' code is deprecated, but left here for backwards compatibility
                 // for API versions earlier than 2015-09-08
                 if ($code == 'rate_limit') {
-                    return new Error\RateLimit($msg, $param, $rcode, $rbody, $resp, $rheaders);
+                    return new Error\RateLimit($msg, $param, $rcode, $rbody, $resp, $rheaders, $code);
                 }
                 if ($type == 'idempotency_error') {
-                    return new Error\Idempotency($msg, $rcode, $rbody, $resp, $rheaders);
+                    return new Error\Idempotency($msg, $rcode, $rbody, $resp, $rheaders, $code);
                 }
 
                 // no break
             case 404:
-                return new Error\InvalidRequest($msg, $param, $rcode, $rbody, $resp, $rheaders);
+                return new Error\InvalidRequest($msg, $param, $rcode, $rbody, $resp, $rheaders, $code);
             case 401:
-                return new Error\Authentication($msg, $rcode, $rbody, $resp, $rheaders);
+                return new Error\Authentication($msg, $rcode, $rbody, $resp, $rheaders, $code);
             case 402:
-                return new Error\Card($msg, $param, $code, $rcode, $rbody, $resp, $rheaders);
+                return new Error\Card($msg, $param, $code, $rcode, $rbody, $resp, $rheaders, $declineCode);
             case 403:
-                return new Error\Permission($msg, $rcode, $rbody, $resp, $rheaders);
+                return new Error\Permission($msg, $rcode, $rbody, $resp, $rheaders, $code);
             case 429:
-                return new Error\RateLimit($msg, $param, $rcode, $rbody, $resp, $rheaders);
+                return new Error\RateLimit($msg, $param, $rcode, $rbody, $resp, $rheaders, $code);
             default:
-                return new Error\Api($msg, $rcode, $rbody, $resp, $rheaders);
+                return new Error\Api($msg, $rcode, $rbody, $resp, $rheaders, $code);
         }
     }
 
--- a/lib/Error/Base.php
+++ b/lib/Error/Base.php
@@ -4,39 +4,39 @@ namespace Stripe\Error;
 
 use Exception;
 
+/**
+ * Base is the base error from which all other more specific Stripe errors derive.
+ *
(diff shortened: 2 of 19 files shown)


ob-stripe:
@remi-stripe Bump! :)

lib/ApiRequestor.php
```diff
@@ -164,31 +164,32 @@ private static function _specificAPIError($rbody, $rcode, $rheaders, $resp, $err
...
                 // intentional fall-through
             case 404:
-                return new Error\InvalidRequest($msg, $param, $rcode, $rbody, $resp, $rheaders);
+                return new Error\InvalidRequest($msg, $param, $rcode, $rbody, $resp, $rheaders, $code);
             case 401:
-                return new Error\Authentication($msg, $rcode, $rbody, $resp, $rheaders);
+                return new Error\Authentication($msg, $rcode, $rbody, $resp, $rheaders, $code);
             case 402:
-                return new Error\Card($msg, $param, $code, $rcode, $rbody, $resp, $rheaders);
+                return new Error\Card($msg, $param, $code, $rcode, $rbody, $resp, $rheaders, $declineCode);
```
remi-stripe: Why does this pass `declineCode` and all the others pass `$code`? I don't fully grasp what you are trying to do here to be honest.

remi-stripe:
@ob-stripe just found this PR, do you know what the next steps are on that one?

lib/ApiRequestor.php
ob-stripe: `Error\Card` was the only class that already accepted the `code` parameter (as its 3rd argument).

When we introduced error codes for non-card errors, we didn't want to modify the constructors so we extracted the error code from the JSON body (cf. https://github.com/stripe/stripe-php/pull/559/files#diff-600ddc684b71253491fd6da75bb5f470L25).

`Error\Card` did something similar for decline codes (cf. https://github.com/stripe/stripe-php/pull/559/files#diff-0d197aceb8c08f2f4ddf6a9b6d0a0a68L28).

This PR updates all constructors so that both code (for all errors) and decline codes (for card errrors) are passed explicitly as constructor arguments.

ob-stripe:
ptal @remi-stripe 

I'd like to resume work on stripe-php v7. Once this PR is merged I'll do something similar to https://github.com/stripe/stripe-ruby/pull/811 to allow for accessing the error object from the exception.

lib/Error/InvalidRequest.php
```diff
@@ -2,17 +2,25 @@
 
 namespace Stripe\Error;
 
+/**
+ * InvalidRequest is raised when a request is initiated with invalid  parameters.
```
brandur-stripe: Minor, but there's an extra space in here between the last two words.

lib/Error/OAuth/InvalidClient.php
```diff
@@ -3,7 +3,11 @@
 namespace Stripe\Error\OAuth;
 
 /**
- * InvalidClient is raised when authentication fails.
+ * InvalidClient is raised when the client doesn't belong to you, or the API key mode (live or test)
+ * doesn't match the client mode. Or the stripe_user_id doesn't exist or isn't connected to your
+ * application.
```
brandur-stripe: Not sure if this was copied from elsewhere, but the grammar is a little awkward. It reads "InvalidClient is raised if ..., or .... Or, [new sentence] ..."

It should probably be all one sentence with just a comma between the first and second possibilities and a single "or" between the second and third.

lib/Error/Base.php
```diff
@@ -4,66 +4,75 @@
...
+    protected $stripeCode;
+
     public function __construct(
         $message,
         $httpStatus = null,
         $httpBody = null,
         $jsonBody = null,
-        $httpHeaders = null
+        $httpHeaders = null,
+        $stripeCode = null
```
brandur-stripe: Any particular reason to call this "stripe code" here, and just "code" above? Should we just call it "code" to be consistent with what's in the API reference/JSON structure?

ob-stripe: We can't call it `code` because the `Exception` class already has an integer attribute (and accessors) with that name: https://www.php.net/manual/en/class.exception.php

ob-stripe:
Fixed the comments and explained why the `code` field needs to be named `stripeCode` in the class.

lib/Error/Base.php
brandur-stripe: Fuuuu. Okay, I don't have too strong of opinion here, but given we can't call it "code" anyway and OAuth errors already have an "error code", should we call it `$errorCode`/`$getErrorCode` instead?

The actual JSON object fields are named differently (`code` in normal errors versus `error` in OAuth errors I think), but it probably wouldn't hurt just to have the naming the same in PHP. `OAuthBase` extends `Error\Base`, but you could probably promote the existing `$errorCode` up to the parent.

Again, no strong opinion, but just wanted to make sure that you considered the possibility.

ob-stripe: So my plan is to do something similar to https://github.com/stripe/stripe-ruby/pull/811 as a followup to this PR, i.e. expose the Stripe error "API resource" on the exception via `getError()` or something similar. On the API resource, we won't have this naming constraint so people will be able to do `$e->getError()->getCode()`.

On the exception object itself, I think we might as well keep the current name to avoid an unnecessary breaking change. wdyt? (Note that this will be released in a new major version, so breaking changes are an option if we think it's for the best.)

brandur-stripe: > So my plan is to do something similar to https://github.com/stripe/stripe-ruby/pull/811 as a followup to this PR, i.e. expose the Stripe error "API resource" on the exception via `getError()` or something similar. On the API resource, we won't have this naming constraint so people will be able to do `$e->getError()->getCode()`.

Ah, I see! Yeah, that sounds great.

> On the exception object itself, I think we might as well keep the current name to avoid an unnecessary breaking change. wdyt? (Note that this will be released in a new major version, so breaking changes are an option if we think it's for the best.)

Oh man, there was enough changes in that file that I didn't even notice that `$stripeCode` /`$getStripeCode` existed before — I thought it was new. +1 that a breaking change isn't worth it. Ship away!

PR #562, opened 2018-11-27
Add support for Review resource.

remi-stripe:
cc @stripe/api-libraries 

This is tagged as WIP as we're missing a release of stripe-mock and the route in the spec is currently incorrect (/v1/reviews)


zanker-stripe:
LGTM

PR #566, opened 2018-12-03
Fix dangling message contents for Card::update

ryangreenberg:
I was using some AST tooling to identify semantically empty statements in our code base and it identified this statement which (I think) is intended to be concatenated with the previous lines.


ob-stripe:
Nice catch! Thanks @ryangreenberg :)

PR #568, opened 2018-12-11
Enable persistent connections (if possible)

ob-stripe:
cc @stripe/api-libraries 

Like #550, but without a breaking change. If running on PHP 5.4, persistent connections simply won't be enabled.


brandur-stripe:
Nice!

LGTM.

ob-stripe:
Released as 6.26.0.

PR #570, opened 2018-12-11
Remove curl_reset check

ob-stripe:
cc @stripe/api-libraries 

Remove `curl_reset` check. This was only necessary as long as we supported PHP 5.4.


brandur-stripe:
LGTM.

PR #571, opened 2018-12-19
Add support for Checkout Sessions

remi-stripe:
This adds support for the Checkout session resource in beta.

We also add the corresponding beta versions of Stripe's OpenAPI's spec and fixtures

cc @stripe/api-libraries @matt-stripe

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ php:
 
 env:
   global:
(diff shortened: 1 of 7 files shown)


remi-stripe:
stripe-mock has been updated and we're merging the dynamic libraries.

r? @ob-stripe

PR #572, opened 2018-12-19
Detect custom OpenAPI files and start stripe-mock from test suite

ob-stripe:

Like https://github.com/stripe/stripe-ruby/pull/715, but for stripe-php.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,7 @@ before_install:
   - |
     stripe-mock/stripe-mock_${STRIPE_MOCK_VERSION}/stripe-mock > /dev/null &
     STRIPE_MOCK_PID=$!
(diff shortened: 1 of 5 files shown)


ob-stripe:
Alright, I've updated this to be consistent with what we settled on in https://github.com/stripe/stripe-ruby/pull/715.

The implementation is slightly different because stripe-php supports talking to a stripe-mock instance running on a different host (cf. #573). That's a pretty neat feature and we should probably port it to other libraries.

ob-stripe:
ptal @brandur-stripe @remi-stripe

remi-stripe:
👍 from me!

brandur-stripe:
LGTM as well!

PR #574, opened 2019-01-02
Added Subscription::$default_source phpdoc property

kimhemsoe:
Hi

Added missing Subscription::$default_source phpdoc property


kimhemsoe:
@remi-stripe Yes ofc and done :)

PR #582, opened 2019-01-21
Update LICENSE

Josh1Dev:


ob-stripe:
Hey @Josh1Dev, thanks for the PR!

I'll reach out internally to see if this is the correct copyright notice. I think it is, but we might want to remove the current year so that we don't have to update the license every year. I'll reply here as soon as I have confirmation.

Josh1Dev:
@ob-stripe Sure! Just a small contribution : )

ob-stripe:
Pulling this in. Thanks @Josh1Dev!

PR #590, opened 2019-01-30
Add support for Subscription Schedules

remi-stripe:
This adds two new resources: Subscription Schedule and Subscription Schedule Revision. The latter is nested under the former.

This should be ready even though we won't merge until all libraries are ready and stripe-mock supports those new endpoints publicly.

r? @ob-stripe 
cc @stripe/api-libraries @alexander-stripe

### DIFF
--- a/lib/SubscriptionScheduleRevision.php
+++ b/lib/SubscriptionScheduleRevision.php
@@ -1,0 +1,77 @@
+<?php
+
+namespace Stripe;
+
+/**
+ * Class Subscription
+ *
+ * @property string $id
+ * @property string $object
+ * @property int $created
+ * @property mixed $invoice_settings
+ * @property boolean $livemode
+ * @property mixed $phases
+ * @property string $previous_revision
+ * @property string $renewal_behavior
+ * @property mixed $renewal_interval
+ * @property string $chedule
+ *
+ * @package Stripe
+ */
+class SubscriptionScheduleRevision extends ApiResource
+{
+
+    const OBJECT_NAME = "subscription_schedule_revision";
+
+    use ApiOperations\All;
+    use ApiOperations\Retrieve;
+
+    /**
+     * @return string The API URL for this Subscription Schedule Revision.
+     */
+    public function instanceUrl()
+    {
+        $id = $this['id'];
+        $schedule = $this['schedule'];
+        if (!$id) {
+            throw new Error\InvalidRequest(
+                "Could not determine which URL to request: " .
+                "class instance has invalid ID: $id",
+                null
+            );
+        }
+        $id = Util\Util::utf8($id);
+        $schedule = Util\Util::utf8($schedule);
+
+        $base = SubscriptionSchedule::classUrl();
+        $scheduleExtn = urlencode($schedule);
+        $extn = urlencode($id);
+        return "$base/$scheduleExtn/revisions/$extn";
+    }
+
+    /**
+     * @param array|string $_id
+     * @param array|string|null $_opts
+     *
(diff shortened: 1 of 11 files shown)


lib/SubscriptionScheduleRevision.php
```diff
@@ -0,0 +1,77 @@
...
+
+    const OBJECT_NAME = "subscription_schedule_revision";
+
+    use ApiOperations\All;
+    use ApiOperations\Retrieve;
+
+    /**
+     * @return string The API URL for this Subscription Schedule Revision.
+     */
+    public function instanceUrl()
```
remi-stripe: OOC do we really need this?

lib/SubscriptionScheduleRevision.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        $id = $this['id'];
+        $schedule = $this['schedule'];
```
remi-stripe: Soooo I just realized we do a lot of magic in the library to find the parent id. And I worry a liiiiittle bit that this is not documented anywhere (internally at Stripe). And that one day we would rename this to `subscription_schedule` because it seems more consistent and then break half the libraries out there.

I wonder if settling on `parent` for the name across all parent resources would be safer. Just food for thoughts

lib/SubscriptionScheduleRevision.php
```diff
@@ -0,0 +1,77 @@
...
+        throw new Error\InvalidRequest($msg, null);
+    }
+
+    /**
+     * @param array|string $_id
+     * @param array|string|null $_opts
+     *
+     * @throws \Stripe\Error\InvalidRequest
+     */
+    public static function all($params = null, $opts = null)
```
remi-stripe: Weirdly this was not verified in `Person.php`. Should it?

remi-stripe:
@ob-stripe I retried the tests numerous times to no avail. Looks like the hhvm builds are failing but I don't think it's related to my changes. Can you double check and if it's all good review and release?

ob-stripe:
> Looks like the hhvm builds are failing but I don't think it's related to my changes.

Yeah, this is unrelated. [HHVM 4.0](https://hhvm.com/blog/2019/02/11/hhvm-4.0.0.html) was released yesterday, and it's no longer compatible with PHP. We'll have to remove HHVM from the CI, or maybe pin it to the last 3.x version.

lib/SubscriptionScheduleRevision.php
ob-stripe: Not really: `SubscriptionScheduleRevision` doesn't have any instance methods, so this is never called. It's fine to leave it though.

lib/SubscriptionScheduleRevision.php
ob-stripe: I'm not convinced those methods that immediately raise an exception are very useful, but it's better to be consistent, so I'd say yes.

tests/Stripe/SubscriptionScheduleTest.php
```diff
@@ -0,0 +1,116 @@
...
+        $resource = SubscriptionSchedule::retrieve(self::TEST_RESOURCE_ID);
+        $this->expectsRequest(
+            'post',
+            '/v1/subscription_schedules/' . $resource->id . '/cancel',
+            []
+        );
+        $resource->cancel([]);
+        $this->assertInstanceOf("Stripe\\SubscriptionSchedule", $resource);
+    }
+
```
ob-stripe: Can you add a test for `release`?

PR #591, opened 2019-01-31
Fix phpdoc of nested resources

franmomu:
Hi there,

Phpdoc references to the `$id` as an array and it is a `string`. And also the `{_create,_retrieve,_update,_delete}NestedResource` methods expect a string.


brandur-stripe:
Thanks @franmomu!

brandur-stripe:
Released as 6.29.2.

PR #592, opened 2019-01-31
Fix some more phpdoc

franmomu:
Sorry I forgot some fixes 🤦‍♂️, I also removed `$request_history` because it was duplicated. There is no need to make a release if you don't want to.


brandur-stripe:
No worries (and thanks)!

> Sorry I forgot some fixes 🤦‍♂️, I also removed `$request_history` because it was duplicated. There is no need to make a release if you don't want to.

Our releases our pretty easy to do, so it's no problem.

brandur-stripe:
Released as 6.29.3.

PR #599, opened 2019-02-16
Fix PHPDocs in SubscriptionSchedule and SubscriptionScheduleRevision

marotpam:
The PHPDocs in `SubscriptionSchedule` and `SubscriptionScheduleRevision` were incorrectly setting their classes to `Subscription`. There was also a typo in the property `chedule` inside of `SubscriptionScheduleRevision`, which would've caused some issues when integrating with the API


ob-stripe:
Released as 6.30.1.

PR #601, opened 2019-02-25
Adjusted phpdocs to match API docs as of 2019-02-25

nickdnk:
Hello guys

Back with some phpdoc updates. This is what I found this time around. Mostly just new or removed props, but we also have a somewhat critical typo in `application_fee_amount ` on `PaymentIntent` as well as some new event constants.

`Invoice` seems to have both `date` and `created` on the docs, but only `created` is described, and since I assume they do the same I removed `date`.

Let me know if this looks good.


lib/BankAccount.php
```diff
@@ -14,7 +14,6 @@
  * @property string $country
  * @property string $currency
  * @property string $customer
- * @property bool $default_for_currency
```
remi-stripe: This one should stay. The class is shared between bank accounts on Customers and bank accounts on Accounts and that property exists for the latter.

lib/Card.php
```diff
@@ -22,7 +22,6 @@
  * @property string $currency
  * @property string $customer
  * @property string $cvc_check
- * @property bool $default_for_currency
```
remi-stripe: same as above

lib/Event.php
```diff
@@ -1,5 +1,6 @@
 <?php
 
+
```
remi-stripe: remove extra line

lib/Event.php
```diff
@@ -27,121 +28,121 @@ class Event extends ApiResource
...
+    const CHARGE_REFUNDED                          = 'charge.refunded';
+    const CHARGE_SUCCEEDED                         = 'charge.succeeded';
+    const CHARGE_UPDATED                           = 'charge.updated';
+    const CHARGE_DISPUTE_CLOSED                    = 'charge.dispute.closed';
+    const CHARGE_DISPUTE_CREATED                   = 'charge.dispute.created';
+    const CHARGE_DISPUTE_FUNDS_REINSTATED          = 'charge.dispute.funds_reinstated';
+    const CHARGE_DISPUTE_FUNDS_WITHDRAWN           = 'charge.dispute.funds_withdrawn';
+    const CHARGE_DISPUTE_UPDATED                   = 'charge.dispute.updated';
+    const CHARGE_REFUND_UPDATED                    = 'charge.refund.updated';
+    const CHECKOUT_BETA_SESSION_SUCCEEDED          = 'checkout_beta.session_succeeded';
```
remi-stripe: Remove that one if possible please, it's a beta event that will go away.

lib/Event.php
```diff
@@ -27,121 +28,121 @@ class Event extends ApiResource
...
+    const CUSTOMER_SOURCE_EXPIRING                 = 'customer.source.expiring';
+    const CUSTOMER_SOURCE_UPDATED                  = 'customer.source.updated';
+    const CUSTOMER_SUBSCRIPTION_CREATED            = 'customer.subscription.created';
+    const CUSTOMER_SUBSCRIPTION_DELETED            = 'customer.subscription.deleted';
+    const CUSTOMER_SUBSCRIPTION_TRIAL_WILL_END     = 'customer.subscription.trial_will_end';
+    const CUSTOMER_SUBSCRIPTION_UPDATED            = 'customer.subscription.updated';
+    const FILE_CREATED                             = 'file.created';
+    const INVOICE_CREATED                          = 'invoice.created';
+    const INVOICE_DELETED                          = 'invoice.deleted';
+    const INVOICE_FINALIZED                        = 'invoice_finalized';
```
remi-stripe: should be `invoice.finalized`

lib/Event.php
```diff
@@ -27,121 +28,121 @@ class Event extends ApiResource
...
+    const ISSUING_AUTHORIZATION_CREATED            = 'issuing_authorization.created';
+    const ISSUING_AUTHORIZATION_UPDATED            = 'issuing_authorization.updated';
+    const ISSUING_CARD_CREATED                     = 'issuing_card.created';
+    const ISSUING_CARD_UPDATED                     = 'issuing_card.updated';
+    const ISSUING_CARDHOLDER_CREATED               = 'issuing_cardholder.created';
+    const ISSUING_CARDHOLDER_UPDATED               = 'issuing_cardholder.updated';
+    const ISSUING_DISPUTE_CREATED                  = 'issuing_dispute.created';
+    const ISSUING_DISPUTE_UPDATED                  = 'issuing_dispute.updated';
+    const ISSUING_SETTLEMENT_CREATED               = 'issuing_settlement.created';
+    const ISSUING_SETTLEMENT_UPDATED               = 'issuing_settlement.updated';
```
remi-stripe: Please remove those two for now, they should not be public (fixing on our end too).

lib/Event.php
```diff
@@ -27,121 +28,121 @@ class Event extends ApiResource
...
-    const SOURCE_REFUND_ATTRIBUTES_REQUIRED    = 'source.refund_attributes_required';
-    const SOURCE_TRANSACTION_CREATED           = 'source.transaction.created';
-    const SOURCE_TRANSACTION_UPDATED           = 'source.transaction.updated';
-    const SUBSCRIPTION_SCHEDULE_ABORTED        = 'subscription_schedule.aborted';
-    const SUBSCRIPTION_SCHEDULE_CANCELED       = 'subscription_schedule.canceled';
-    const SUBSCRIPTION_SCHEDULE_COMPLETED      = 'subscription_schedule.completed';
-    const SUBSCRIPTION_SCHEDULE_CREATED        = 'subscription_schedule.created';
-    const SUBSCRIPTION_SCHEDULE_EXPIRING       = 'subscription_schedule.expiring';
-    const SUBSCRIPTION_SCHEDULE_RELEASED       = 'subscription_schedule.released';
-    const SUBSCRIPTION_SCHEDULE_UPDATED        = 'subscription_schedule.updated';
```
remi-stripe: Please don't remove those. It'd be a breaking change. They will be released in the future.

lib/Person.php
```diff
@@ -14,7 +14,6 @@
  * @property mixed $address_kana
  * @property mixed $address_kanji
  * @property int $created
- * @property bool $deleted
```
remi-stripe: please leave this one. A Person can be deleted and the returned resource would have `deleted: true`

nickdnk:
@Remi I think I got all of them. I just noticed the problem with removing event constants as I received your reply. I forgot I was not in phpdoc-land for a second :)

PR #615, opened 2019-03-22
[WIP] Static request methods

ob-stripe:

This is a terrible, terrible hack that uses the [`__call()` and `__callStatic()` magic methods](https://secure.php.net/manual/en/language.oop5.overloading.php#object.call).

Pros:
- it does the job

Cons:
- no PHPDoc on either the static or the instance method, since the "real" method name must not exist in order for this trick to work
- slightly different behavior when calling a method that doesn't exist (I'm manually triggering a similar error, but it's not quite the same: `E_USER_ERROR` instead of `E_ERROR`, no stacktrace, maybe other differences)?
- probably other things

### DIFF
--- a/lib/ApiOperations/Delete.php
+++ b/lib/ApiOperations/Delete.php
@@ -10,12 +10,30 @@ namespace Stripe\ApiOperations;
 trait Delete
 {
     /**
+     * @param string $id The ID of the resource to update.
+     * @param array|null $params
+     * @param array|string|null $opts
+     *
+     * @return \Stripe\ApiResource The updated resource.
+     */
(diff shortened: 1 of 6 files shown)


brandur-stripe:
Ah man — kudos on getting this working, haha.

I think we should discuss this a little more (with Remi as well). IMO, the trade off here is a little too significant (all the reasons you list above plus the fact that it'll probably mess up IDEs and their ability to autocomplete as well). I think I'd rather see us just naming the static method something slightly different (e.g. "destroy") so that we can get a similar result, but with fewer hacks.

ob-stripe:
Closing this PR as we've decided not to go forward with this.

PR #618, opened 2019-03-25
Account::allExternalAccounts() returns a list

adri:
This actually returns a list of bank accounts or cards.


ob-stripe:
Hi @adri, thanks for the PR. The correct return type is actually `\Stripe\Collection`, and the list with the bank accounts or cards is in the collection's `data` attribute (but unfortunately I don't think we can represent that in PHPDoc).

ob-stripe:
Closing in favor of #619.

PR #621, opened 2019-03-30
Add support for CreditNote

remi-stripe:
cc @stripe/api-libraries 
cc @nickdnk because I finally remembered to add the right constants this time :)


mickjermsurawong-stripe:
LGTM

PR #623, opened 2019-04-02
Only use HTTP/2 with curl >= 7.60.0

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

When using HTTP/2, Stripe's API servers are configured to send a [GOAWAY](https://http2.github.io/http2-spec/#GOAWAY) frame after 1000 requests. curl versions older than 7.60.0 had a bug where curl would ignore the GOAWAY frame (https://github.com/curl/curl/issues/2416).

This PR changes the behavior of `CurlClient` so that we only set `CURLOPT_HTTP_VERSION` to `CURL_HTTP_VERSION_2TLS` if the version of curl is >= 7.60.0 unless the user explicitly overrides this, either by setting `CURLOPT_HTTP_VERSION` themselves (this was already the case prior to this PR) or by calling `setEnableHttp2(true)` on the `CurlClient` instance.

### DIFF
--- a/lib/HttpClient/CurlClient.php
+++ b/lib/HttpClient/CurlClient.php
@@ -44,6 +44,8 @@ class CurlClient implements ClientInterface
 
     protected $enablePersistentConnections = null;
 
+    protected $enableHttp2 = null;
+
     protected $curlHandle = null;
 
     /**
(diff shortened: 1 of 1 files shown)


ob-stripe:
Thanks Brandur!

ob-stripe:
Released as 6.31.3.

PR #634, opened 2019-04-18
[WIP] Static request methods

ob-stripe:
WIP PR for adding static methods for all API requests.

So far:
- `delete()` instance method is renamed to `selfDelete()`
- added `delete()` static method


remi-stripe:
I think this is the only viable approach, but this will change most integrations, because if we do this, we will do the same for all custom methods right?

Part of me wants to do a "real major" that also removes a lot of the legacy methods from the library (like `Customer::updateSubscription()` using the singular `/v1/customers/:id/subscription` endpoint).

But maybe it makes the major even more of a hurdle and we should instead limit the changes to just delete/custom methods.

Will wait to see what @brandur-stripe thinks (and if any developer using stripe-php has opinions please let us know! :)

ob-stripe:
Closing this as it's become stale and we're unlikely to go down this road (rather we'll probably go with the services approach).

PR #635, opened 2019-04-19
[WIP] Add deprecation notices

ob-stripe:
r? @remi-stripe

### DIFF
--- a/lib/Charge.php
+++ b/lib/Charge.php
@@ -118,9 +118,19 @@ class Charge extends ApiResource
      * @param array|string|null $options
      *
      * @return Charge The refunded charge.
+     *
+     * @deprecated This method is deprecated and planned for removal. Prefer using
+     *             \Stripe\Refund::create() instead.
+     * @see \Stripe\Refund::create()
      */
+    // TODO: remove in next major version
     public function refund($params = null, $options = null)
     {
+        trigger_error(
+            "This method is deprecated and planned for removal. Prefer using \Stripe\Refund::create() instead",
+            E_USER_DEPRECATED
+        );
+
         $url = $this->instanceUrl() . '/refund';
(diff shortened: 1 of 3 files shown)


lib/Charge.php
```diff
@@ -118,9 +118,19 @@ class Charge extends ApiResource
      * @param array|string|null $options
      *
      * @return Charge The refunded charge.
+     *
+     * @deprecated This method is deprecated and planned for removal. Prefer using
+     *             \Stripe\Refund::create() instead.
+     * @see \Stripe\Refund::create()
      */
+    // TODO: remove in next major version
```
remi-stripe: We could technically remove the TODO since we have @deprecated

lib/Charge.php
```diff
@@ -118,9 +118,19 @@ class Charge extends ApiResource
...
     public function refund($params = null, $options = null)
     {
+        trigger_error(
+            "This method is deprecated and planned for removal. Prefer using \Stripe\Refund::create() instead",
+            E_USER_WARNING
+        );
```
remi-stripe: This I'm quite meh about. While devs should have warning disabled, I am not convinced everyone would and this is quite dangerous to do to users on a minor rev.

ob-stripe: I've just found out that `E_USER_DEPRECATED` exists since PHP 5.3, so I've updated the PR to use it instead of `E_USER_WARNING`.

remi-stripe: I'm still quite bearish here. I don't think it helps to surface this client-side ultimately. Like sure people will see it and maybe decide to do something with it, but the risk of having this appear to a real customer seems not worth it

ob-stripe:
Going to close this one as well, we'll just remove the methods in stripe-php v7 and document the breaking changes in a migration guide.

PR #639, opened 2019-04-24
Fixed phpdoc/constants for Customer, Invoice and TaxID

nickdnk:
In relation to customer tax ID logic, there have been a few changes to the customer object.

Updates reflected in this PR.

I removed the deprecated properties. Not sure if you want them to remain. I figure it's good practice to remove them as this draws attention to them having been deprecated without breaking anything.

### DIFF
--- a/lib/Customer.php
+++ b/lib/Customer.php
@@ -8,6 +8,7 @@ namespace Stripe;
  * @property string $id
  * @property string $object
  * @property int $account_balance
+ * @property mixed $address
  * @property string $created
  * @property string $currency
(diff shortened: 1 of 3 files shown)


nickdnk:
@remi-stripe

*cracks knuckles* - watch this!

nickdnk:
Made a few adjustments. Namely to `custom_fields` that were annotated as `mixed` before - it is `array - I missed that. And then I made a similar mistake in this PR on `customer_address` - it was a `string`, now `mixed`.

Should be good now.

PR #640, opened 2019-04-25
Added phpdocs for tax_rates, fixed Coupon and Subscription

nickdnk:
Hey

I don't know if I missed these yesterday or if they were just added, but I've added `tax_rates` properties in all places I could find them.

Also found an int/float error on `Coupon` and missing `default_payment_method` on Subscription.

I've removed `tax_percent` from `Invoice` as it seems it's deprecated. Not explicitly though.

### DIFF
--- a/lib/Invoice.php
+++ b/lib/Invoice.php
@@ -31,6 +31,7 @@ namespace Stripe;
  * @property array $customer_tax_ids
  * @property string $default_payment_method
  * @property string $default_source
+ * @property array $default_tax_rates
  * @property string $description
  * @property Discount $discount
  * @property int $due_date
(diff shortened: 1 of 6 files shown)


lib/Invoice.php
```diff
@@ -58,9 +58,9 @@
  * @property int $subscription_proration_date
  * @property int $subtotal
  * @property int $tax
- * @property float $tax_percent
  * @property mixed $threshold_reason
```
remi-stripe: you're missing `default_tax_rates` too: https://stripe.com/docs/api/invoices/object#invoice_object-default_tax_rates

nickdnk: Damn. This is not the golden standard I've grown to expect from myself. Fixed.

PR #642, opened 2019-04-26
Fix idempotency key check when using automatic retries

### LINKED ISSUE #637: Using network retries causes client defined Idempotency Key to be over written
When setting the `idempotency_key` option on a request, it gets overwritten if you are using network retries.  The issue is being caused by a check for an existing `Idempotency-Key` in the headers array by key, however, the `$headers` array was already converted to a non-keyed array in `ApiRequestor::_requestRaw()`.

Ref: https://github.com/stripe/stripe-php/blob/06c612bca4fcad33efa3cd54bc47942454eca031/lib/HttpClient/CurlClient.php#L208

Example:
```
\Stripe\Stripe::setMaxNetworkRetries( 2 );
\Stripe\Charge::create( $charge_parameters, [
  'api_key' => 'abcdefg',
  'idempotency_key' => '1234-5678-abcd-efg',
]);
```

The request getting sent to the API are getting a newly generated idempotency_key instead of the one manually set.

ob-stripe:
cc @stripe/api-libraries 

When using automatic retries, `CurlClient` overwrites the existing idempotency key because `$headers` is a list of strings and not a map of key/values. This PR fixes it by adding a `hasHeader` helper method to check if a list of headers already has a specific header.

Fixes #637.

No tests because `CurlClient` is ~impossible to test :(

### DIFF
--- a/lib/HttpClient/CurlClient.php
+++ b/lib/HttpClient/CurlClient.php
@@ -205,7 +205,7 @@ class CurlClient implements ClientInterface
         // It is only safe to retry network failures on POST requests if we
         // add an Idempotency-Key header
         if (($method == 'post') && (Stripe::$maxNetworkRetries > 0)) {
-            if (!isset($headers['Idempotency-Key'])) {
+            if (!$this->hasHeader($headers, "Idempotency-Key")) {
                 array_push($headers, 'Idempotency-Key: ' . $this->randomGenerator->uuid());
             }
         }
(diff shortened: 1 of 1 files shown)


lib/HttpClient/CurlClient.php
```diff
@@ -439,4 +439,22 @@ private function canSafelyUseHttp2()
...
+    private function hasHeader($headers, $name)
+    {
+        foreach ($headers as $header) {
+            if (strncasecmp($header, "{$name}: ", strlen($name) + 2) === 0) {
+                return true;
+            }
+        }
+
+        return false;
+    }
```
remi-stripe: minor but isn't this quite slow/overhead? Should we maybe store "this information" (the fact that we set an idempotency key) elsewhere?

ob-stripe: Not really possible without modifying `ClientInterface` unfortunately.

PR #643, opened 2019-04-26
Use \Stripe\<class>::class constant instead of strings

nickdnk:
I figured since we've dropped 5.5 in v7 we might as well do this. Unless of course there's a reason not to that I'm not aware of.


nickdnk:
Visually we could go with:

```
\Stripe\Account::OBJECT_NAME              => \Stripe\Account::class,
\Stripe\AccountLink::OBJECT_NAME          => \Stripe\AccountLink::class,
\Stripe\AlipayAccount::OBJECT_NAME        => \Stripe\AlipayAccount::class,
\Stripe\ApplePayDomain::OBJECT_NAME       => \Stripe\ApplePayDomain::class,
\Stripe\ApplicationFee::OBJECT_NAME       => \Stripe\ApplicationFee::class,
\Stripe\Balance::OBJECT_NAME              => \Stripe\Balance::class,
```

Like all other constant lists, if you prefer. Not that it matters, but since we're changing this anyway.

brandur-stripe:
Looks way cleaner to me — thanks for the follow up @nickdnk!

I'll leave merge decision up to Olivier.

ptal @ob-stripe

ob-stripe:
Nice! Yeah, this seems like a strict improvement.

@nickdnk There are a few other places where we use strings for class names, mind updating these too? If not I can take as a followup.

nickdnk:
@ob-stripe Could you point me in the direction of those? I can't really find anything of this kind elsewhere.

nickdnk:
I just alphabetised the list based on the constant names instead of the `OBJECT_NAME` value, now that there are no string are in the list.

Also added minor indent on `Person`.

ob-stripe:
@nickdnk In the library itself, I think the only other place where we use strings for class names is in the `class_alias()` call in `lib/FileUpload.php`.

There are a lot more class name strings in tests though, mostly in `assertInstanceOf()` calls. Can you take fixing those as well? If not I can take. Hopefully it's just a few search and replaces!

nickdnk:
I updated the constants I could find in tests, but I cannot find `lib/FileUpload.php`?

ob-stripe:
@nickdnk Here: https://github.com/stripe/stripe-php/blob/60335d5dad000fb4b08aa7176b832a59d13d0420/lib/FileUpload.php#L6

nickdnk:
@ob-stripe
This file doesn't exist on this branch. `lib/File` exists but doesn't use class_alias.

ob-stripe:
@nickdnk 🤦‍♂ Sorry, I forgot I removed it for the next major version. Don't mind me. Thanks for updating everything else!

nickdnk:
@ob-stripe No problem.

Let me know when you're satisfied and I'll squash to one commit and reword it.

nickdnk:
Squashed to single commit.

PR #644, opened 2019-04-30
Update return type to `static` to improve static analysis

virgofx:
While there is no official PHPDoc standard ... Pretty much all static analyzers (phpstan, etc.) and leading IDEs (e.g. PHPStorm) all support the `@return static`. Making this change will improve static analysis and prevent unnecessary docblock declarations.

### DIFF
--- a/lib/ApiOperations/Retrieve.php
+++ b/lib/ApiOperations/Retrieve.php
@@ -15,7 +15,7 @@ trait Retrieve
      *     or an options array containing an `id` key.
      * @param array|string|null $opts
      *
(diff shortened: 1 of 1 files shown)


ob-stripe:
Hi @virgofx, thanks for the contribution! This looks peaceful to me.

virgofx:
:+1: Very minor and there were a few other areas in PHP docblock in the code that already use `@return static` so it just keeps it inline with those.

ob-stripe:
Released in 6.34.3.

PR #645, opened 2019-04-30
Fix constant for payment_intent.payment_failed

### LINKED ISSUE #609: `payment_intent.failed` or `payment_intent.payment_failed` event type
The `\Stripe\Event::PAYMENT_INTENT_FAILED` constant is equal to `'payment_intent.failed'`, but the actual webhook event type I'm getting is `'payment_intent.payment_failed'` as described in [the docs](https://stripe.com/docs/payments/payment-intents/verifying-status#handling-webhook-events).
I did not find information about it being renamed in the changelogs.
Is the event type still subject to changes as PaymentIntent is in beta or is it just a mistake that the old name was added in this library's code instead of the new one?

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Fixes the constant for `payment_intent.payment_failed`. Technically a breaking change because I also renamed the constant, but given that the existing constant has the wrong value, I don't think anyone is using it.

Fixes #609.


ob-stripe:
Released as 6.34.3.

PR #649, opened 2019-05-04
Add support for Customer Balance Transaction resource and APIs

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/lib/Customer.php
+++ b/lib/Customer.php
@@ -7,8 +7,8 @@ namespace Stripe;
  *
  * @property string $id
  * @property string $object
- * @property int $account_balance
  * @property mixed $address
+ * @property int $balance
  * @property string $created
  * @property string $currency
  * @property string $default_source
@@ -62,6 +62,7 @@ class Customer extends ApiResource
         return $savedNestedResources;
     }
 
+    const PATH_BALANCE_TRANSACTIONS = '/balance_transactions';
     const PATH_SOURCES = '/sources';
     const PATH_TAX_IDS = '/tax_ids';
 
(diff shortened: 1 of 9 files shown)


lib/CustomerBalanceTransaction.php
```diff
@@ -0,0 +1,75 @@
...
+                null
+            );
+        }
+        $id = Util\Util::utf8($id);
+        $customer = Util\Util::utf8($customer);
+
+        $base = Customer::classUrl();
+        $customerExtn = urlencode($customer);
+        $extn = urlencode($id);
+        return "$base/$customerExtn/customer_balance_transactions/$extn";
```
remi-stripe: this all method bothers me every time I copy it. I really think there needs to be a cleaner helper that takes the name of the parent property or something and does the rest automatically.

ob-stripe: Agreed. In this case I think we don't even need the method, since there are no instance methods.

lib/CustomerBalanceTransaction.php
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * @param array|string $_id
+     * @param array|string|null $_opts
+     *
+     * @throws \Stripe\Error\InvalidRequest
+     */
+    public static function retrieve($_id, $_opts = null)
+    {
+        $msg = "Customer Balance Transactions cannot be accessed without a customer ID. " .
+               "Retrieve a Customer Balance Transaction using Customer::retrieveCustomerBalanceTransaction('cbtxn_123') instead.";
```
ob-stripe:
```suggestion
               "Retrieve a Customer Balance Transaction using Customer::retrieveCustomerBalanceTransaction('cus_123', 'cbtxn_123') instead.";
```

lib/CustomerBalanceTransaction.php
remi-stripe: Should I remove it?

ob-stripe: 🤷‍♂ Let's keep it for now, it doesn't hurt to keep it.

lib/Customer.php
```diff
@@ -62,6 +62,7 @@ public static function getSavedNestedResources()
         return $savedNestedResources;
     }
 
+    const PATH_CUSTOMER_BALANCE_TRANSACTIONS = '/balance_transactions';
```
ob-stripe: Can you rename this to `PATH_BALANCE_TRANSACTIONS`?

lib/Customer.php
```diff
@@ -265,4 +266,55 @@ public static function allTaxIds($id, $params = null, $opts = null)
     {
         return self::_allNestedResources($id, static::PATH_TAX_IDS, $params, $opts);
     }
+
+    /**
+     * @param string|null $id The ID of the customer on which to create the customer balance transaction.
```
ob-stripe:
```suggestion
     * @param string|null $id The ID of the customer on which to create the balance transaction.
```

And same for the comments for the other methods.

remi-stripe:
@ob-stripe forgot to re-assign that one, PTAL

lib/Customer.php
```diff
@@ -265,4 +266,55 @@ public static function allTaxIds($id, $params = null, $opts = null)
...
+    }
+
+    /**
+     * @param string|null $id The ID of the customer on which to retrieve the customer balance transactions.
+     * @param array|null $params
+     * @param array|string|null $opts
+     *
+     * @return Collection The list of customer balance transactions.
+     */
+    public static function allCustomerBalanceTransactions($id, $params = null, $opts = null)
```
ob-stripe: Rename to `allBalanceTransactions`

remi-stripe: fixed, PTAL

ob-stripe:
Released as 6.38.0.

PR #651, opened 2019-05-09
Add support for the Capability resource and APIs

remi-stripe:
cc @stripe/api-libraries @rc-stripe


ob-stripe:
Released as 6.35.0.

PR #655, opened 2019-05-13
Add constants for payment intent statuses

marotpam:
Just noticed constants for payment intents status were also missing. I left out the ones that were deprecated in version [2019-02-11](https://stripe.com/docs/upgrades#2019-02-11) as I assume they are no longer needed


ob-stripe:
Released as 6.35.2.

PR #658, opened 2019-05-20
Use absolute value when checking timestamp tolerance

### LINKED ISSUE #657: Why is timestamp tolerance check in WebhookSignature one-sided?
PHP version: 7.2.18
stripe-php version: v6.35.0

If this is deliberate, then fair enough, can a maintainer close this issue and make sure it's documented.

Line 63 of lib/WebhookSignature.php:
```php
        // Check if timestamp is within tolerance
        if (($tolerance > 0) && ((time() - $timestamp) > $tolerance)) {
            throw new Error\SignatureVerification(
                "Timestamp outside the tolerance zone",
                $header,
                $payload
            );
        }
```

When time() is greater than supplied timestamp -
time() - 1558331514
timestamp - 1558331213

and default 300 second tolerance, that blows up because time() is 301 seconds greater than timestamp.

In light of the doc block for verifyHeader, namely
```
     * @param int $tolerance maximum difference allowed between the header's
     *  timestamp and the current time
```

Why, since tolerances tend to be two-sided in my experience, doesn't the timestamp check blow up when the values of time() and timestamp are reversed - ie, timestamp is more than 5 minutes ahead of time() ?

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Use absolute value when checking timestamp tolerance, since the local clock may be running either early or late compared to Stripe's clock.

Fixes #657.

Also, I've moved the signature check to happen before the event deserialization -- not much point deserializing the event if we're going to throw an exception afterwards.

### DIFF
--- a/lib/Webhook.php
+++ b/lib/Webhook.php
@@ -24,6 +24,8 @@ abstract class Webhook
      */
     public static function constructEvent($payload, $sigHeader, $secret, $tolerance = self::DEFAULT_TOLERANCE)
     {
+        WebhookSignature::verifyHeader($payload, $sigHeader, $secret, $tolerance);
+
         $data = json_decode($payload, true);
         $jsonError = json_last_error();
         if ($data === null && $jsonError !== JSON_ERROR_NONE) {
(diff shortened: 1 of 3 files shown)


ob-stripe:
Thanks Brandur!

Released as 6.35.1.

PR #659, opened 2019-05-20
Fix parameter type DocBlocks for various nested Account actions

virgofx:
Pretty straightforward. Detecting these discrepancies by running static analysis tools over application using this library.

cc: @ob-stripe


ob-stripe:
Released as 6.35.2.

PR #660, opened 2019-05-20
Phpdocs

nickdnk:
The usual minor edits here and there. Added some constants to BankAccount as well.

I'd suggest you merge in https://github.com/stripe/stripe-php/pull/655 at the same time as I skipped those constants because of that PR.

### DIFF
--- a/lib/TaxId.php
+++ b/lib/TaxId.php
@@ -12,7 +12,6 @@ namespace Stripe;
  * @property string $country
  * @property int $created
  * @property string $customer
- * @property bool $deleted
  * @property bool $livemode
  * @property string $type
  * @property string $value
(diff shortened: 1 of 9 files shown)


lib/TaxId.php
```diff
@@ -12,7 +12,6 @@
  * @property string $country
  * @property int $created
  * @property string $customer
- * @property bool $deleted
```
ob-stripe: Technically that one is possible -- all deleted resources are represented as `{"id": "foo_123", "deleted": true}` in the API. That said it doesn't look like we document this attribute on other deletable resources, and there is a documented helper for it anyway (https://github.com/stripe/stripe-php/blob/master/lib/StripeObject.php#L519) so let's be consistent and remove it here as well.

ob-stripe:
Released as 6.35.2.

PR #662, opened 2019-05-22
Added BalanceTransaction type constants

nickdnk:
Pretty self-explanatory this one.

Cheers.

### DIFF
--- a/lib/BalanceTransaction.php
+++ b/lib/BalanceTransaction.php
@@ -31,6 +31,40 @@ class BalanceTransaction extends ApiResource
     use ApiOperations\Retrieve;
 
     /**
+     * Possible string representations of the type of balance transaction.
+     * @link https://stripe.com/docs/api/balance/balance_transaction#balance_transaction_object-type
+     */
+    const TYPE_ADJUSTMENT                    = 'adjustment';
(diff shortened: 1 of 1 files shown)


nickdnk:
The only thing that's a bit odd here is that on https://stripe.com/docs/api/balance/balance_transaction#balance_transaction_object-type, the `tax_fee` is called simply `tax`.

Should we include a type `tax` in this list to be able to use these constants inside fee_details as well?

ob-stripe:
Released as 6.36.0.

nickdnk:
Hehe. That was fast. Did you get a chance to read my comment above?

ob-stripe:
Oops, sorry, I missed it. Technically this is a different attribute so I think it would be more appropriate to define a second set of constants `FEE_DETAILS_TYPE_...`. However this attribute is not defined as an enum in the OpenAPI spec (despite what the description implies) so I'd rather avoid adding constants for now. It probably _should_ be an enum but I think the current backend implementation makes it difficult. I'll try to clarify with the owning team.

nickdnk:
@ob-stripe I also think it would be kind of silly to add three constants where two of them already exist on the same object. Much like the overlapping review_reasons on the Review object.

PR #667, opened 2019-06-07
Ensure correct type for `default_source`

virgofx:
The `default_source` for a customer is optional and returns `null` when not set.

/cc @ob-stripe

### DIFF
--- a/lib/Customer.php
+++ b/lib/Customer.php
@@ -11,7 +11,7 @@ namespace Stripe;
  * @property mixed $address
  * @property string $created
  * @property string $currency
- * @property string $default_source
+ * @property string|null $default_source
  * @property bool $delinquent
  * @property string $description
  * @property Discount $discount


nickdnk:
Hello

None of the phpdoc properties are listed as nullable, even though many of them can be null. This change would be a very inconsistent.

ob-stripe:
Ah, good point, I didn't realize that.

We'll close this PR for now for consistency, and hopefully we will soon be able to autogenerate the PHPDoc comments with correct types and descriptions.

Thanks for the contribution anyway @virgofx!

virgofx:
Any info on when updated docblock properties will be implemented?

ob-stripe:
I can't offer a firm ETA, but we're actively working on autogenerating parts of our client libraries (including documentation). We haven't started work on stripe-php yet, but we will get to it sometimes during the second half of this year.

PR #671, opened 2019-06-14
Add PHPDoc for submit_type to Checkout\Session

cjavilla-stripe:
?r @ob-stripe 
cc @stripe/api-libraries


lib/Checkout/Session.php
```diff
@@ -16,6 +16,7 @@
  * @property string $payment_intent
  * @property string[] $payment_method_types
  * @property string $subscription
+ * @property string $submit_type
```
ob-stripe: Should be before `$subscription`.

cjavilla-stripe: Done.

?r @ob-stripe

PR #676, opened 2019-06-26
Prevent variable from being parsed

yoavf:
The error message in `\Stripe\CustomerBalanceTransaction::update` contains a variable that will be parsed since the string is enclosed in double quotes and the `$` is not escaped


ob-stripe:
Fixed in 6.39.2.

PR #679, opened 2019-07-02
Added "requires_source" payment status constant

Skullbock:


remi-stripe:
@Skullbock the status `requires_source` is only available on older version of the API. This value was deprecated in the API version [2019-02-11](https://stripe.com/docs/upgrades#2019-02-11) and replaced by `requires_payment_method`.

We try not to add deprecated constants to the library to avoid making things more confusing for new integration.

I would recommend that you upgrade your API version in your integration to ensure you're always on the most recent shape of the API. You can even pin specific API requests to an API version as documented here: https://stripe.com/docs/api/versioning

I'm going to close this PR for now, I hope it's okay!

Skullbock:
Hey Remi! Thanks for the feedback, i realized my mistake when you wrote it, sorry for the trouble.
Thanks!

remi-stripe:
@Skullbock No trouble at all, API versioning can be confusing with stripe-php since we support all API versions automatically!

PR #686, opened 2019-08-02
Use array_column to flatten params

ob-stripe:
r? @remi-stripe @brandur-stripe 
cc @stripe/api-libraries 

Just a small improvement that depended on PHP > 5.4.


brandur-stripe:
Yep, seems like a strict improvement to me.

LGTM.

PR #693, opened 2019-08-09
Standardize UsageRecord methods

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Replace `UsageRecord::create($params)` (where `$params` must include a `subscription_item` key) with `SubscriptionItem::createUsageRecord($id, $params)`.


ob-stripe:
> Approving but technically it'd be better to add the new way to master and update the docs and such though

Good point. I've added the method to the current branch and released it as 6.43.0. Will close this PR and amend #692 to remove `UsageRecord::create`.

PR #698, opened 2019-08-20
add  property and setter

ThomasDupont:
Issue:

For unittests, we need to set and get an id on Stripe Object. But the id is protected.

Test example:
```php
$source = new Stripe\Source();
$source->id = 'src_test';
$this->stripeClientMock->expects($this->at(1))->method('attachSource')->withArgs('cus_test', 'src_test')->willReturn(true);
// some other mocks (customer, DB ...)
$stripeSource = (new SourceManager())->updateSource('cus_test', 'src_test');
$this->assertEquals($stripeSource->id, 'src_test');

```

This update will not modify the previus behavior of the SDK

### DIFF
--- a/lib/StripeObject.php
+++ b/lib/StripeObject.php
@@ -9,6 +9,8 @@ namespace Stripe;
  */
 class StripeObject implements \ArrayAccess, \Countable, \JsonSerializable
 {
+    protected static $_permanentAttributes = null;
+
     protected $_opts;
     protected $_originalValues;
     protected $_values;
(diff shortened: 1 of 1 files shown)


ob-stripe:
Hey @ThomasDupont, thanks for the contribution!

I'm not necessarily opposed to pulling this in as it looks it maintains the immutability of the ID as long as you're not using the new `setPermanentAttributes` method, but from your example I don't see the obvious advantage over simply doing this:

```php
$source = \Stripe\Source::constructFrom(['id' => 'src_test']);
```

Basically, if you only need to set the ID once (at creation time) then I think the above is preferable.

If you need to update the ID on an existing instance of any `StripeObject`, then of course you can't use `constructFrom` (which is a static method that returns a new instance) and your PR would be a good escape hatch. I can't think of a case where that would be needed though, since IDs are immutable in the API.

Would merging this still be useful for you, or would `constructFrom` be sufficient for your testing use case?

ThomasDupont:
Hello @ob-stripe 

Your solution works. Maybe you could add your answer to the README for mocking.

Thanks for your support

Thomas

ob-stripe:
No problem. Closing this, but thanks anyway for the contribution!

PR #703, opened 2019-08-22
Use yield for autopagination

ob-stripe:
cc @stripe/api-libraries 

Replace the `AutoPagingIterator` class with a generator function. Fixes a 3+ years old [frustration](https://github.com/stripe/stripe-php/pull/224#issuecomment-181024997) of mine :)


lib/Collection.php
```diff
@@ -77,14 +77,30 @@ public function getIterator()
     }
 
     /**
-     * @return Util\AutoPagingIterator An iterator that can be used to iterate
-     *    across all objects across all pages. As page boundaries are
+     * @return \Generator|StripeObject[] A generator that can be used to
```
ob-stripe: Clunky, but this is apparently the best (or least bad) way of documenting generators: https://stackoverflow.com/a/36952114/5307473

brandur-stripe: Oh crazy. Well thanks for doing the research on that!

PR #705, opened 2019-08-22
Add ErrorObject to Stripe exceptions

ob-stripe:
cc @stripe/api-libraries 

Like https://github.com/stripe/stripe-ruby/pull/811, but for stripe-php.

### DIFF
--- a/lib/ErrorObject.php
+++ b/lib/ErrorObject.php
@@ -1,0 +1,66 @@
+<?php
+
+namespace Stripe;
+
+/**
+ * Class ErrorObject
+ *
+ * @property string $charge For card errors, the ID of the failed charge.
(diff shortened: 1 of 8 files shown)


lib/ErrorObject.php
```diff
@@ -0,0 +1,66 @@
...
+     *
+     * @param array $values
+     * @param null|string|array|Util\RequestOptions $opts
+     * @param boolean $partial Defaults to false.
+     */
+    public function refreshFrom($values, $opts, $partial = false)
+    {
+        // Unlike most other API resources, the API will omit attributes in
+        // error objects when they have a null value. We manually set default
+        // values here to facilitate generic error handling.
```
brandur-stripe: Good idea.

PR #708, opened 2019-08-23
Update error messages for nested resource operations

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Make the error messages more consistent, recommend the static methods everywhere.

### DIFF
--- a/lib/AlipayAccount.php
+++ b/lib/AlipayAccount.php
@@ -47,8 +47,9 @@ class AlipayAccount extends ApiResource
      */
     public static function retrieve($_id, $_opts = null)
     {
-        $msg = "Alipay accounts cannot be accessed without a customer ID. " .
-               "Retrieve an Alipay account using \$customer->sources->retrieve('alipay_account_id') instead.";
+        $msg = "Alipay accounts cannot be retrieved without a customer ID. " .
+               "Retrieve an Alipay account using `Customer::retrieveSource(" .
+               "'customer_id', 'alipay_account_id')`.";
         throw new Error\InvalidRequest($msg, null);
     }
 
@@ -64,8 +65,9 @@ class AlipayAccount extends ApiResource
      */
     public static function update($_id, $_params = null, $_options = null)
     {
-        $msg = "Alipay accounts cannot be accessed without a customer ID. " .
-               "Call save() on \$customer->sources->retrieve('alipay_account_id') instead.";
+        $msg = "Alipay accounts cannot be updated without a customer ID. " .
+               "Update an Alipay account using `Customer::updateSource(" .
+               "'customer_id', 'alipay_account_id', \$updateParams)`.";
         throw new Error\InvalidRequest($msg, null);
     }
 }
--- a/lib/BankAccount.php
+++ b/lib/BankAccount.php
@@ -71,9 +71,11 @@ class BankAccount extends ApiResource
      */
     public static function retrieve($_id, $_opts = null)
     {
-        $msg = "Bank accounts cannot be accessed without a customer ID or account ID. " .
-               "Retrieve a bank account using \$customer->sources->retrieve('bank_account_id') or " .
-               "\$account->external_accounts->retrieve('bank_account_id') instead.";
+        $msg = "Bank accounts cannot be retrieved without a customer ID or " .
+               "an account ID. Retrieve a bank account using " .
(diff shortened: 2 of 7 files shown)


lib/AlipayAccount.php
```diff
@@ -47,8 +47,9 @@ public function instanceUrl()
      */
     public static function retrieve($_id, $_opts = null)
     {
-        $msg = "Alipay accounts cannot be accessed without a customer ID. " .
-               "Retrieve an Alipay account using \$customer->sources->retrieve('alipay_account_id') instead.";
+        $msg = "Alipay accounts cannot be retrieved without a customer ID. " .
+               "Retrieve an Alipay account using `Customer::retrieveSource(" .
+               "'customer_id', 'alipay_account_id')`.";
```
remi-stripe: technically we should say "This is deprecated please go away"

lib/BankAccount.php
```diff
@@ -71,9 +71,11 @@ public function instanceUrl()
...
     public static function retrieve($_id, $_opts = null)
     {
-        $msg = "Bank accounts cannot be accessed without a customer ID or account ID. " .
-               "Retrieve a bank account using \$customer->sources->retrieve('bank_account_id') or " .
-               "\$account->external_accounts->retrieve('bank_account_id') instead.";
+        $msg = "Bank accounts cannot be retrieved without a customer ID or " .
+               "an account ID. Retrieve a bank account using " .
+               "`Customer::retrieveSource('customer_id', " .
+               "'bank_account_id')` or `Account::retrieveExternalAccount(" .
+               "'account_id', 'bank_account_id')`.";
```
remi-stripe: I know with have rules for a row max width but this makes it so hard to read :(

lib/AlipayAccount.php
```diff
@@ -64,8 +65,9 @@ public static function retrieve($_id, $_opts = null)
      */
     public static function update($_id, $_params = null, $_options = null)
     {
-        $msg = "Alipay accounts cannot be accessed without a customer ID. " .
-               "Call save() on \$customer->sources->retrieve('alipay_account_id') instead.";
+        $msg = "Alipay accounts cannot be updated without a customer ID. " .
+               "Update an Alipay account using `Customer::updateSource(" .
```
remi-stripe: Why doesn't it say `\Stripe\Customer` instead? Aren't most people using it that way?

lib/CustomerBalanceTransaction.php
```diff
@@ -67,8 +67,10 @@ public function instanceUrl()
      */
     public static function retrieve($_id, $_opts = null)
     {
-        $msg = "Customer Balance Transactions cannot be accessed without a customer ID. " .
-               "Retrieve a balance transaction using Customer::retrieveBalanceTransaction('cus_123', 'cbtxn_123') instead.";
+        $msg = "Customer Balance Transactions cannot be retrieved without a " .
+               "customer ID. Retrieve a Customer Balance Transaction using" .
```
remi-stripe: missing a space here

lib/CustomerBalanceTransaction.php
```diff
@@ -81,8 +83,10 @@ public static function retrieve($_id, $_opts = null)
      */
     public static function update($_id, $_params = null, $_options = null)
     {
-        $msg = "Customer Balance Transactions cannot be accessed without a customer ID. " .
-               "Update a balance transaction using Customer::updateBalanceTransaction('cus_123', 'cbtxn_123', \$params) instead.";
+        $msg = "Customer Balance Transactions cannot be updated without a " .
+               "customer ID. Update a Customer Balance Transaction using" .
```
remi-stripe: missing a space

lib/AlipayAccount.php
ob-stripe: 🤷‍♂  Depends on whether they have `using Stripe;` at the top of their code or not. I think it's clear enough from context that the class is namespaced under `\Stripe`.

lib/AlipayAccount.php
ob-stripe: but what if someone wants to retrieve an Alipay account from 5 years ago

lib/BankAccount.php
ob-stripe: Ah, sorry. We don't even enforce a max line length in stripe-php, I just try to keep everything under 80 columns out of habit.

Issue #605, opened 2019-02-28
phpdocs for exceptions


nickdnk:
Hello guys

As I'm sure most of you maintainers are aware of, I'm a big fan of phpdocs and use it extensively in my code. I am well aware that annotation changes nothing at runtime, but I find it very helpful in determining if I am missing a return type, not handling an exception, passing a wrong type and so on when I'm coding and reviewing my code.

Now, after you made the change to use traits (I think it was major version 5?), the phpdocs for thrown exceptions are all out of whack because of the shared methods used across (as far as I can tell) most requests.

For me, this results in _a lot_ of this:

```
} /** @noinspection PhpRedundantCatchClauseInspection */ catch (\Stripe\Error\Base $e) {

    // I have to include above notation to avoid warnings.

}
```

And if I don't include this comment, PHPStorm gives me a "redundant catch phrase" error, which (if I erroneously should accept this fix) would result in a working catch block being removed entirely. Not so great.

So, I tried to "start from the bottom" (of traits) to include all exceptions everywhere they are thrown. This of course results in *all* requests throwing almost all exceptions - which is simply incorrect; such as retrieving a customer then claims it can throw `\Stripe\Error\Card`. Not a good solution.

Would it be possible to somehow at the very least have `InvalidRequest`, `Card`, and `Base` added in cases where these errors can be thrown? So that we *have* to handle (or silence) `Base`, which I'm under the impression *any* endpoint can throw, and which would of course also catch any other unexpected exception (RateLimiting, Authentication etc.). If so, how could we do this? I would be happy to put in the work.

nickdnk:
I'm thinking that you *could* do something like this, which would also fix return types:

``` 
use ApiOperations\Retrieve {
    retrieve as protected __retrieve;
}
```

And then

```
/**
 * @param string     $id
 * @param null|array $opts
 *
 * @throws Error\InvalidRequest
 * @throws Error\Permission
 * @throws Error\RateLimit
 * @throws Error\Base
 *
 * @returns Customer
 */
public function retrieve($id, $opts = null) {

   $this->__retrieve($id, $opts);

}
```

This is a lot of work and maintenance, but would work without actually changing or breaking anything.

remi-stripe:
@nickdnk Thanks a lot for providing so much details, this will definitely help us investigate a good solution.

[...]

PR #709, opened 2019-08-23
[wip] Modernize exceptions

ob-stripe:
This is a large PR that is kind of hard to review -- sorry!

- Rename `\Stripe\Error` namespace to `\Stripe\Exception`
- Rename all exception classes to append an` Exception` suffix
- Remove the `Base` / `OAuthBase` classes. Exception classes now inherit directly from PHP's `\Exception`
- Add an `ExceptionInterface` interface implemented by all our exception classes
- Add an `ExceptionTrait` traited used by all our exception classes to provide all common methods formerly provided by `Base` \ `OAuthBase`.
    - We're no longer overriding the constructor, instead the trait provides a static `factory` method to create instances
- Add new exception classes `BadMethodCallException`, `InvalidArgumentException` and `UnexpectedValueException` that wrap the matching [SPL exceptions](https://www.php.net/manual/en/spl.exceptions.php) (the purpose of these is to add the `ExceptionInterface` interface)

All the above changes make the exceptions much more idiomatic and less surprising for PHP developers. Defining an `ExceptionInterface` implemented by all exceptions in the codebase (including standard SPL exceptions by defining wrappers) is standard practice for modern PHP codebases.

WIP, still some remaining work to do:
- [ ] add docstrings to all exceptions classes and methods
- [ ] do a pass over every `throw` to make sure we're throwing the best-fitting type
- [ ] decide if API exceptions should derive from `\RuntimeException` instead of `\Exception`


nickdnk:
I think this is a fine solution.

On your last point left unchecked: In my opinion it should definitely *not* inherit from `RuntimeException`, especially now when everything can be handled by simply catching the interface. No other PHP libraries I've worked with use `RuntimeException` for errors that are expected to happen and that you _should_ recover gracefully from. Doing this would remove all warnings about missing or redundant exception handling in PhpStorm, which renders the entire `@throws` PhpDoc part of this library useless. I don't believe that was the point of this change.

While this thread is a question about Java, its answers outline the reasoning behind checked/unchecked (runtime) exceptions and when and why they are useful (and when they are not): https://stackoverflow.com/questions/6115896/understanding-checked-vs-unchecked-exceptions-in-java.

Since we're interacting with an API far away and outside our control as users of the library, it makes very much sense to *force* us to always take into consideration that Stripe may return an error - even in cases we expect it not to. We should *always* recover gracefully from this. Using `RuntimeExceptions` for our parent class discourages this behaviour, as developers may simply forget to handle errors which break their program. If you get tired of handling this for Stripe, the `ExceptionInterface` can simply be added to ignored Exceptions in PHPStorm (I assume most serious PHP developers use this program). `RuntimeException` is ignored by default, so inheriting from this would disable the try/catch warnings out of the box and make it impossible to enable for Stripe alone: If we remove `RuntimeException` from the list of ignored exceptions we'll be forced to implement a lot of suppression by annotation or a lot more try/catch blocks. `\mysqli_sql_exception` for instance inherits from `RuntimeException` and can be thrown almost anywhere MySQLi is used, while Stripe is probably in most cases isolated to specific classes or files where error handling can easily be contained and handled properly by whoever implements it. I realise this depends a lot on how one's project is structured, but at least that's the case for me.

That's my take on it anyway. I know checked vs. unchecked is somewhat of a religion, but hey, I believe they are a good thing when used properly.

nickdnk:
> LGTM OB. A cleanup of epic proportions!
> 
> I might be missing something due to general PHP ignorance, but one question I had was around the UX of `ExceptionInterface`. I can see from you examples that you can just use duck typing to access information on it (e.g. `$e->getMessage()`), but doesn't having no properties of its own make things inconvenient for editors and for people to reason about?
> 
> Put another way, having an `ExceptionInterface` doesn't tell you anything about what you're allowed to with a caught exception. You'd need to go and look through the code to see what exceptions inherit it, and what common properties they might have between them that you could use in your rescue handler. Should `ExceptionInterface` define some common Stripe exception properties for usability purposes?

Perhaps this thread can make enlighten us: https://github.com/guzzle/guzzle/issues/2184 - we may end up creating a situation similar to the problems developers experience with this Guzzle interface.

I'd recommend that we have all exceptions extend from an abstract class that has some functionality common to all Stripe exceptions (such as `$->getStripeErrorCode()`) and if necessary this can also have abstract methods that all exceptions must implement if we want "interfacy" behaviour. Catching this exception would accomplish the same as catching the interface.

nickdnk:
> That was basically the way it was implemented before, all exceptions derived from the `\Stripe\Error\Base` abstract class.

But was that a problem? I don't think this implementation solves any of the problems we had with the `Base` error (except for its name). It solves other stuff, sure, but when it comes to the issue I raised with PhpDocs, the @throws interface solution is pretty similar to throwing a base class.

nickdnk:
> I think the interface is preferable because it lets us wrap SPL exceptions, which are now used for some client-side errors (previously we used `InvalidRequest` or `Api` but I think that was bad because it meant those exceptions could be thrown on either purely client-side errors or map to an API error, which are two very different things).
> 
> Right now the stuff common to all API errors is implemented by the `ExceptionTrait` trait. I could change that to be an abstract class named e.g. `StripeServerException` or something.

I think having a different inheritance tree on errors that are caused by programming errors is a good approach, such as throwing an `InvalidArgumentException` when providing something that cannot possibly be correct to the library - before even sending the request. Obviously this cannot have functionality that only applies to HTTP responses from the Stripe API.

nickdnk:
@ob-stripe I think renaming it to `UnknownApiErrorException` is a good idea. I like having the base class being abstract, and they're named too similarly otherwise, which can be confusing. Perhaps you could add to the documentation of that exception that when it's thrown it may indicate that you're using an outdated version of the library and you should check for updates to properly handle the error.

nickdnk:
@ob-stripe

You're welcome.

I'm also a little confused about the purpose of `ExceptionInterface` after this latest change. What does it actually accomplish at this point? Is it just a way to do a quick catch-all? I mean this can be accomplished by catching \Exception as well?

nickdnk:
Just to clairfy; https://github.com/stripe/stripe-php/issues/605 is not solved by these changes, as we'll still get the warnings that all the actual exceptions are "not thrown" (as phpdocs only mention the abstract base class). I don't think that can be solved without explicitly adding all possible API errors to all methods.

nickdnk:
> > I'm also a little confused about the purpose of ExceptionInterface after this latest change. What does it actually accomplish at this point? Is it just a way to do a quick catch-all? I mean this can be accomplished by catching \Exception as well?
> 
> I'm no PHP expert but I saw this pattern being employed by a lot of PHP codebases (https://www.google.com/search?q=exceptioninterface.php+site%3Agithub.com). My understanding is that it serves as a nice way to know whether an exception was thrown by a library or by your own code, as long as the library makes sure that _all_ exceptions it throws implement its interface.

That does makes sense. Let's stick with that.

> > Just to clairfy; #605 is not solved by these changes, as we'll still get the warnings that all the actual exceptions are "not thrown" (as phpdocs only mention the abstract base class). I don't think that can be solved without explicitly adding all possible API errors to all methods.
> 
> Yes, sorry, I should have said that it "partially" solves the issue. Previously I had annotated the methods with `ExceptionInterface`, which would also cause warnings for the non-API error exceptions. That's no longer the case. Besides, with the exception of `CardException`, all API methods can potentially throw any API error exception.

I like only being forced to handle errors that come from the Stripe API, so I think this solution is probably the best middle-ground, with the problem still being that if you catch something like `InvalidRequestException`, PHPStorm will tell you that it's never thrown, which is wrong and is only because we've deliberately only annotated the abstract base class and ignored all the "missing exception tag" warnings that show up in the library now.

PR #711, opened 2019-08-26
Add `deleted` property

baijunyao:


nickdnk:
@ob-stripe Wouldn't this need to be implemented either on all objects or none of them? `deleted` is not listed on the API documentation and seems to apply to most or all objects that have been deleted, and is not present as a property until actually deleted.

baijunyao:
Thanks, I will use the `isDeleted()`.  
By the way, set `deleted ` to `false` in the non-deleted object, Will this be better?

PR #713, opened 2019-08-27
Add typehint to `ApiResponse`

baijunyao:
https://github.com/stripe/stripe-php/blob/v6.43.0/lib/ApiRequestor.php#L124-L127

https://github.com/stripe/stripe-php/blob/v6.43.0/lib/HttpClient/CurlClient.php#L214


ob-stripe:
Released as 6.43.1.

PR #718, opened 2019-08-28
Retry requests on a 429 that's a lock timeout

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Similar to what we did in Ruby and Go.


ob-stripe:
Thanks!

Issue #721, opened 2019-08-29
LoggerInterface claims compatibility with PSR-3, breaks PHPStan analysis when used as such


martin-georgiev:
`LoggerInterface` defines the [return type of `error()`](https://github.com/stripe/stripe-php/blob/b325b5c123d2786eeac21e62ab594836300fb10a/lib/Util/LoggerInterface.php#L35) as `null`:
```php
    /**
     * Runtime errors that do not require immediate action but should typically
     * be logged and monitored.
     *
     * @param string $message
     * @param array $context
     * @return null
     */
    public function error($message, array $context = []);
```

It also [claims](https://github.com/stripe/stripe-php/blob/b325b5c123d2786eeac21e62ab594836300fb10a/lib/Util/LoggerInterface.php#L8) that the interface is compatible with PSR-3, implying that a project can benefit from custom logger implementation that satisfies both Stripe's and PSR-3's interface. The `null` return here is actually incompatible with PSR-3 as `error()` is [defined as non-returning](https://github.com/php-fig/fig-standards/blob/614870bb31d4ed488f9c34a59beba8233d1a1c8d/accepted/PSR-3-logger-interface.md#3-psrlogloggerinterface):
```php
    /**
     * Runtime errors that do not require immediate action but should typically
     * be logged and monitored.
     *
     * @param string $message
     * @param array $context
     * @return void
     */
    public function error($message, array $context = array());
```

This creates an issue in some cases where a custom implementation of the `LoggerInterface` is added to a project that benefits from static analysis. As an example level 3 of PHPStan analysis detects this situation and errors. Sample of the error:

``` 
------ -------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   Library/Stripe/Logger.php                                                                                                                                                 
------ -------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  19     Return type (void) of method Acme\Library\Stripe\Logger::error() should be compatible with return type (null) of method Stripe\Util\LoggerInterface::error()  
------ -------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
```

___
Affected versions of the library: all
Observed on PHP versions: 7.1, 7.2, 7.3

ob-stripe:
Fixed in 6.43.1 via #722. Thanks again @martin-georgiev! Much appreciated.

martin-georgiev:
You're welcome. Thank you in return for the timely merge and release, @ob-atripe.

PR #722, opened 2019-08-29
Make LoggerInterface::error compatible with its PSR-3 counterpart

martin-georgiev:
Addresses #721


ob-stripe:
Nice, thanks a lot for the report and the fix!

ob-stripe:
Released as 6.43.1.

PR #725, opened 2019-08-30
Couple of fixes for PHP 7.4

### LINKED ISSUE #723: Array Functions on Objects Soon to be Deprecated

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Fixes a couple of warnings for PHP 7.4.

The PHP 7.4 builds are still red, but that's because php-cs-fixer doesn't support 7.4 yet. The test suite is passing without warnings.

Fixes #723.

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -368,7 +368,7 @@ class ApiRequestor
             $hasFile
         );
 
-        if (array_key_exists('request-id', $rheaders)) {
+        if (isset($rheaders['request-id'])) {
             self::$requestTelemetry = new RequestTelemetry(
                 $rheaders['request-id'],
                 Util\Util::currentTimeMillis() - $requestStartMs
--- a/lib/EphemeralKey.php
+++ b/lib/EphemeralKey.php
@@ -35,7 +35,7 @@ class EphemeralKey extends ApiResource
      */
     public static function create($params = null, $opts = null)
     {
-        if (!$opts['stripe_version']) {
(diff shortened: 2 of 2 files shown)


ankitkbl:
array_key_exists(): Using array_key_exists() on objects is deprecated. Use isset() or property_exists() instead in /var/www/html/backend/vendor/stripe/stripe-php/lib/ApiRequestor.php on line 402

same error coming again

ankitkbl:
i'm using PHP 7.4.3 (cli)

ob-stripe:
@ankitkbl Make sure you're using stripe-php 7.0.0 or more recent. If you're still observing this problem after upgrading, please open a new issue on GitHub.

andrii-pukhalevych:
Can it be back ported to version 6?

remi-stripe:
@andrii-pukhalevych Unfortunately we don't backport fixes like this. We recommend upgrading to the latest instead. Otherwise you can fork the library and patch this locally.

PR #729, opened 2019-09-06
Fix usage of SignatureVerification instead of SignatureVerificationException

robertboloc:
Fixed leftover `SignatureVerification` docblock usage instead of `SignatureVerificationException` after the refactor in 4699c5d.


ob-stripe:
Thanks @robertboloc!

ob-stripe:
Released as 7.0.2.

PR #736, opened 2019-09-13
Make CaseInsensitiveArray countable and traversable

### LINKED ISSUE #735: Allow CaseInsensitiveArray to be traversed
Stripe's exceptions return a lot of extra information about the HTTP request, which is awesome when debugging problems.

The HTTP Headers are held in a `Stripe\Util\CaseInsensitiveArray` collection. This makes it easy to get individual headers, but there's no way to get all of them, either via a `->all()` method or by iterating:
```
foreach ($exception->getHttpHeaders() as $k => $v) {
   // This is never entered
}
```

When handling the exception I would like to record all headers, so getting them would be very handy.

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Make `CaseInsensitiveArray` countable and traversable.

I also fixed a bug where the keys of the initial array passed to the constructor were not correctly converted to lowercase.

Fixes #735.

### DIFF
--- a/lib/Util/CaseInsensitiveArray.php
+++ b/lib/Util/CaseInsensitiveArray.php
@@ -2,8 +2,6 @@
 
 namespace Stripe\Util;
 
-use ArrayAccess;
-
 /**
  * CaseInsensitiveArray is an array-like class that ignores case for keys.
  *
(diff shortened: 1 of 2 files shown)


ob-stripe:
Thanks Brandur!

Released as 7.1.0.

PR #738, opened 2019-09-17
Added missing setup intent constants to events

nickdnk:
Hey guys

Please merge in these event constants :)

Edit: It would be really nice if you could release this on 6.x, as I don't have the time to migrate to 7.x right now but I still would like the constants if possible.


nickdnk:
Then it's time to host my own package until then. Haha. Thanks.

PR #748, opened 2019-09-26
Add `null` type for the `latest_invoice` of `Subscription`

baijunyao:


ob-stripe:
Hi @baijunyao and thanks for the contribution.

Currently, none of the properties listed in PHPDoc blocks are marked as nullable, even though many of them are.

We will fix this in the future, but for consistency we would prefer to fix all of them at once. We're currently working on autogenerating parts of our client libraries code which should make this easy. We haven't started working on stripe-php just yet but we should get to it before the end of the year.

Thanks for your patience!

baijunyao:
OK, thank you.

PR #751, opened 2019-09-30
Add `total_count` property to phpdoc of `Collection`

baijunyao:


remi-stripe:
@baijunyao Thanks for opening a PR for this! Unfortunately, we consider `total_count` deprecated and have for a few years now. We return it for legacy reasons today but we don't want to make it visible in the libraries. We recommend that you don't rely on this parameter moving forward!

baijunyao:
Thanks.

PR #755, opened 2019-10-05
Respect Stripe-Should-Retry and Retry-After headers

rattrayalex-stripe:
r? @ob-stripe 
cc @richardm-stripe (because php)
cc @stripe/api-libraries 

reference: https://github.com/stripe/stripe-node/pull/692

### DIFF
--- a/lib/HttpClient/CurlClient.php
+++ b/lib/HttpClient/CurlClient.php
@@ -204,18 +204,6 @@ class CurlClient implements ClientInterface
             }
         }
 
-        // Create a callback to capture HTTP headers for the response
-        $rheaders = new Util\CaseInsensitiveArray();
-        $headerCallback = function ($curl, $header_line) use (&$rheaders) {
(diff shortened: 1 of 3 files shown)


lib/HttpClient/CurlClient.php
```diff
@@ -365,51 +364,41 @@ private function shouldRetry($errno, $isPost, $rcode, $rbody, $numRetries)
             return true;
         }
 
-        // 409 Conflict
-        if ($rcode === 409) {
-            return true;
+        // The API may ask us not to retry (eg; if doing so would be a no-op)
+        // or advise us to retry (eg; in cases of lock timeouts); we defer to that.
+        if ($rheaders['stripe-should-retry'] === 'false') {
```
ob-stripe: This works, but causes a warning if the key doesn't exist.

```suggestion
        if (isset($rheaders['stripe-should-retry']) && $rheaders['stripe-should-retry'] === 'false') {
```

Issue #760, opened 2019-10-11
Inform callback when retries are attempted


suadhuskic:
I'd like to be notified when a retry attempt is going to take place when network errors occur. I think i have an idea on how to add this, how can i push?

ob-stripe:
Hi @suadhuskic. Only Stripe employees may push directly to this repository, but you are welcome to fork the repository and submit a PR from the form and we'll be happy to take a look at it.

ob-stripe:
Version 7.4.0 adds support for setting a callback to be notified of the result of each HTTP request sent by the library (cf. #764).

PR #763, opened 2019-10-15
Inform clients on retries

suadhuskic:
This adds a property to the Stripe class where consumers can opt into being informed when retry attempts are made, this is useful to have for monitoring. 
For example, finding patterns that can help determine why network issues occur every Friday for a specific time range.

relates to:
https://github.com/stripe/stripe-php/issues/760


ob-stripe:
Hi @suadhuskic, thanks for the contribution.

Given that this is a fairly specific use case, I'm reluctant to expose it on the global `Stripe` object. I'll take a stab at a slightly different approach and will CC you on the PR for feedback.

ob-stripe:
Closing this in favor of #764, but thanks for the feature suggestion and the contribution!

PR #764, opened 2019-10-15
Add support for HTTP request monitoring callback

ob-stripe:
r? @rattrayalex-stripe 
cc @stripe/api-libraries @suadhuskic

Add support for passing a callback that is called after every HTTP request. This can be used for logging, monitoring, stats, or anything else.

Used like this:

```php
$curl = new CurlClient();
$curl->setRequestStatusCallback(function ($rbody, $rcode, $rheaders, $errno, $message, $willBeRetried, $numRetries) {
  // Do whatever
});
```

@suadhuskic This is a bit more complex than what you submitted but also more powerful, since it lets you access more information about the request (e.g. the HTTP status code, the request ID in the headers, etc.). I think it would also work for your use case, what do you think?

### DIFF
--- a/lib/HttpClient/CurlClient.php
+++ b/lib/HttpClient/CurlClient.php
@@ -46,6 +46,8 @@ class CurlClient implements ClientInterface
 
     protected $curlHandle = null;
 
+    protected $requestStatusCallback = null;
+
     /**
      * CurlClient constructor.
      *
@@ -122,6 +124,33 @@ class CurlClient implements ClientInterface
     public function setEnableHttp2($enable)
     {
         $this->enableHttp2 = $enable;
+    }
+
+    /**
+     * @return callable|null
+     */
+    public function getRequestStatusCallback()
+    {
+        return $this->requestStatusCallback;
+    }
+
+    /**
+     * Sets a callback that is called after each request. The callback will
+     * receive the following parameters:
+     *
+     * 1. string $rbody The response body
+     * 2. integer $rcode The response status code
+     * 3. \Stripe\Util\CaseInsensitiveArray $rheaders The response headers
+     * 4. integer $errno The curl error number
+     * 5. string|null $message The curl error message
+     * 6. boolean $shouldRetry Whether the request will be retried
+     * 7. integer $numRetries The number of the retry attempt
+     *
+     * @param callable|null $requestStatusCallback
+     */
+    public function setRequestStatusCallback($requestStatusCallback)
+    {
+        $this->requestStatusCallback = $requestStatusCallback;
     }
 
     // USER DEFINED TIMEOUTS
@@ -250,6 +279,7 @@ class CurlClient implements ClientInterface
         while (true) {
             $rcode = 0;
             $errno = 0;
+            $message = null;
 
             // Create a callback to capture HTTP headers for the response
             $rheaders = new Util\CaseInsensitiveArray();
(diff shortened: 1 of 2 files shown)


tests/Stripe/HttpClient/CurlClientTest.php
```diff
@@ -297,4 +297,24 @@ public function testResponseHeadersCaseInsensitive()
...
         $this->assertEquals($headers['request-id'], $headers['Request-Id']);
     }
+
+    public function testSetRequestStatusCallback()
+    {
+        try {
+            $called = false;
+
+            $curl = new CurlClient();
+            $curl->setRequestStatusCallback(function ($rbody, $rcode, $rheaders, $errno, $message, $shouldRetry, $numRetries) use (&$called) {
```
rattrayalex-stripe: can we add a test that each arg is roughly as desired?

lib/HttpClient/CurlClient.php
```diff
@@ -124,6 +126,33 @@ public function setEnableHttp2($enable)
...
+     * Sets a callback that is called after each request. The callback will
+     * receive the following parameters:
+     *
+     * 1. string $rbody The response body
+     * 2. integer $rcode The response status code
+     * 3. \Stripe\Util\CaseInsensitiveArray $rheaders The response headers
+     * 4. integer $errno The curl error number
+     * 5. string|null $message The curl error message
+     * 6. boolean $shouldRetry Whether the request will be retried
+     * 7. integer $numRetries The number of the retry attempt
```
rattrayalex-stripe: Curious, why include these?

tests/Stripe/HttpClient/CurlClientTest.php
ob-stripe: 👍

ob-stripe:
ptal @rattrayalex-stripe

lib/HttpClient/CurlClient.php
rattrayalex-stripe: Ah, I see, a user wanted it: https://github.com/stripe/stripe-php/pull/763

suadhuskic:
Doesn't seem like a fun usage from a client perspective knowing all network retry logic is on the `Stripe` class and now i have to manually set this curl client up before attempting my charge.

Currently I have:
```php
   Stripe::setApiKey('magic');
   Stripe::$maxNetworkRetries = 3;

   $paymentIntent = PaymentIntent::create($requestParams, $requestOptions);
```
This is why my proposal was what it was in https://github.com/stripe/stripe-php/pull/763 to keep the pattern there. Thank you for quickly putting this up!

ob-stripe:
> Doesn't seem like a fun usage from a client perspective knowing all network retry logic is on the Stripe class and now i have to manually set this curl client up

That's a fair criticism. The max retries number should really be a property of `CurlClient` since that's the class actually handling the retries. I want to clean up the interfaces of both the `Stripe` global class and `CurlClient`, but that will have to wait for a future major version since it would likely include some non-backwards compatible changes.

ob-stripe:
Released as 7.4.0.

jmraker:
As a user who alters the CurlClient.php source to add logging it's not quite there.  I log the url +parameters of the api call and I don't see a way to do that.  It would be nice if it also passed the $opts array.

ob-stripe:
@jmraker Interesting use case! I think it might be more appropriate to have two callbacks: one _before_ the request with the request properties (method, URL, parameters) and one _after_ the request with the response properties (status code, response body, etc.).

I don't want to rush into things though as I'm planning on some major changes to the HTTP client interface. In the meantime, I think a cleaner workaround that would not require you to modify the library's source code would be to subclass `CurlClient`:

```php
class LoggingCurlClient extends \Stripe\HttpClient\CurlClient
{
  public function request($method, $absUrl, $headers, $params, $hasFile)
  {
    // log request parameters

    list($rbody, $rcode, $rheaders) = parent::request($method, $absUrl, $headers, $params, $hasFile);

    // log response parameters

    return [$rbody, $rcode, $rheaders];
  }
}

$loggingClient = new LoggingCurlClient();
\Stripe\ApiRequestor::setHttpClient($loggingClient);
```

PR #765, opened 2019-10-17
Don't ignore null values in parameters

### LINKED ISSUE #761: Setting metadata to 'null' should unset it
stripe/stripe-php: v6.43.0

When updating an object to unset metadata, I'd expect setting the metadata key to `null` to unset it. But `null` doesn't have any effect (unlike with the Stripe Node SDK) and instead an empty string is required.

## This has no effect:
```php
Stripe\Customer::update(
    $customerId,
    ['metadata' => [
        'some_metadata_key' => null,
    ]]
);
```

## This unsets `some_metadata_key`:
```php
Stripe\Customer::update(
    $customerId,
    ['metadata' => [
        'some_metadata_key' => '',
    ]]
);
```

InternetJones from the #stripe IRC agrees that unsetting via `null` would be a little more expected/consistent behaviour. :-)

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

No longer ignore null values in parameter arrays, instead encode them as empty strings so that they're sent to the API.

I'm fairly sure this should be safe. Parameters can come from two different sources:
1. provided directly by the user (in which case it's safe to assume that any `null` value has been explicitly inserted by the user and they want it to be sent in the request)
2. computed automatically in `->save()` calls. In this case there shouldn't be any `null` values, because when we serialize update parameters we're already inserting empty strings as the parameter value for fields that been `null`ed: https://github.com/stripe/stripe-php/blob/96c59742329c4a9b3f531e91d64e432c82654b91/lib/StripeObject.php#L362-L363

This is difficult to test end-to-end because of the current HTTP client infrastructure in stripe-php, but I've verified this works:

```php
$customer = \Stripe\Customer::update('cus_123', ['metadata' => ['foo' => null]]);
```

Fixes #761.

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -79,6 +79,8 @@ class ApiRequestor
     {
         if ($d instanceof ApiResource) {
             return Util\Util::utf8($d->id);
+        } elseif ($d === null) {
+            return '';
         } elseif ($d === true) {
             return 'true';
         } elseif ($d === false) {
--- a/tests/Stripe/ApiRequestorTest.php
+++ b/tests/Stripe/ApiRequestorTest.php
@@ -34,6 +34,11 @@ class ApiRequestorTest extends TestCase
         $v = false;
         $enc = $method->invoke(null, $v);
         $this->assertSame('false', $enc);
+
+        // Encodes null
+        $v = null;
+        $enc = $method->invoke(null, $v);
+        $this->assertSame('', $enc);
     }
 
     public function testHttpClientInjection()


ob-stripe:
Closing for now since this is a potentially dangerous behavior change. (I.e. users may be relying on the fact that `null` values in parameter arrays are _not_ sent to the API.)

remi-stripe:
To provide a bit more context, the problem with this change is that this is in practice a breaking change. For example today if you do this:
```
$customer = "cus_123";
\Stripe\Charge::create([
  "amount" => 2000,
  "currency" => "usd",
  "customer" => $customer,
  "source" => $card,
  "description" => "Charge for jenny.rosen@example.com"
]);
```
This works and charges the customer's default source, because $card is undefined and just ignored.

In principle, it's not great and our libraries should catch this or the API should reject the request (which your change solves) but in practice today the `source` parameter is just ignored. 

By making this change, any integration that has that bug would suddenly be broken because they'd get API errors. It's likely a great thing in practice because their integration has a bug, but it's a breaking change in that case and would need a major. And a major version which such a subtle change and nothing else is quite dangerous.

PR #766, opened 2019-10-17
[wip] Instantiated services

ob-stripe:
cc @remi-stripe 

Pretty experimental yet, but it works. I will polish this up and split it into smaller PRs.

Sample syntax:

```php
require_once('vendor/autoload.php');

// Instantiate client
$stripe = new \Stripe\StripeClient("sk_test_123");

// Retrieve method
$coupon = $stripe->coupons->retrieve("mycoupon");

// Create method, namespaced resource
$issuingCard = $stripe->issuing->cards->create(["authorization_controls" => ["max_amount" => 123]]);

// Custom method, namespaced resource
$details = $stripe->issuing->cards->details("ic_123");

// Update method, nested polymorphic resource
$updatedPaymentSource = $stripe->paymentSources->update("cus_123", " card_123", ["metadata" => ["key" => "value"]]);
```

PR #767, opened 2019-10-17
New `StripeClient` class

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

(Note: this PR targets the `integration-services` branch, not `master`.)

Adds a new `StripeClient` class, whose purpose is to hold configuration values:
- the API key
- the client ID (for OAuth requests)
- the base URLs for regular API requests, file API requests and OAuth requests

This client will be used by services to issue requests.

Also defines a `StripeClientInterface`, so users can easily write mock clients.

### DIFF
--- a/lib/StripeClient.php
+++ b/lib/StripeClient.php
@@ -1,0 +1,146 @@
+<?php
+
+namespace Stripe;
+
+class StripeClient implements StripeClientInterface
+{
+    /**
+     * @var string Default base URL for Stripe's API.
+     */
+    const DEFAULT_API_BASE = 'https://api.stripe.com';
+
+    /**
+     * @var string Default base URL for Stripe's OAuth API.
+     */
+    const DEFAULT_CONNECT_BASE = 'https://connect.stripe.com';
+
+    /**
+     * @var string Default base URL for Stripe's Files API.
+     */
+    const DEFAULT_FILES_BASE = 'https://files.stripe.com';
+
+    private $apiKey;
+    private $clientId;
+
+    private $apiBase;
+    private $connectBase;
+    private $filesBase;
+
+    /**
+     * Initializes a new instance of the {@link StripeClient} class.
+     *
+     * @param string|null $apiKey The API key used by the client to make requests.
+     * @param string|null $clientId The client ID used by the client in OAuth requests.
+     * @param string|null $apiBase The base URL for Stripe's API. Defaults to {@link DEFAULT_API_BASE}.
+     * @param string|null $connectBase The base URL for Stripe's OAuth API. Defaults to {@link DEFAULT_CONNECT_BASE}.
+     * @param string|null $filesBase The base URL for Stripe's Files API. Defaults to {@link DEFAULT_FILES_BASE}.
+     */
+    public function __construct(
+        $apiKey,
+        $clientId = null,
+        $apiBase = null,
+        $connectBase = null,
+        $filesBase = null
+    ) {
+        if (!is_null($apiKey) && ($apiKey === '')) {
+            $msg = "API key cannot be the empty string.";
+            throw new \Stripe\Exception\InvalidArgumentException($msg);
+        }
+        if (!is_null($apiKey) && (preg_match('/\s/', $apiKey))) {
+            $msg = "API key cannot contain whitespace.";
+            throw new \Stripe\Exception\InvalidArgumentException($msg);
+        }
+
+        $this->apiKey = $apiKey;
+        $this->clientId = $clientId;
+
+        $this->apiBase = $apiBase ?: self::DEFAULT_API_BASE;
+        $this->connectBase = $connectBase ?: self::DEFAULT_CONNECT_BASE;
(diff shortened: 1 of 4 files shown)


lib/StripeClient.php
```diff
@@ -0,0 +1,146 @@
...
+     * Initializes a new instance of the {@link StripeClient} class.
+     *
+     * @param string|null $apiKey The API key used by the client to make requests.
+     * @param string|null $clientId The client ID used by the client in OAuth requests.
+     * @param string|null $apiBase The base URL for Stripe's API. Defaults to {@link DEFAULT_API_BASE}.
+     * @param string|null $connectBase The base URL for Stripe's OAuth API. Defaults to {@link DEFAULT_CONNECT_BASE}.
+     * @param string|null $filesBase The base URL for Stripe's Files API. Defaults to {@link DEFAULT_FILES_BASE}.
+     */
+    public function __construct(
+        $apiKey,
```
remi-stripe: should be nullable

lib/StripeClient.php
```diff
@@ -0,0 +1,146 @@
...
+     *
+     * @param string $method The HTTP method.
+     * @param string $path The path of the request.
+     * @param array $params The parameters of the request.
+     * @param array|\Stripe\RequestOptions $opts The special modifiers of the request.
+     * @return \Stripe\StripeObject The object returned by Stripe's API.
+     */
+    public function request($method, $path, $params, $opts)
+    {
+        $opts = \Stripe\Util\RequestOptions::parse($opts);
```
remi-stripe: we should make sure to remove all legacy behaviour when parsing options

also validate keys we don't know (error), validate null/empty values, etc.

lib/StripeClient.php
```diff
@@ -0,0 +1,146 @@
...
+        $baseUrl = $opts->apiBase ?: $this->getApiBase();
+        $requestor = new \Stripe\ApiRequestor($this->apiKeyForRequest($opts), $baseUrl);
+        list($response, $opts->apiKey) = $requestor->request($method, $path, $params, $opts->headers);
```
remi-stripe: it's weird to me that we override $opts to store the API key for the next line. You said it's to handle `save` on the object but we should consider banning save if you use a Client.

tests/Stripe/StripeClientTest.php
```diff
@@ -0,0 +1,59 @@
...
+        $this->assertNull($client->getApiKey());
+    }
+
+    /**
+     * @expectedException \Stripe\Exception\InvalidArgumentException
+     * @expectedExceptionMessage API key cannot be the empty string.
+     */
+    public function testCtorThrowsIfApiKeyIsEmpty()
+    {
+        $client = new StripeClient('');
```
remi-stripe: double quote

tests/Stripe/StripeClientTest.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        $client = new StripeClient("sk_test_123\n");
+    }
+
+    public function testRequestWithClientApiKey()
+    {
+        $client = new StripeClient("sk_test_123", null, MOCK_URL);
+        $charge = $client->request("get", "/v1/charges/ch_123", [], []);
+        $this->assertNotNull($charge);
+        $this->assertEquals("ch_123", $charge->id);
```
remi-stripe: the test implies you test the API key, you can look at $charge to confirm it's the right key
But really we shouldn't have the $charge have the API key (people dump it with var_dump).

lib/StripeClient.php
remi-stripe: We should also override RequestOptions so that when you use var_dump() you hide/redact the API key

ob-stripe: Opened a separate PR for this: #768.

lib/StripeClient.php
ob-stripe: To be clear, the argument _is_ nullable, we're just not providing a default value. On second thought, I think this is preferable: it'll likely be very uncommon to create a client without an API key, so I think it's more helpful to require the argument. Users who don't want to set an API key on the client will just have to pass `null` explicitly.

lib/StripeClient.php
ob-stripe: We'd need to use a different class (or a subclass) to do this additional validation while not affecting existing integrations, but I think it's a good idea. I'll tackle it in a different PR.

lib/StripeClient.php
ob-stripe: 👍 Will try to think of a way to avoid storing the API key and freeze the resulting object when it's instantiated by `StripeClient`.

virgofx:
Any chance we can get some updated docs/examples on how to use the Stripe client?

ob-stripe:
@virgofx This isn't released yet (this PR was merged into another integration branch), but we will of course provide documentation when the feature is released.

You can get a sneak peek of what the API will look like here: https://github.com/stripe/stripe-php/pull/766#issue-329081312

PR #768, opened 2019-10-18
Redact API key in RequestOpts debug info

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Redacts the API key in `RequestOpts` debug info (when using `var_dump` or `print_r`).

### DIFF
--- a/lib/Util/RequestOptions.php
+++ b/lib/Util/RequestOptions.php
@@ -25,6 +25,15 @@ class RequestOptions
         $this->apiBase = $base;
     }
 
+    public function __debugInfo()
+    {
+        return [
+            'apiKey' => $this->redactedApiKey(),
+            'headers' => $this->headers,
+            'apiBase' => $this->apiBase,
+        ];
+    }
+
     /**
      * Unpacks an options array and merges it into the existing RequestOptions
      * object.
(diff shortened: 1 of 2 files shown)


lib/Util/RequestOptions.php
```diff
@@ -25,6 +25,21 @@ public function __construct($key = null, $headers = [], $base = null)
         $this->apiBase = $base;
     }
 
+    public function __debugInfo()
+    {
+        $redactedApiKey = strlen($this->apiKey) >= 12
+            ? (substr($this->apiKey, 0, 8)
+                . str_repeat('*', strlen($this->apiKey) - 12)
+                . substr($this->apiKey, -4))
+            : $this->apiKey;
```
remi-stripe: While that works, legacy keys didn't start with `sk_test` or `sk_live`. Can we just redact all but last 4 if they don't start with one of those?

ob-stripe: Fixed.

ob-stripe:
> Not sure how to confirm this works with var_dump too though

You can't capture the output of `var_dump` so it's not possible to write an automated test for it. I think we'll just have to trust [PHP's documentation](https://www.php.net/manual/en/language.oop5.magic.php#object.debuginfo) here :)

ob-stripe:
ptal @remi-stripe

ob-stripe:
Released as 7.5.0.

PR #772, opened 2019-10-18
Add `AbstractService` and a couple of concrete services

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

(Note: this PR targets the `integration-services` branch, not `master`.)

Adds the `AbstractService` base class, and a couple of concrete services:
- `CouponService` (implements all standard CRUDL methods)
- `PaymentIntentService` (implements CRUL + two custom methods)

### DIFF
--- a/lib/Service/CouponService.php
+++ b/lib/Service/CouponService.php
@@ -1,0 +1,74 @@
+<?php
+
+namespace Stripe\Service;
+
+class CouponService extends AbstractService
+{
+    public function basePath()
+    {
+        return "/v1/coupons";
+    }
+
+    /**
+     * List all coupons.
+     *
+     * @param array $params
+     * @param array $opts
+     * @return \Stripe\Collection
+     */
+    public function all($params = [], $opts = [])
+    {
+        return $this->allObjects($params, $opts);
+    }
+
+    /**
+     * Create a coupon.
+     *
+     * @param array $params
+     * @param array $opts
+     * @return \Stripe\Coupon
+     */
+    public function create($params = [], $opts = [])
+    {
+        return $this->createObject($params, $opts);
+    }
+
+    /**
+     * Delete a coupon.
+     *
+     * @param string $id
+     * @param array $params
+     * @param array $opts
+     * @return \Stripe\Coupon
+     */
+    public function delete($id, $params = [], $opts = [])
+    {
+        return $this->deleteObject($id, $params, $opts);
+    }
+
+    /**
+     * Retrieve a coupon.
+     *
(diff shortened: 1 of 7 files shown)


lib/Service/CouponService.php
```diff
@@ -0,0 +1,74 @@
...
+    {
+        return "/v1/coupons";
+    }
+
+    /**
+     * List all coupons.
+     *
+     * @param array $params
+     * @param array $opts
+     * @return \Stripe\Coupon
```
remi-stripe: it should be a list of coupons right?

lib/Service/CouponService.php
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Update a coupon.
+     *
+     * @param string $id
+     * @param array $params
+     * @param array $opts
+     * @return \Stripe\Coupon
+     */
+    public function update($id, $params = [], $opts = [])
```
remi-stripe: Does this imply all services must have this method? How does it work for things that can't be updated/deleted?

tests/Stripe/Service/AbstractServiceTest.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace Stripe\Service;
+
+class AbstractServiceTest extends \Stripe\TestCase
+{
+    const TEST_RESOURCE_ID = "25OFF";
```
remi-stripe: really minor but can we do `RESOURCE_ID` or similar instead. Trying to get rid of the fact that you can do your own ids on some objects

lib/Service/CouponService.php
ob-stripe: Nice catch. Yes, the return type should be `\Stripe\Collection`.

remi-stripe: Can it be special-cased for Coupons? so that `$ret->data` is "typed" in IDEs?

lib/Service/CouponService.php
ob-stripe: No, all methods are implemented on a service-by-service basis. The helper methods like `updateObject` are protected so can't be called externally.

tests/Stripe/Service/AbstractServiceTest.php
ob-stripe: 👍

lib/Service/CouponService.php
ob-stripe: Unfortunately AFAIK no, PHPDoc has no support for "generics".

PR #775, opened 2019-10-22
Add PHPStan for static analysis

nauxliu:


ob-stripe:
Hi @nauxliu, thanks a lot for the PR!

Unfortunately, because we still need to support PHP 5.6 and 7.0 for the time being, we are unable to merge this PR as-is.

I don't think that Composer supports conditional dependencies (e.g. "only install this dependency if PHP >= 7.1"). We'd probably need to do something custom to install PHPStan as a PHAR archive, both locally and in Travis CI. Is that something you'd be willing to look into? If not, I can try to take a look.

In the meantime, if you want to submit a separate PR for the actual bug fixes that are in this PR, I'd be happy to review and merge it.

nauxliu:
@ob-stripe Seems no good solutions here, I'll just remove PHPStan from composer requirements so this PR will only contain bug fixes and PHPStan config.

nauxliu:
Is there an ETA for removing the support for the end of life PHP versions to courage people to upgrade to safe versions since they don't even have security support.

ob-stripe:
I thought it was a shame to not be using PHPStan going forwards since you got us started here, so I took a stab at it in a new PR: #787. I worked around the old PHP versions issue by setting up a Makefile to easily install PHPStan without using Composer.

> Is there an ETA for removing the support for the end-of-life PHP versions to encourage people to upgrade to safe versions since they don't even have security support.

No ETA at this time. We monitor usage numbers pretty closely and drop support for old versions once we feel confident that we're not preventing too many users from upgrading to new versions of the library. It's definitely not ideal, but many PHP users don't have any control over which PHP version they're using (they simply use whatever version is set up by their hosting provider, and sadly not all providers do a great job of keeping PHP up to date).

nauxliu:
@ob-stripe Thanks for the explanation, it makes sense.

PR #781, opened 2019-10-25
Fix telemetry header

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

I just noticed a bug with the telemetry header: it's serialized as `{"last_request_metrics":{"request_id":["req_123"],"request_duration_ms":234}}`, i.e. `request_id` is an array instead of a string.

This is because the response headers returned by the requestor is a map of string to list of strings (because a header may be repeated and have multiple values).

In this case we're fairly confident that there will only be a single value for the `Request-Id` header, so we fix this by fetching the first value.

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -368,9 +368,9 @@ class ApiRequestor
             $hasFile
         );
 
-        if (isset($rheaders['request-id'])) {
+        if (isset($rheaders['request-id']) && isset($rheaders['request-id'][0])) {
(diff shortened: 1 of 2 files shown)


ob-stripe:
> Approving but I don't understand why the test didn't change

We're stubbing the `request` method in the test, and we were returning the incorrect type (i.e. a map of string->string instead of a a map of string->list of strings). This is now fixed.

ob-stripe:
Released as 7.7.1.

PR #783, opened 2019-10-29
Add service with methods for nested resources

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Add `CustomerService`, a service that includes methods for nested resources.


lib/Service/CustomerService.php
```diff
@@ -0,0 +1,267 @@
+<?php
+
+namespace Stripe\Service;
+
+class CustomerService extends AbstractService
+{
+    const BALANCE_TRANSACTIONS = "balance_transactions";
+    const SOURCES = "sources";
+    const TAX_IDS = "tax_ids";
```
remi-stripe: I think we should prefix all those constants by something since we are also going to have various constants in all classes about types and such. Avoids a naming conflict with a purely internal constant. If we can we should even find a way to make those purely internal.

lib/Service/AbstractService.php
```diff
@@ -45,31 +45,61 @@ protected function allObjects($params, $opts)
         return $this->request("get", $this->basePath(), $params, $opts);
     }
 
+    protected function allNestedObjects($nestedPath, $parentId, $params, $opts)
+    {
+        return $this->request("get", $this->baseNestedPath($parentId, $nestedPath), $params, $opts);
+    }
+
```
remi-stripe: Minor but as much as I like alphabetization, I think I'd rather have all nested methods together. Maybe an argument to do nestedObjectsAll or similar? No strong feels though.

lib/Service/CustomerService.php
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    /**
+     * Create a source.
+     *
+     * @param string $customerId
+     * @param array $params
+     * @param array $opts
+     * @return \Stripe\Account|\Stripe\AlipayAccount|\Stripe\BankAccount|\Stripe\BitcoinReceiver|\Stripe\Card|\Stripe\Source
+     */
```
remi-stripe: You technically can't get an `Account` back here at all. You also can't get `AlipayAccount` or `BitcoinReceiver` I guess).

lib/Service/CustomerService.php
```diff
@@ -0,0 +1,267 @@
...
+        return $this->createNestedObject(self::SOURCES, $customerId, $params, $opts);
+    }
+
+    /**
+     * Create a tax ID.
+     *
+     * @param string $customerId
+     * @param array $params
+     * @param array $opts
+     * @return \Stripe\TaxId
```
remi-stripe: We likely need to talk more about whether to prefix by the parent resource name or not, so TaxId versus CustomerTaxId. Unrelated to your PR, just something to think about

lib/Service/CustomerService.php
```diff
@@ -0,0 +1,267 @@
...
+    public function retrieve($id, $params = [], $opts = [])
+    {
+        return $this->retrieveObject($id, $params, $opts);
+    }
+
+    /**
+     * Retrieve a balance transaction.
+     *
+     * @param string $customerId
+     * @param string $id
```
remi-stripe: should the second parameter have the name `balanceTransactionId` to be clearer?

lib/Service/CustomerService.php
```diff
@@ -0,0 +1,267 @@
...
+     * @param array $params
+     * @param array $opts
+     * @return \Stripe\Account|\Stripe\AlipayAccount|\Stripe\BankAccount|\Stripe\BitcoinReceiver|\Stripe\Card|\Stripe\Source
+     */
+    public function updateSource($customerId, $id, $params = [], $opts = [])
+    {
+        return $this->updateNestedObject(self::SOURCES, $customerId, $id, $params, $opts);
+    }
+}
```
remi-stripe: We should talk about how we'll do custom methods on nested resources like verifying ACH.

lib/Service/CustomerService.php
ob-stripe: I don't think we'll need other constants in the service classes. We might want to remove the constants entirely anyway once we codegen the service classes. Going to keep as-is for now, and we can revisit in another PR.

> If we can we should even find a way to make those purely internal.

Support for private constants was added in PHP 7.1 :p

lib/Service/AbstractService.php
ob-stripe: Going to keep as-is for now, I think we'll probably get rid of the helper methods anyway when we codegen the services and just use `request()` directly.

lib/Service/CustomerService.php
ob-stripe: The spec says otherwise :p But fair enough, I'll remove these.

remi-stripe: I'm fine to stick to the spec in that case honestly it's likely best

lib/Service/CustomerService.php
ob-stripe: I added a `verifySource` method.

lib/Service/CustomerService.php
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    /**
+     * Verify a source.
+     *
+     * @param string $customerId
+     * @param string $id
+     * @param array $params
+     * @param array $opts
+     * @return \Stripe\BankAccount
```
remi-stripe: Technically this also works for ACH `Source`

lib/Service/CustomerService.php
ob-stripe: Tax IDs have an `object` key with value `tax_id`, so I think `TaxId` is the appropriate name even if they are always nested under customers.

lib/Service/CustomerService.php
ob-stripe: Hm, maybe... Will keep as-is for now and add a note in the integration PR to decide if we want to change it.

PR #784, opened 2019-10-29
Add FileService

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Add `FileService`.

### DIFF
--- a/lib/Service/FileService.php
+++ b/lib/Service/FileService.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace Stripe\Service;
+
+class FileService extends AbstractService
+{
+    public function basePath()
+    {
+        return "/v1/files";
+    }
+
+    /**
+     * List all files.
+     *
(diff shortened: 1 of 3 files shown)


lib/Service/FileService.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+namespace Stripe\Service;
+
+class FileService extends AbstractService
+{
+    public function basePath()
+    {
+        return "/v1/files";
+    }
```
remi-stripe: curious how you know to use `files.stripe.com` instead of `api.stripe.com`

ob-stripe: Here: https://github.com/stripe/stripe-php/blob/0bc7bd4c9e33d0cfb695e8016b29ef72ac82462c/lib/Service/FileService.php#L35.

It's only used for file creation requests. Retrieve and list requests are sent to the normal API server.

PR #787, opened 2019-10-30
Use PHPStan in Travis CI

ob-stripe:
cc @stripe/api-libraries @nauxliu

Add support for [PHPStan](https://github.com/phpstan/phpstan), a static analyzer for PHP.

Because PHPStan requires PHP 7.1 or more recent, we cannot add it as a dev dependency in `composer.json`. I worked around this by setting up a Makefile to download PHPStan separately as a self-contained PHAR archive.

PHPStan offers different analysis levels, from 0 (loosest) to 7 (strictest). Right now I've set it at 1 and fixed most issues it reported. We can increase it and fix more issues in future PRs.

There are two issues I could not fix without some refactoring, so what I did was create a [baseline file](https://medium.com/@ondrejmirtes/phpstans-baseline-feature-lets-you-hold-new-code-to-a-higher-standard-e77d815a5dff), which is basically the PHPStan equivalent of the `.rubucop_todo.yml` file. The `phpstan-baseline` make target can be used to regenerate the file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ cache:
 before_install:
   # Install various build dependencies. We use `travis_retry` because Composer
   # will occasionally fail intermittently.
-  - travis_retry composer install
+  - travis_retry make vendor
 
   # Unpack and start stripe-mock so that the test suite can talk to it
   - |
@@ -68,5 +68,6 @@ before_install:
 script:
   - ./build.php ${AUTOLOAD}
   - ./vendor/bin/php-cs-fixer fix -v --dry-run --using-cache=no .
+  - if [[ `php -r "echo \version_compare(PHP_VERSION, '7.1', '>=');"` && $AUTOLOAD == 1 ]]; then make phpstan; fi
 
 after_script: ./vendor/bin/coveralls -v
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -129,7 +129,7 @@ class ApiRequestor
         if (!is_array($resp) || !isset($resp['error'])) {
             $msg = "Invalid response object from API: $rbody "
               . "(HTTP response code was $rcode)";
-            throw new Exception\UnexpectedValueException($msg, $rcode, $rbody, $resp, $rheaders);
+            throw new Exception\UnexpectedValueException($msg);
         }
 
         $errorData = $resp['error'];
--- a/lib/HttpClient/CurlClient.php
+++ b/lib/HttpClient/CurlClient.php
@@ -38,6 +38,9 @@ class CurlClient implements ClientInterface
 
     protected $defaultOptions;
 
+    /** @var \Stripe\Util\RandomGenerator */
+    protected $randomGenerator;
+
     protected $userAgentInfo;
 
     protected $enablePersistentConnections = true;
@@ -380,7 +383,7 @@ class CurlClient implements ClientInterface
      *
      * @param int $errno
      * @param int $rcode
-     * @param array|CaseInsensitiveArray $rheaders
+     * @param array|\Stripe\Util\CaseInsensitiveArray $rheaders
      * @param int $numRetries
      *
      * @return bool
@@ -435,7 +438,7 @@ class CurlClient implements ClientInterface
      * Provides the number of seconds to wait before retrying a request.
      *
      * @param int $numRetries
-     * @param array|CaseInsensitiveArray $rheaders
+     * @param array|\Stripe\Util\CaseInsensitiveArray $rheaders
      *
      * @return int
      */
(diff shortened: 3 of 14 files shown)


phpstan-baseline.neon
```diff
@@ -0,0 +1,15 @@
...
+	ignoreErrors:
+		-
+			message: "#^Access to undefined constant Stripe\\\\ApiResource\\:\\:OBJECT_NAME\\.$#"
+			count: 1
+			path: lib/ApiResource.php
+
+		-
+			message: "#^Access to undefined constant Stripe\\\\SingletonApiResource\\:\\:OBJECT_NAME\\.$#"
+			count: 1
+			path: lib/SingletonApiResource.php
```
ob-stripe: Note that these two errors are "real" issues: there is no guarantee that the `OBJECT_NAME` constant will exist in these methods, and there is no mechanism to force a derived class to declare a specific static constant. We should probably replace the `OBJECT_NAME` constant with a `getObjectName()` method that would be declared as `abstract` in `ApiResource`.

This would be a non-trivial refactor though and the current code has been working for a long time, so let's not leave it as is for now.

lib/HttpClient/CurlClient.php
```diff
@@ -38,6 +38,9 @@ public static function instance()
 
     protected $defaultOptions;
 
+    /** @var \Stripe\Util\RandomGenerator */
+    protected $randomGenerator;
+
```
remi-stripe: Can you say more about this change?

lib/ApiRequestor.php
```diff
@@ -129,7 +129,7 @@ public function handleErrorResponse($rbody, $rcode, $rheaders, $resp)
         if (!is_array($resp) || !isset($resp['error'])) {
             $msg = "Invalid response object from API: $rbody "
               . "(HTTP response code was $rcode)";
-            throw new Exception\UnexpectedValueException($msg, $rcode, $rbody, $resp, $rheaders);
+            throw new Exception\UnexpectedValueException($msg);
```
remi-stripe: Can you say more about this change? I looked in the class but didn't find much context

lib/StripeObject.php
```diff
@@ -361,7 +361,7 @@ public function serializeParamsValue($value, $original, $unsaved, $force, $key =
         // user intended it to.
         if ($value === null) {
             return "";
-        } elseif (($value instanceof APIResource) && (!$value->saveWithParent)) {
+        } elseif (($value instanceof ApiResource) && (!$value->saveWithParent)) {
```
remi-stripe: This one seems worrying. Does that mean the instanceof never worked in that `if`?

tests/Stripe/HttpClient/CurlClientTest.php
```diff
@@ -60,7 +81,7 @@ private function setInitialNetworkRetryDelay($initialNetworkRetryDelay)
 
     private function createFakeRandomGenerator($returnValue = 1.0)
     {
-        $fakeRandomGenerator = $this->createMock('Stripe\Util\RandomGenerator', ['randFloat']);
+        $fakeRandomGenerator = $this->createMock('Stripe\Util\RandomGenerator');
```
remi-stripe: why this change?

lib/HttpClient/CurlClient.php
ob-stripe: We're just declaring a class property that we were already using. PHP doesn't force you to declare class properties, but it's good practice to do so.

tests/Stripe/HttpClient/CurlClientTest.php
ob-stripe: The [`createMock()`](https://phpunit.de/manual/5.7/en/test-doubles.html) method only expects a single argument. Don't ask me how it worked before.

.travis.yml
```diff
@@ -68,5 +68,6 @@ before_install:
 script:
   - ./build.php ${AUTOLOAD}
   - ./vendor/bin/php-cs-fixer fix -v --dry-run --using-cache=no .
+  - if [[ $TRAVIS_PHP_VERSION == '7.3' && $AUTOLOAD == 1 ]]; then make phpstan; fi
```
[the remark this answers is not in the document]
ob-stripe: I think that's a good idea and wanted to do something like that, but comparing version strings in bash is kinda hard. I'll try to work around it though.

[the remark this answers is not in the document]
ob-stripe: I updated the condition to use PHP's `version_compare` method. Kind of a hack, but it works :)

ob-stripe:
Thanks again @nauxliu for getting the ball rolling with this!

.travis.yml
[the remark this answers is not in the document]
brandur-stripe: Pretty succinct actually. Nice!

PR #788, opened 2019-10-31
Early codegen fixes

ob-stripe:
r? @remi-stripe @marko-stripe
cc @stripe/api-libraries 

A bunch of fixes I pulled from @marko-stripe's almost-ready codegen branch:
- fix a lot of PHPDoc blocks
- use `$opts` instead of `$options` everywhere
- standardize all but one methods to accept `$params` and `$opts`
    - the only exception is `Dispute.close()` which already accepts `$opts`, so adding `$params` first would be a breaking change. I left a TODO so we can fix in the next major version

### DIFF
--- a/lib/Event.php
+++ b/lib/Event.php
@@ -9,11 +9,11 @@ namespace Stripe;
  * @property string $object
  * @property string $account
  * @property string $api_version
- * @property int    $created
- * @property mixed  $data
- * @property bool   $livemode
- * @property int    $pending_webhooks
- * @property mixed  $request
+ * @property int $created
+ * @property mixed $data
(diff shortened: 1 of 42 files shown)


lib/Event.php
```diff
@@ -9,19 +9,23 @@
...
- * @property bool   $livemode
- * @property int    $pending_webhooks
- * @property mixed  $request
+ * @property int $created
+ * @property mixed $data
+ * @property bool $livemode
+ * @property int $pending_webhooks
+ * @property mixed $request
  * @property string $type
+ * @property string $account
```
remi-stripe: should be only once

PR #790, opened 2019-11-05
Mark nullable fields in PHPDoc

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

- Mark nullable fields in PHPDoc
- Use fully qualified names everywhere for `\Stripe\Collection`, `\Stripe\StripeObject`, etc.

### DIFF
--- a/lib/Account.php
+++ b/lib/Account.php
@@ -7,8 +7,8 @@ namespace Stripe;
  *
  * @property string $id
  * @property string $object
- * @property mixed $business_profile
- * @property string $business_type
+ * @property mixed|null $business_profile
+ * @property string|null $business_type
  * @property mixed $capabilities
  * @property bool $charges_enabled
  * @property mixed $company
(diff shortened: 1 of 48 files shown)


lib/Radar/ValueListItem.php
```diff
@@ -17,10 +17,10 @@
  */
 class ValueListItem extends \Stripe\ApiResource
 {
-    const OBJECT_NAME = "radar.value_list_item";
+    const OBJECT_NAME = 'radar.value_list_item';
 
-    use \Stripe\ApiOperations\All;
     use \Stripe\ApiOperations\Create;
     use \Stripe\ApiOperations\Delete;
+    use \Stripe\ApiOperations\All;
```
remi-stripe: Nice catch but this needs a test

ob-stripe: Actually that was just the codegen tool moving the trait, not adding it. I've reverted the change.

ob-stripe:
Released as 7.8.0.

ruudk:
Thanks so much 💙

PR #794, opened 2019-11-05
Add phpdoc to ApiResponse

ruudk:


lib/ApiResponse.php
```diff
@@ -11,9 +11,24 @@
...
+     */
     public $body;
+
+    /**
+     * @var array|null
+     */
     public $json;
+
+    /**
+     * @var int
```
ob-stripe:
```suggestion
     * @var int
```

The trailing whitespace is causing the linter to fail.

ruudk: Oops, fixed.

ob-stripe:
Released as 7.9.0. Thanks!

PR #796, opened 2019-11-06
Convert `filters` of `Collection` to array

baijunyao:
The default value of `filters` is `null`, An error is thrown when executing the `array_merge` function.

https://github.com/stripe/stripe-php/blob/e4da443d8f0df7d7a3a07b77edcaa6292d050b4b/lib/Collection.php#L180-L184

### DIFF
--- a/lib/Collection.php
+++ b/lib/Collection.php
@@ -45,7 +45,7 @@ class Collection extends StripeObject implements \IteratorAggregate
      */
     public function setFilters($filters)
     {
-        $this->filters = $filters;
+        $this->filters = (array) $filters;
         unset($this->filters['starting_after']);
         unset($this->filters['ending_before']);
     }


ob-stripe:
Hi @baijunyao, thanks for the contribution.

`filters` should never be `null` -- it's initialized with an empty array, not `null`: https://github.com/stripe/stripe-php/blob/4b245d2ed1007f56c8babdb1dc396a6582caa4e0/lib/Collection.php#L22

That said, I just modified this class' code (for unrelated reasons) in #797 and I added a safeguard so that the `array_merge` calls will succeed even if `filters` is `null` for some reason: https://github.com/stripe/stripe-php/blob/4b245d2ed1007f56c8babdb1dc396a6582caa4e0/lib/Collection.php#L195-L196.

Closing this, but feel free to reply if you believe there's still a bug in stripe-php!

baijunyao:
Sorry, I didn't say it clearly. I mean that `filters` in the parameter is `null` by default.  
https://github.com/stripe/stripe-php/blob/e4da443d8f0df7d7a3a07b77edcaa6292d050b4b/lib/Collection.php#L66
The following code will report an error `array_merge() expects parameter 1 to be array, null given`.
```php
    foreach (\Stripe\Account::all()->autoPagingIterator() as $payout) {
        var_dump($payout);
    }
```
The `$this->filters ?: []`  you added two hours ago solved this issue.
But `array_key_exists` can't accept `null` values either.
So I opened a new PR #798 to fix the issue.

PR #798, opened 2019-11-07
The argument of `array_key_exists` cannot be `null`

baijunyao:
#796

### DIFF
--- a/lib/Collection.php
+++ b/lib/Collection.php
@@ -132,8 +132,9 @@ class Collection extends StripeObject implements \IteratorAggregate
         $page = $this;
 
         while (true) {
-            if (array_key_exists('ending_before', $this->filters) &&
-                !array_key_exists('starting_after', $this->filters)) {
+            $filters = $this->filters ?: [];
+            if (array_key_exists('ending_before', $filters) &&
+                !array_key_exists('starting_after', $filters)) {
                 foreach ($page->getReverseIterator() as $item) {
                     yield $item;
                 }


baijunyao:
1. For example, I call the `all()` method of `Account`
```
    foreach (\Stripe\Account::all()->autoPagingIterator() as $payout) {
        var_dump($payout);
    }
```
2. lib/ApiOperations/All.php  
The `all()` method calls the `setFilters()` method

```php
    public static function all($params = null, $opts = null)
    {
        self::_validateParams($params);
        $url = static::classUrl();

        list($response, $opts) = static::_staticRequest('get', $url, $params, $opts);
        $obj = \Stripe\Util\Util::convertToStripeObject($response->json, $opts);
        if (!($obj instanceof \Stripe\Collection)) {
            throw new \Stripe\Exception\UnexpectedValueException(
                'Expected type ' . \Stripe\Collection::class . ', got "' . get_class($obj) . '" instead.'
            );
        }
        $obj->setLastResponse($response);
        $obj->setFilters($params);
        return $obj;
    }
```
- `$params = null`.
- ` $obj->setFilters($params);`
3. lib/Collection.php
```php
    public function setFilters($filters)
    {
        $this->filters = $filters;
    }
```
`$this->filters = null`

ob-stripe:
Ah, I see. Thanks for the explanation (and the patch!).

baijunyao:
So, I think #796 may need to be merged, this completely solves the problem that `filters` maybe `null`.

PR #802, opened 2019-11-07
Add continuous integration workflow with Github Actions

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Adds a continuous integration workflow using Github Actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,69 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  php-cs-fixer:
+    name: PHP-CS-Fixer
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@master
+
+      - name: PHP-CS-Fixer
+        uses: docker://oskarstark/php-cs-fixer-ga
+        with:
+          args: --diff-format=udiff --dry-run --show-progress=dots --using-cache=no --verbose .
+
+  phpstan:
+    name: PHPStan
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@master
+
+      - name: PHPStan
+        uses: docker://oskarstark/phpstan-ga
+        with:
+          args: analyse -c phpstan.neon lib tests
+
+  tests:
+    name: Tests
+
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: composer install --prefer-dist --no-progress --no-suggest
+      - name: Start stripe-mock
+        run: docker run -d -p 12111-12112:12111-12112 stripemock/stripe-mock && sleep 5
+      - name: Run test suite
+        run: ./build.php $env:AUTOLOAD
+      - name: Send code coverage report to coveralls.io
+        run: vendor/bin/php-coveralls -c .coveralls.github-actions.yml -v
+        env:
+          COVERALLS_RUN_LOCALLY: 1
+          COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
```
brandur-stripe: This is one aspect I like compared to Travis: env is injected more discriminately so it's more obvious what variables are supposed to apply to which steps.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,55 @@
...
+          - AUTOLOAD=0
+          - AUTOLOAD=1
+        php-version:
+          - "5.6"
+          - "7.0"
+          - "7.1"
+          - "7.2"
+          - "7.3"
+    steps:
+      - uses: actions/checkout@master
```
brandur-stripe: Super minor, but if you're down: I found the spacing between steps that you had in the prototype made this list easier to read.

ob-stripe:
> One other thing worth mentioning: the N * M build matrix of AUTOLOAD * version does create a huge list of checkboxes on pull requests, as you probably noticed. Not sure if there's anything we can or want to do about this, but just a note that it does seem slightly unwieldy compared to what we had before.

Yeah, agreed. I just tried not setting the job title based on the matrix variables, but it looks like Github Actions is too smart and outputs the variables anyway.

Not the end of the world, and the issue probably won't be as bad for other languages because the `AUTOLOAD` thing is specific to PHP.

brandur-stripe:
> Not the end of the world, and the issue probably won't be as bad for other languages because the `AUTOLOAD` thing is specific to PHP.

Yep, true.

PR #806, opened 2019-11-15
Automated native_function_invocation fixes

draco2003:
Reproduce with:
php php-cs-fixer --rules=native_function_invocation fix ./ --allow-risky=yes
(php-cs-fixer from : https://cs.symfony.com/)

Details:
https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/


ob-stripe:
Hi @draco2003, thanks for the contribution! We're actually already using PHP CS Fixer to format the library's code, but I didn't know about this rule.

I'm generally in favor of this change, but it looks like one of the tests is no longer passing. Mind taking a closer look? Otherwise I'm happy to do so myself when I have a chance.

draco2003:
Thanks for the quick response. 
Let me take a look and see If i can fix the test failure issue. 
Other Repos with this similar PR failed due to particular mocking of builtin functions.

draco2003:
Quick question based on other libraries feedback. 
Would you perfer the \is_array syntax or the:
use function is_array;

?

ob-stripe:
> Quick question based on other libraries feedback.
> Would you perfer the \is_array syntax or the:
> use function is_array;
> 
> ?

I'd prefer `\is_array`.

ob-stripe:
Closed in favor of #858.

Issue #793, opened 2019-11-05
PaymentMethod $card should be Card instead of `mixed`


ruudk:
https://github.com/stripe/stripe-php/blob/b6436760920e11bd4f83611a9f338ddb31e6adf5/lib/PaymentMethod.php#L11

Shouldn't this be `null|\Stripe\Card`?

remi-stripe:
@ruudk No, this is `mixed` because it's a hash (or an array) here. The `Card` object is entirely separate in the API.

ruudk:
Hmm, oke. I type hinted it with `Card` and then called `$paymentMethod->card->fingerprint` and it worked. 

So I checked, and that means that the `mixed` type should be `null|\Stripe\StripeObject`?

remi-stripe:
Today, we refer to array/hashes as `mixed` instead. I guess it could be `StripeObject` but that's not how we reference those.

It really is not a Stripe resource like Card or PaymentMethod. It's just a hash with some properties. It has `fingerprint` because it describes a card, not because it's the same as the Card object.

ruudk:
I understand and agree. I was wrong assuming it was a Card. 

But I think we should help developers with a bit more type safety / editor support so that they know what kind of value is inside those variables. Mixed means: find it out yourself.

remi-stripe:
@ruudk That's fair. We're working on auto-generating the library, which is how we got the `null` in yesterday so that will get better in the near future. cc @ob-stripe

ob-stripe:
Yep, this is already on my radar, but thanks for pointing it out anyway. We're still working on automatically generating the PHPDoc type annotations, but when we're done we should be able to replace all instances of `mixed` with `\Stripe\StripeObject` (or a more precise subclass, when possible).

ruudk:
Another thing that I see, is that the Stripe SDK doesn't really use PHP's type hints. Might be good to make things stricter when possible. But first, I think you should say goodbye to PHP 5.6 ;) https://www.php.net/supported-versions.php

ob-stripe:
Yes, we definitely plan on using PHP type hints once we drop support for 5.6. We don't have an ETA for that yet -- a large part of our user base is still using that version. We're closely monitoring usage numbers and will drop support once we feel confident that we're not preventing too many users from upgrading.

PR #809, opened 2019-11-21
Diffs from codegen

rattrayalex-stripe:
r? @ob-stripe 
cc @marko-stripe @remi-stripe 
cc @stripe/api-libraries 

The biggest difference I can find is that we no longer output `|null` for optional fields in comments. Should we? I don't think it should be too hard.

There are a few other comment regressions, like
```diff
- * @property \Stripe\Collection $reversals
+ * @property mixed $reversals
```
Feels on how important those are to fix? Should be doable.

### DIFF
--- a/lib/Account.php
+++ b/lib/Account.php
@@ -7,8 +7,8 @@ namespace Stripe;
  *
  * @property string $id
  * @property string $object
- * @property mixed|null $business_profile
- * @property string|null $business_type
+ * @property mixed $business_profile
+ * @property string $business_type
  * @property mixed $capabilities
  * @property bool $charges_enabled
  * @property mixed $company
@@ -16,13 +16,13 @@ namespace Stripe;
  * @property int $created
  * @property string $default_currency
  * @property bool $details_submitted
- * @property string|null $email
- * @property \Stripe\Collection $external_accounts
+ * @property string $email
+ * @property mixed $external_accounts
  * @property mixed $individual
  * @property \Stripe\StripeObject $metadata
  * @property bool $payouts_enabled
  * @property mixed $requirements
- * @property mixed|null $settings
+ * @property mixed $settings
  * @property mixed $tos_acceptance
  * @property string $type
  *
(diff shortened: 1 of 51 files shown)


marko-stripe:
There still is a fix for `|null` on my computer (@ob-stripe applied that before somehow already) and I also was investigating better types. It's likely something we can get from autogen from `stripe-java`.

rattrayalex-stripe:
Cool, then a patch would likely be appreciated but I can probably take!

lib/Account.php
```diff
@@ -150,13 +178,32 @@ public function deauthorize($clientId = null, $opts = null)
...
+     * @return Account The rejected account.
+     */
+    public function reject($params = null, $opts = null)
+    {
+        $url = $this->instanceUrl() . '/reject';
+        list($response, $opts) = $this->_request('post', $url, $params, $opts);
+        $this->refreshFrom($response, $opts);
+        return $this;
+    }
+    use ApiOperations\NestedResource;
```
remi-stripe: Minor but I'd recommend having this at the top. It's really easy to miss and it's an important concept core to those classes

lib/Account.php
```diff
@@ -201,6 +256,8 @@ public static function allCapabilities($id, $params = null, $opts = null)
         return self::_allNestedResources($id, static::PATH_CAPABILITIES, $params, $opts);
     }
 
+    const PATH_EXTERNAL_ACCOUNTS = '/external_accounts';
```
remi-stripe: This was useful when we were writing code manually. I wonder if it isn't cleaner/easier to just have the route in the function now.

ob-stripe:
> Feels on how important those are to fix?

I think we should definitely fix this. See e.g. this issue: #793. Ideally, we would replace all instances of `mixed` with the actual type:
- `\Stripe\<Resource>` for actual API resources (e.g. `\Stripe\Plan` for `plan` objects, etc.)
- `\Stripe\Collection` for list objects (unfortunately AFAIK PHPDoc has no support for "generics", so there's no way to do e.g. `\Stripe\Collection<\Stripe\Refund>` or anything like that)
- `\Stripe\StripeObject` for nested hashes that are not API resources
- `array` for actual lists

We don't have to fix everything at once, but we should at least not make the current state worse.

lib/Account.php
rattrayalex-stripe: Stupid PHP question: would removing those constants be a backwards-breaking change? Not sure if they are public or not.

rattrayalex-stripe:
Awesome, thanks for the feedback! I think those should all be pretty feasible; hopefully we'll be able to get through them next week.

ob-stripe:
Closing in favor of #837.

PR #816, opened 2019-12-04
Disable autoloader when checking for `Throwable`

### LINKED ISSUE #813: Stripe searches Throwable.php in the spl_autoload_register directory and gives a non-blocking error
Hi,
Hello, I am experiencing a non-blocking error when including Stripe.php -> init.php.

I have a controller that loads php classes from my API, they are all root Api.com/PHP/.

To avoid any conflict and since I had the following error: PHP Warning: include (PHP / Throwable.php): failed to open stream: No such file or directory
- When including Stripe.php in my code: 

include './STRIPE/init.php';
\ Stripe \ Stripe :: setApiKey (pay :: SK_PAY);

So I placed all Stripe.php in a separate folder it is in Api.com/STRIPE/

but the error continues.

I had to make a hack to no longer have error messages in my error_log: I created an empty Throwable.php file in the root that is loaded by the autoloader -> / PHP / and there more mistake ...

This is not a very good way to counter the error, but I did not find any Throwable.php files in Stripe.php.

Aside from that, the error that is thrown is non-blocking because the API is working fine and the test payments are accepted. I wanted to point it out here to understand
1 / Is this a bug?
2 / Is it me the bug?

Thank you for your answers,
Regards,
Raf

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

The [`\Throwable`](https://www.php.net/manual/en/class.throwable.php) interface was added in PHP 7, so we use `interface_exists` to check that it exists. However by default this will attempt to autoload the interface from a `Throwable.php` file, which can throw a warning in certain environments.

In this case the interface should be provided by PHP itself, so it's safe to disable autoloading by passing `false` as a second argument to [`interface_exists`](https://www.php.net/manual/en/function.interface-exists.php).

Fixes #813.

### DIFF
--- a/lib/Exception/ExceptionInterface.php
+++ b/lib/Exception/ExceptionInterface.php
@@ -3,7 +3,7 @@
 namespace Stripe\Exception;
 
 // TODO: remove this check once we drop support for PHP 5
-if (interface_exists(\Throwable::class)) {
+if (interface_exists(\Throwable::class, false)) {
     /**
      * The base interface for all Stripe exceptions.
      *


ob-stripe:
Released as 7.14.2.

PR #820, opened 2019-12-23
Update `Account` PHPDoc

baijunyao:
image

image

### DIFF
--- a/lib/Account.php
+++ b/lib/Account.php
@@ -7,23 +7,23 @@ namespace Stripe;
  *
  * @property string $id
  * @property string $object
- * @property mixed|null $business_profile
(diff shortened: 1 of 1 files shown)


ob-stripe:
Hi @baijunyao, thanks for the contribution!

We're not going to merge this immediately: there are a lot of `mixed` types in stripe-php's docblocks, and I'd prefer that we fix all of them rather than just some.

The good news is that we're actively working on automatically generating the library's source code from the OpenAPI specification file (https://github.com/stripe/openapi). We don't have a firm ETA yet, but it should happen early in 2020.

Thanks for your understanding! Closing, but feel free to reply.

baijunyao:
OK, thanks.

PR #821, opened 2019-12-27
Add `array<string, mixed>` as return type for `getJsonBody`

ruudk:
This makes it easier for PHPStan et all to understand that the array will be mixed.

### DIFF
--- a/lib/Exception/ApiErrorException.php
+++ b/lib/Exception/ApiErrorException.php
@@ -135,7 +135,7 @@ abstract class ApiErrorException extends \Exception implements ExceptionInterfac
     /**
      * Gets the JSON deserialized body.
      *
-     * @return array|null
+     * @return array<string, mixed>|null
      */
     public function getJsonBody()
     {
@@ -145,7 +145,7 @@ abstract class ApiErrorException extends \Exception implements ExceptionInterfac
     /**
      * Sets the JSON deserialized body.
      *
-     * @param array|null $jsonBody
+     * @param array<string, mixed>|null $jsonBody
      */
     public function setJsonBody($jsonBody)
     {


remi-stripe:
@ruudk Thanks for doing the work. I'm not sure if we should change this one so I'm going to defer to @ob-stripe who'll get back to you in a few days!

ob-stripe:
Could you clarify what this helps with exactly?

I was unable to find what `array<T>` implies for the key type (as opposed to the explicit `array<TKey, TValue>` notation). Does PHPStan assume that the array is indexed or associative if no type is specified for the keys? The former would be incorrect in this case, and if it's the latter I'm not sure this change is needed as surely `array` with no other information is the same thing as `array<mixed, mixed>` for PHPStan?

ruudk:
See https://phpstan.org/r/466e6dbe-7e8d-4fae-bfce-6c65b9e69df0

PHPStan does not understand what is inside `array` and wants to know. 

Either by changing it to `array<mixed>` (https://phpstan.org/r/81754f98-ec7f-440d-a126-00a2bea6d019) or `array<mixed, mixed>` (https://phpstan.org/r/d51cb42e-7f44-44c0-9a3d-86e52b807039) it's solved.

ob-stripe:
Interesting! Thanks for sharing.

Note that this warning only occurs at level 6 or higher. Currently we use [level 1](https://github.com/stripe/stripe-php/blob/7f6e7edfc8a2641da714ce122efbb730a22b4ddb/phpstan.neon#L5) in our CI pipeline and we'd have many other things to fix before we can get to level 6 (just gave it a try, there are 457 errors at the moment).

Anyway, I'm not opposed to this change, but would you mind making a couple of updates?
1. changing the type to `array<string, mixed>`
2. updating the on the `setJsonBody()` method as well

ruudk:
@ob-stripe We run at level 8 🤓 Applied your feedback.

ob-stripe:
> We run at level 8 🤓

I'm jealous 😛 

Thanks for the PR! Will release this shortly.

ob-stripe:
Released as 7.17.0.

PR #825, opened 2020-01-07
Update PHPDoc

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries

### DIFF
--- a/lib/BalanceTransaction.php
+++ b/lib/BalanceTransaction.php
@@ -16,6 +16,7 @@ namespace Stripe;
  * @property int $fee
  * @property mixed $fee_details
  * @property int $net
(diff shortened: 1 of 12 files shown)


lib/InvoiceItem.php
```diff
@@ -23,6 +23,7 @@
  * @property string|null $subscription
  * @property string $subscription_item
  * @property array|null $tax_rates
+ * @property bool $unified_proration
```
remi-stripe: That one shouldn't be here, ugh. If you can, I'd rather you remove it (and I flagged internally that it should be hidden)

ob-stripe:
Released as 7.16.0.

PR #826, opened 2020-01-07
Rename remaining `$options` to `$opts`

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Rename a few remaining `$options` arguments to `$opts` to be consistent with other methods.


ob-stripe:
Released as 7.16.0.

Issue #828, opened 2020-01-09
php_uname is disabled on some hosting providers


mihaijoldis:
On some hosting providers the php_uname function is disabled and i've seen some cases with integrations where the notices caused by that to interrupt a payment process.

richardm-stripe:
Thanks for the report @mihaijoldis! Addressing in #829.

PR #829, opened 2020-01-09
Don't call php_uname function if disabled by php.ini

richardm-stripe:
Addresses #828. Some hosting providers disable the php_uname function, presumably by adding it to 'disable_functions' in php.ini. This PR retrieves the value of 'disable_functions', parses the comma-separated list it to see if 'php_uname' is contained. In that case, we hardcode uname to "(disabled)" to avoid calling php_uname.

r? @ob-stripe 
cc @remi-stripe

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -260,7 +260,8 @@ class ApiRequestor
         $uaString = 'Stripe/v1 PhpBindings/' . Stripe::VERSION;
 
         $langVersion = phpversion();
-        $uname = php_uname();
+        $uname_disabled = in_array('php_uname', explode(',', ini_get('disable_functions')));
+        $uname = $uname_disabled ? '(disabled)' : php_uname();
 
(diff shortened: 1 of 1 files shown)


lib/ApiRequestor.php
```diff
@@ -260,7 +260,8 @@ private static function _defaultHeaders($apiKey, $clientInfo = null)
         $uaString = 'Stripe/v1 PhpBindings/' . Stripe::VERSION;
 
         $langVersion = phpversion();
-        $uname = php_uname();
+        $uname_disabled = in_array('php_uname', explode(',', ini_get('disable_functions')));
+        $uname = $uname_disabled ? '(disabled)' : php_uname();
```
remi-stripe: Minor but have we considered making that call once when the library or client is initialized and then use a cached value instead of doing that on each request?

ob-stripe: I'll probably do just that if/when I ever get to rewriting the HTTP client layer.

PR #831, opened 2020-01-14
Fix UnexpectedValueException

skurty:
Here is the error:

Fatal error:  Uncaught Error: Wrong parameters for Stripe\Exception\UnexpectedValueException([string $message [, long $code [, Throwable $previous = NULL]]]) in vendor/stripe/stripe-php/lib/ApiRequestor.php:424


ob-stripe:
Released as 7.19.1.

PR #837, opened 2020-01-23
PHPDoc fixes, CreditNote.previewLines, and autogen prep

richardm-stripe:
r? @ob
cc @stripe/api-libraries @remi-stripe @cjavilla-stripe @marko-stripe 
Changes:
  * Document missing properties on `Session`, `CreditNote`, `Invoice`, `InvoiceItem`, `Subscription`, and `SubscriptionSchedule`.
  * Undocument properties on `SubscriptionSchedule` that were removed in the latest API versions.
  * Add '.previewLines' method to CreditNote.
  * Adds more specific types for many docstrings. A full list of the PHPDoc changes is [here](https://gist.github.com/richardm-stripe/2c335c99b08f819b66e9c649695c9b8c).
  * Several formatting changes/reordering of methods. This is an artifact of autogeneration.

### DIFF
--- a/lib/Account.php
+++ b/lib/Account.php
@@ -36,10 +36,11 @@ class Account extends ApiResource
     use ApiOperations\Create;
     use ApiOperations\Delete;
     use ApiOperations\NestedResource;
+    use ApiOperations\Update;
+
     use ApiOperations\Retrieve {
         retrieve as protected _retrieve;
     }
-    use ApiOperations\Update;
 
     /**
      * Possible string representations of an account's business type.
@@ -85,11 +86,6 @@ class Account extends ApiResource
         return $savedNestedResources;
     }
 
-    const PATH_CAPABILITIES = '/capabilities';
-    const PATH_EXTERNAL_ACCOUNTS = '/external_accounts';
-    const PATH_LOGIN_LINKS = '/login_links';
-    const PATH_PERSONS = '/persons';
-
     public function instanceUrl()
     {
         if ($this['id'] === null) {
@@ -97,6 +93,55 @@ class Account extends ApiResource
         } else {
             return parent::instanceUrl();
         }
+    }
+
+    public function serializeParameters($force = false)
+    {
+        $update = parent::serializeParameters($force);
+        if (isset($this->_values['legal_entity'])) {
+            $entity = $this['legal_entity'];
+            if (isset($entity->_values['additional_owners'])) {
+                $owners = $entity['additional_owners'];
+                $entityUpdate = isset($update['legal_entity']) ? $update['legal_entity'] : [];
+                $entityUpdate['additional_owners'] = $this->serializeAdditionalOwners($entity, $owners);
+                $update['legal_entity'] = $entityUpdate;
+            }
+        }
+        if (isset($this->_values['individual'])) {
+            $individual = $this['individual'];
+            if (($individual instanceof Person) && !isset($update['individual'])) {
+                $update['individual'] = $individual->serializeParameters($force);
+            }
+        }
+        return $update;
+    }
+
+    private function serializeAdditionalOwners($legalEntity, $additionalOwners)
+    {
+        if (isset($legalEntity->_originalValues['additional_owners'])) {
+            $originalValue = $legalEntity->_originalValues['additional_owners'];
+        } else {
+            $originalValue = [];
+        }
+        if (($originalValue) && (count($originalValue) > count($additionalOwners))) {
+            throw new Exception\InvalidArgumentException(
+                "You cannot delete an item from an array, you must instead set a new array"
+            );
+        }
+
+        $updateArr = [];
+        foreach ($additionalOwners as $i => $v) {
+            $update = ($v instanceof StripeObject) ? $v->serializeParameters() : $v;
+
+            if ($update !== []) {
+                if (!$originalValue ||
+                    !array_key_exists($i, $originalValue) ||
+                    ($update != $legalEntity->serializeParamsValue($originalValue[$i], null, false, true))) {
+                    $updateArr[$i] = $update;
+                }
+            }
+        }
+        return $updateArr;
     }
 
     /**
@@ -118,22 +163,6 @@ class Account extends ApiResource
     }
 
     /**
-     * @param array|null $params
-     * @param array|string|null $opts
-     *
-     * @throws \Stripe\Exception\ApiErrorException if the request fails
-     *
-     * @return Account The rejected account.
-     */
-    public function reject($params = null, $opts = null)
-    {
-        $url = $this->instanceUrl() . '/reject';
-        list($response, $opts) = $this->_request('post', $url, $params, $opts);
-        $this->refreshFrom($response, $opts);
-        return $this;
-    }
-
-    /**
      * @param array|null $clientId
      * @param array|string|null $opts
      *
@@ -150,11 +179,30 @@ class Account extends ApiResource
         return OAuth::deauthorize($params, $opts);
     }
 
+    /**
+     * @param array|null $params
+     * @param array|string|null $opts
+     *
+     * @throws \Stripe\Exception\ApiErrorException if the request fails
+     *
+     * @return Account The rejected account.
+     */
+    public function reject($params = null, $opts = null)
+    {
+        $url = $this->instanceUrl() . '/reject';
+        list($response, $opts) = $this->_request('post', $url, $params, $opts);
+        $this->refreshFrom($response, $opts);
+        return $this;
+    }
+    
     /*
      * Capabilities methods
      * We can not add the capabilities() method today as the Account object already has a
      * capabilities property which is a hash and not the sub-list of capabilities.
      */
+
+
+    const PATH_CAPABILITIES = '/capabilities';
 
     /**
      * @param string $id The ID of the account on which to retrieve the capabilities.
@@ -199,6 +247,8 @@ class Account extends ApiResource
     {
         return self::_updateNestedResource($id, static::PATH_CAPABILITIES, $capabilityId, $params, $opts);
     }
+
+    const PATH_EXTERNAL_ACCOUNTS = '/external_accounts';
 
     /**
      * @param string $id The ID of the account on which to retrieve the external accounts.
@@ -273,6 +323,8 @@ class Account extends ApiResource
         return self::_updateNestedResource($id, static::PATH_EXTERNAL_ACCOUNTS, $externalAccountId, $params, $opts);
     }
 
+    const PATH_LOGIN_LINKS = '/login_links';
+
     /**
      * @param string $id The ID of the account on which to create the login link.
      * @param array|null $params
@@ -286,6 +338,8 @@ class Account extends ApiResource
     {
         return self::_createNestedResource($id, static::PATH_LOGIN_LINKS, $params, $opts);
     }
+
+    const PATH_PERSONS = '/persons';
 
     /**
      * @param array|null $params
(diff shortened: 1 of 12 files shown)


lib/EphemeralKey.php
```diff
@@ -11,18 +11,18 @@
  * @property int $expires
  * @property bool $livemode
  * @property string $secret
- * @property array $associated_objects
```
richardm-stripe: Override so this doesn't disappear

lib/Invoice.php
```diff
@@ -55,14 +55,15 @@
...
  * @property string|null $statement_descriptor
  * @property string|null $status
- * @property mixed $status_transitions
+ * @property \Stripe\StripeObject $status_transitions
  * @property string|null $subscription
  * @property int $subscription_proration_date
  * @property int $subtotal
  * @property int|null $tax
- * @property mixed $threshold_reason
+ * @property float|null $tax_percent
```
richardm-stripe: This is interesting

lib/InvoiceItem.php
```diff
@@ -16,13 +16,14 @@
...
- * @property mixed $period
+ * @property \Stripe\StripeObject $period
  * @property \Stripe\Plan|null $plan
  * @property bool $proration
  * @property int $quantity
  * @property string|null $subscription
  * @property string $subscription_item
- * @property array|null $tax_rates
+ * @property \Stripe\TaxRate[]|null $tax_rates
+ * @property bool $unified_proration
```
richardm-stripe: this is interesting

lib/Source.php
```diff
@@ -7,42 +7,42 @@
...
  * @property string $flow
- * @property mixed $giropay
- * @property mixed $ideal
- * @property mixed $klarna
+ * @property \Stripe\StripeObject $giropay
+ * @property \Stripe\StripeObject $ideal
+ * @property \Stripe\StripeObject $klarna
  * @property bool $livemode
  * @property \Stripe\StripeObject|null $metadata
- * @property mixed $multibanco
```
richardm-stripe: gonna need to do overrides

lib/Mandate.php
```diff
@@ -7,12 +7,12 @@
  *
  * @property string $id
  * @property string $object
- * @property mixed $customer_acceptance
+ * @property \Stripe\StripeObject $customer_acceptance
  * @property bool $livemode
- * @property mixed|null $multi_use
```
richardm-stripe: source-style polymorphism

lib/PaymentMethod.php
```diff
@@ -7,16 +7,16 @@
  *
  * @property string $id
  * @property string $object
- * @property mixed|null $au_becs_debit
```
richardm-stripe: source-style polymorphism :(

lib/SubscriptionSchedule.php
```diff
@@ -7,22 +7,19 @@
  *
  * @property string $id
  * @property string $object
- * @property mixed|null $billing_thresholds
```
richardm-stripe: this is due to spec change

lib/Account.php
```diff
@@ -305,6 +358,7 @@ public function persons($params = null, $opts = null)
         return $obj;
     }
 
+
```
remi-stripe: extra new line

lib/Account.php
```diff
@@ -377,53 +431,4 @@ public static function allPersons($id, $params = null, $opts = null)
...
-            if ($update !== []) {
-                if (!$originalValue ||
-                    !array_key_exists($i, $originalValue) ||
-                    ($update != $legalEntity->serializeParamsValue($originalValue[$i], null, false, true))) {
-                    $updateArr[$i] = $update;
-                }
-            }
-        }
-        return $updateArr;
-    }
```
remi-stripe: if we could easily put the override like that at the end of the file it'd be great

lib/ApplicationFee.php
```diff
@@ -30,6 +30,7 @@ class ApplicationFee extends ApiResource
     use ApiOperations\NestedResource;
     use ApiOperations\Retrieve;
 
+
```
remi-stripe: extra new line (won't flag others)

lib/CreditNote.php
```diff
@@ -78,6 +82,22 @@ public static function preview($params = null, $opts = null)
...
+     * @param array|null $params
+     * @param array|string|null $opts
+     *
+     * @throws \Stripe\Exception\ApiErrorException if the request fails
+     *
+     * @return CreditNote The previewed credit note.
+     */
+    public function previewLines($params = null, $opts = null)
+    {
+        $url = $this->instanceUrl() . '/lines';
```
remi-stripe: [blocker]

This is incorrect because it should be static and it doesn't have an instanceUrl because it's not on an existing credit note. Your think works if you have called preview first but IMO this is the wrong abstraction/name

We should blacklist this method for now instead of trying to fix it

lib/Customer.php
```diff
@@ -78,192 +74,199 @@ public function deleteDiscount($params = null, $opts = null)
...
     /**
-     * @param string $id The ID of the customer on which to create the source.
+     * @param string $id The ID of the customer on which to create the customer balance transaction.
      * @param array|null $params
      * @param array|string|null $opts
      *
      * @throws \Stripe\Exception\ApiErrorException if the request fails
      *
-     * @return ApiResource
+     * @return BalanceTransaction
```
remi-stripe: [blocker] this should be CustomerBalanceTransaction right?
Also this makes me realize that `TaxId` the resource should really be `CustomerTaxId` instead (todo for later)

lib/Customer.php
```diff
@@ -78,192 +74,199 @@ public function deleteDiscount($params = null, $opts = null)
...
-     * @param string $id The ID of the customer to which the source belongs.
-     * @param string $sourceId The ID of the source to delete.
+     * @param string $id The ID of the customer on which to retrieve the customer balance transactions.
      * @param array|null $params
      * @param array|string|null $opts
      *
      * @throws \Stripe\Exception\ApiErrorException if the request fails
      *
-     * @return ApiResource
+     * @return Collection The list of customer balance transactions.
```
remi-stripe: [blocker] This should be `\Stripe\Collection`

lib/Customer.php
```diff
@@ -78,192 +74,199 @@ public function deleteDiscount($params = null, $opts = null)
...
      * @param array|null $params
      * @param array|string|null $opts
      *
      * @throws \Stripe\Exception\ApiErrorException if the request fails
      *
-     * @return Collection The list of sources.
+     * @return AlipayAccount|BankAccount|BitcoinReceiver|Card|Source
      */
-    public static function allSources($id, $params = null, $opts = null)
+    public static function createSource($id, $params = null, $opts = null)
```
remi-stripe: It would be great if we enforced the order of those operations. Either we alphabetize entirely so all is first, or we keep what we had). I'm not strongly opposed but I worry that we rely on the order in the spec which could change in the future, making diffs impossible to review

lib/Invoice.php
remi-stripe: yeah it's legacy/deprecated, but still supported for the foreseeable future

lib/Invoice.php
```diff
@@ -76,6 +77,7 @@ class Invoice extends ApiResource
     use ApiOperations\Delete;
     use ApiOperations\Retrieve;
     use ApiOperations\Update;
+
     use ApiOperations\NestedResource;
```
remi-stripe: that one seems to not always be ordered the same way. Nitpicky :p

lib/Invoice.php
```diff
@@ -117,6 +119,37 @@ class Invoice extends ApiResource
...
+    {
+        $url = static::classUrl() . '/upcoming';
+        list($response, $opts) = static::_staticRequest('get', $url, $params, $opts);
+        $obj = Util\Util::convertToStripeObject($response->json, $opts);
+        $obj->setLastResponse($response);
+        return $obj;
+    }
+
+    /**
+     * @param string $id The ID of the invoice on which to retrieve the lins.
```
remi-stripe: typo `lins` => `lines`

lib/Invoice.php
```diff
@@ -132,7 +165,6 @@ public function finalizeInvoice($params = null, $opts = null)
         $this->refreshFrom($response, $opts);
         return $this;
     }
-
```
remi-stripe: lost empty line

lib/InvoiceItem.php
remi-stripe: Yeah that one is problematic because it was ungated by mistake, so it stays in the spec to avoid a major version.
In a perfect world, we'd blacklist this in php at the override level

lib/Mandate.php
remi-stripe: yeah discussed with ob that this could be problematic because that supports `hide_nil` and so it technically can be undefined. But we should fix in a future PR

lib/Order.php
```diff
@@ -46,14 +46,13 @@ class Order extends ApiResource
      *
      * @throws \Stripe\Exception\ApiErrorException if the request fails
      *
-     * @return Order The paid order.
+     * @return OrderReturn The newly created return.
```
remi-stripe: This should be `\Stripe\OrderReturn` likely similar to collection from earlier, won't flag all

lib/PaymentMethod.php
```diff
@@ -7,16 +7,16 @@
...
- * @property mixed $billing_details
- * @property mixed $card
- * @property mixed $card_present
+ * @property \Stripe\StripeObject $au_becs_debit
+ * @property \Stripe\StripeObject $billing_details
+ * @property \Stripe\Card $card
```
remi-stripe: AFAIK this is totally incorrect. a PaymentMethod's `card` is just a hash and not a Card object

lib/Review.php
```diff
@@ -12,13 +12,13 @@
  * @property string|null $closed_reason
  * @property int $created
  * @property string|null $ip_address
- * @property mixed|null $ip_address_location
+ * @property \Stripe\Terminal\Location|null $ip_address_location
```
remi-stripe: This is definitely incorrect

lib/Review.php
```diff
@@ -12,13 +12,13 @@
...
  * @property bool $livemode
  * @property bool $open
  * @property string $opened_reason
  * @property string $payment_intent
  * @property string $reason
- * @property mixed|null $session
+ * @property \Stripe\Checkout\Session|null $session
```
remi-stripe: Same as above, this is a hash not a checkout session

lib/SetupIntent.php
```diff
@@ -13,14 +13,14 @@
  * @property int $created
  * @property string|null $customer
  * @property string|null $description
- * @property mixed|null $last_setup_error
+ * @property \Stripe\StripeObject|null $last_setup_error
```
remi-stripe: Technically this is an `Error` object today not sure if we care

lib/Source.php
```diff
@@ -7,42 +7,42 @@
...
- * @property mixed $source_order
+ * @property \Stripe\StripeObject $multibanco
+ * @property \Stripe\StripeObject|null $owner
+ * @property \Stripe\StripeObject $p24
+ * @property \Stripe\StripeObject $receiver
+ * @property \Stripe\StripeObject $redirect
+ * @property \Stripe\StripeObject $sepa_credit_transfer
+ * @property \Stripe\StripeObject $sepa_debit
+ * @property \Stripe\StripeObject $sofort
+ * @property \Stripe\Order $source_order
```
remi-stripe: [blocker] This should not be an Order at all but a hash. the Order part should be gated and not user-facing see https://stripe.com/docs/api/sources/object#source_object-source_order

lib/Source.php
```diff
@@ -7,42 +7,42 @@
...
  * @property string|null $statement_descriptor
  * @property string $status
- * @property mixed $three_d_secure
+ * @property \Stripe\ThreeDSecure $three_d_secure
```
remi-stripe: [blocker] Not entirely sure here but I think this is wrong for similar reasons

lib/SubscriptionSchedule.php
remi-stripe: yep normal that one, API version recently

marko-stripe:
> Also flagging that I think we have the same problem as with other codegen-ed resources: we never touch the nested resources like `CustomerBalanceTransaction` and `TaxId`. I could be wrong, but if I'm correct, if you removed all resources files, then re-codegen you should see all the missing files. It's okay-ish if so for now and something we should fix across all codegen soon

I think that's right. Brian and Josh from Slack saw that issue when they were starting to generate `stripe-hack` based on `stripe-php`. They started to also include nested resources when generating the files and it looked like it worked well.

ob-stripe:
Rebased on latest master and regenerated. I think the remaining diff is as small as it's going to get.

Changes are:
- path constants for nested resources are moved next to the method definitions for nested resources
- some changes in whitespace and method definition order due to overrides (we don't have fine grained control over where code snippets manually defined via overrides will end up)

ob-stripe:
r? @cjavilla-stripe @remi-stripe @richardm-stripe 

Just in case I missed anything, but I think we're ready to make the switch to codegen!

richardm-stripe:
lgtm

cjavilla-stripe:
I think we're good to go. Hard to see the blind spots but excited to see this. Awesome work @richardm-stripe,  @ob-stripe, and @remi-stripe!!

ob-stripe:
Released as 7.23.0.

PR #841, opened 2020-01-29
More PHPDoc fixes

ob-stripe:
r? @cjavilla-stripe @remi-stripe 
cc @stripe/api-libraries 

More PHPDoc fixes from #837.

### DIFF
--- a/lib/PaymentIntent.php
+++ b/lib/PaymentIntent.php
@@ -23,23 +23,23 @@ namespace Stripe;
  * @property string|null $customer
  * @property string|null $description
  * @property string|null $invoice
- * @property mixed|null $last_payment_error
+ * @property \Stripe\StripeObject|null $last_payment_error
  * @property bool $livemode
  * @property \Stripe\StripeObject $metadata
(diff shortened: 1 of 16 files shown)


lib/PaymentIntent.php
```diff
@@ -23,23 +23,23 @@
  * @property string|null $customer
  * @property string|null $description
  * @property string|null $invoice
- * @property mixed|null $last_payment_error
+ * @property \Stripe\StripeObject|null $last_payment_error
```
ruudk: @ob-stripe Shouldn't this be an `\Stripe\ErrorObject` type? That's how we always use it.

ob-stripe: @ruudk It is. We're currently switching to a a new "codegen" tool that automatically generates PHPDoc type annotations based on the OpenAPI specification for Stripe's API. The tool didn't know about `\Stripe\ErrorObject`, but that will be fixed very soon (it's already been fixed internally, we just need to update the public repository). Thanks for the report!

PR #854, opened 2020-02-01
Upgrade PHPStan and fix settings

ob-stripe:
r? @remi-stripe 

A few changes to PHPStan:
- upgrade to latest version 0.12.8
- rename `phpstan.neon` to `phpstan.neon.dist` and add `phpstan.neon` to `.gitignore` (so one can play with custom settings locally)
- add `tests/bootstrap.php` to the `bootstrap` setting, otherwise the `MOCK_URL` constant is seen as missing
- ignore the `Unsafe usage of new static().` error. We're relying pretty heavily on this right now. We should be able to fix most of these once we stop relying on traits

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ clover.xml
 # Ignore PHP CS Fixer local config and cache
 .php_cs
 .php_cs.cache
+
+# Ignore PHPStan local config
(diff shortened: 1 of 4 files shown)


ob-stripe:
Released as 7.23.0.

PR #857, opened 2020-02-04
Update PHPDoc on nested resources

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Update PHPDoc on nested resources that are not being generated yet.

### DIFF
--- a/lib/Person.php
+++ b/lib/Person.php
@@ -10,28 +10,28 @@ namespace Stripe;
  * @property string $id
  * @property string $object
  * @property string $account
- * @property mixed $address
- * @property mixed $address_kana
- * @property mixed $address_kanji
+ * @property \Stripe\StripeObject $address
+ * @property \Stripe\StripeObject|null $address_kana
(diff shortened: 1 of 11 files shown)


lib/Person.php
```diff
@@ -10,12 +10,12 @@
...
- * @property mixed $address
- * @property mixed $address_kana
- * @property mixed $address_kanji
+ * @property \Stripe\StripeObject $address
+ * @property \Stripe\StripeObject|null $address_kana
+ * @property \Stripe\StripeObject|null $address_kanji
  * @property int $created
  * @property bool $deleted
- * @property mixed $dob
+ * @property \Stripe\StripeObject $dob
```
remi-stripe: can't this be null? IF not is it a weird hash with January 1 1970?

ob-stripe: Not according to the spec, but I think the spec may be wrong in this instance.

ob-stripe:
Released as 7.23.0.

PR #858, opened 2020-02-04
Use native_function_invocation rule

ob-stripe:
r? @brandur-stripe @richardm-stripe 
cc @stripe/api-libraries 

Enables the `native_function_invocation` rule.

Replaces #806.


.php_cs.dist
```diff
@@ -1,13 +1,17 @@
 <?php
 
 return PhpCsFixer\Config::create()
+    ->setRiskyAllowed(true)
     ->setRules([
+        '@PHP56Migration:risky' => true,
+        '@PHPUnit57Migration:risky' => true,
```
ob-stripe: I've already run the fixes for those two rules in #846 and #847. Adding them to the configuration file ensures any new code will also match those rules.

tests/Stripe/Util/DefaultLoggerTest.php
```diff
@@ -1,28 +1,25 @@
...
 namespace Stripe\Util;
 
 class DefaultLoggerTest extends \Stripe\TestCase
 {
     public function testDefaultLogger()
     {
-        $logger = new DefaultLogger();
-        $logger->error("message");
+        // DefaultLogger uses PHP's `error_log` function. In order to capture
+        // the output, we need to temporarily redirect it to a temporary file.
```
ob-stripe: I had to completely rewrite this test, as previously it was using a hack where we manually defined an `error_log()` function in the `\Stripe\Util` namespace and relied on the fact that `DefaultLogger` called `error_log()` in its non-qualified form.

The new test actually captures the output of PHP's `\error_log()` function.

tests/Stripe/StripeObjectTest.php
```diff
@@ -127,8 +127,21 @@ public function testToArrayRecursive()
...
+            $this->assertNull($s->nonexistent);
+
+            $this->assertRegExp(
+                "/Stripe Notice: Undefined property of Stripe\\\\StripeObject instance: nonexistent/",
+                \stream_get_contents($capture)
+            );
+        } finally {
+            \ini_set('error_log', $origErrorLog);
+            \fclose($capture);
+        }
```
ob-stripe: While running the test suite, I noticed that this test was printing "Stripe Notice: Undefined property of Stripe\StripeObject instance: nonexistent" to stdout. I used the same technique as `DefaultLoggerTest` to capture the output. This has two upsides: the test suite's output is clean, and we actually test the behavior that accessing a non-existent property outputs a notice to the logger.

ob-stripe:
Released as 7.23.0.

PR #865, opened 2020-02-07
Get closer to @PhpCsFixer standard ruleset

ob-stripe:
r? @richardm-stripe @cjavilla-stripe 
cc @stripe/api-libraries 

This PR enables a lot PHP-CS-Fixer rules. The goal is to get as close as possible to the `@PhpCsFixer` standard ruleset, which enforces fairly opinionated style rules.

There are still a few rules that are not enabled, or are enabled with a custom configuration, because they cause too many changes in codegen'd files and I'd rather we generate compliant code in the first place instead of relying too heavily on PHP-CS-Fixer to reformat everything.

Once all those issues are fixed, we should be able to remove the long list of rules from `.php_cs.dist` and just enable the `@PhpCsFixer` ruleset instead.

I recommend reviewing the PR commit by commit. There are lot of commits, but most of them simply add a rule to `.php_cs.dist` and don't do anything else because we were already compliant with that rule.

### DIFF
--- a/lib/HttpClient/CurlClient.php
+++ b/lib/HttpClient/CurlClient.php
@@ -204,7 +204,7 @@ class CurlClient implements ClientInterface
 
         $params = Util\Util::objectsToIds($params);
 
-        if ($method === 'get') {
+        if ('get' === $method) {
             if ($hasFile) {
                 throw new Exception\UnexpectedValueException(
                     "Issuing a GET request with a file parameter"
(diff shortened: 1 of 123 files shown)


lib/HttpClient/CurlClient.php
```diff
@@ -251,19 +251,19 @@ public function request($method, $absUrl, $headers, $params, $hasFile)
...
         if (!Stripe::getVerifySslCerts()) {
-            $opts[CURLOPT_SSL_VERIFYPEER] = false;
+            $opts[\CURLOPT_SSL_VERIFYPEER] = false;
         }
 
-        if (!isset($opts[CURLOPT_HTTP_VERSION]) && $this->getEnableHttp2()) {
+        if (!isset($opts[\CURLOPT_HTTP_VERSION]) && $this->getEnableHttp2()) {
             // For HTTPS requests, enable HTTP/2, if supported
-            $opts[CURLOPT_HTTP_VERSION] = CURL_HTTP_VERSION_2TLS;
+            $opts[\CURLOPT_HTTP_VERSION] = \CURL_HTTP_VERSION_2TLS;
```
ob-stripe: Check that this still works if the constant is manually defined via `define()`

ob-stripe: It does, `define` ignores the current namespace and defines the new variable into the root namespace.

ob-stripe:
Released as 7.24.0.

PR #872, opened 2020-02-12
Another batch of PHP-CS-Fixer rule updates

ob-stripe:

Enable more rules from the `@PhpCsFixer` ruleset.

I will open an internal PR so that the codegen output is consistent with the updates.

I will do a last pass over the remaining disabled rules, but I think most of them will have to stay disabled because they would break our code / require significant refactoring.


examples/oauth.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-require('../init.php');
+require '../init.php';
```
remi-stripe: whyyyyyyyyyyyyyyyy is this deemed better.

lib/Account.php
```diff
@@ -150,8 +150,8 @@ private function serializeAdditionalOwners($legalEntity, $additionalOwners)
     }
 
     /**
-     * @param null|array|string $id The ID of the account to retrieve, or an
-     *     options array containing an `id` key.
+     * @param null|array|string $id the ID of the account to retrieve, or an
```
remi-stripe: Surprised they don't like the leading upper case letter here, especially as the rule is about not ending with a dot
Nitpicky but not sure it looks nicer in IDEs to be honest

lib/Account.php
```diff
@@ -3,7 +3,7 @@
 namespace Stripe;
 
 /**
- * Class Account
+ * Class Account.
```
remi-stripe: and here they want a dot. whyyyyyyyyy

ob-stripe:
Thanks Remi.

> whyyyyyyyyyyyyyyyy is this deemed better.

Probably because `include`/`require` are statements (i.e. reserved keywords), not functions.

> Surprised they don't like the leading upper case letter here, especially as the rule is about not ending with a dot
> Nitpicky but not sure it looks nicer in IDEs to be honest
>
> and here they want a dot. whyyyyyyyyy

I think the convention of "summaries are capitalized and have a final dot, parameter and return value descriptions are not capitalized and don't have a final dot" is the same as Javadoc.

PR #875, opened 2020-02-13
Last set of PHP-CS-Fixer updates

ob-stripe:

Okay, this should be the last PR for PHP-CS-Fixer rules for a while. Thanks for bearing with me!

This PR enables a few more rules, and updates `.php_cs.dist` to document why we differ from the `@PhpCsFixer` ruleset.


tests/Stripe/StripeTelemetryTest.php
```diff
@@ -33,7 +33,8 @@ public function testNoTelemetrySentIfNotEnabled()
         $stub = $this
             ->getMockBuilder('HttpClient\\ClientInterface')
             ->setMethods(['request'])
-            ->getMock();
+            ->getMock()
+        ;
```
remi-stripe: seriously you want to make me mad right?

PR #876, opened 2020-02-14
Use codegen for enum constants

ob-stripe:
r? @remi-stripe 

WIP PR for automatically generating constants for enum values.

This PR:
- removes the PHPDoc block for constants. I don't think that's a huge issue since the existing PHPDoc only applies to the topmost constant anyway. At the moment we can't really generate better docs, but as we migrate enum fields to our new-style enums where each value can be documented separately, we'll likely be able to generate good PHPDoc for each value in the future.
- adds some missing values
- changes the order of a few things

There a few constants that are still generated using the existing override system:
- `DECLINED_*` on `Charge`, because the field they're describing is actually kinda removed from Charge
- `REASON_*` on `Review`, because these values should actually be split into `CLOSED_REASON_*` and `OPENED_REASON_*`. ~The new values are generated and I added a deprecation notice.~

### DIFF
--- a/lib/Invoice.php
+++ b/lib/Invoice.php
@@ -79,7 +79,6 @@ class Invoice extends ApiResource
     const BILLING_CHARGE_AUTOMATICALLY = 'charge_automatically';
     const BILLING_SEND_INVOICE = 'send_invoice';
 
-    const BILLING_REASON_AUTOMATIC_PENDING_INVOICE_ITEM_INVOICE = 'automatic_pending_invoice_item_invoice';
     const BILLING_REASON_MANUAL = 'manual';
     const BILLING_REASON_SUBSCRIPTION = 'subscription';
     const BILLING_REASON_SUBSCRIPTION_CREATE = 'subscription_create';
--- a/lib/Invoice.php
+++ b/lib/Invoice.php
@@ -76,13 +76,10 @@ class Invoice extends ApiResource
     use ApiOperations\Retrieve;
     use ApiOperations\Update;
 
-    use ApiOperations\NestedResource;
+    const BILLING_CHARGE_AUTOMATICALLY = 'charge_automatically';
+    const BILLING_SEND_INVOICE = 'send_invoice';
 
-    /**
-     * Possible string representations of the billing reason.
-     *
-     * @see https://stripe.com/docs/api/invoices/object#invoice_object-billing_reason
-     */
+    const BILLING_REASON_AUTOMATIC_PENDING_INVOICE_ITEM_INVOICE = 'automatic_pending_invoice_item_invoice';
     const BILLING_REASON_MANUAL = 'manual';
     const BILLING_REASON_SUBSCRIPTION = 'subscription';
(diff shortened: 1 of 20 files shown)


lib/Refund.php
```diff
@@ -32,29 +32,14 @@ class Refund extends ApiResource
     use ApiOperations\Retrieve;
     use ApiOperations\Update;
 
-    /**
-     * Possible string representations of the failure reason.
-     *
-     * @see https://stripe.com/docs/api/refunds/object#refund_object-failure_reason
-     */
-    const FAILURE_REASON = 'expired_or_canceled_card';
+    const FAILURE_REASON_EXPIRED_OR_CANCELED_CARD = 'expired_or_canceled_card';
```
ob-stripe: This one is technically a breaking change, but the old constant had a blatantly wrong name :/ `FAILURE_REASON` instead of `FAILURE_REASON_EXPIRED_OR_CANCELED_CARD`.

We can use an override to keep the old value around temporarily if needed.

lib/Invoice.php
```diff
@@ -76,13 +76,10 @@ class Invoice extends ApiResource
...
-    use ApiOperations\NestedResource;
+    const BILLING_CHARGE_AUTOMATICALLY = 'charge_automatically';
+    const BILLING_SEND_INVOICE = 'send_invoice';
 
-    /**
-     * Possible string representations of the billing reason.
-     *
-     * @see https://stripe.com/docs/api/invoices/object#invoice_object-billing_reason
-     */
+    const BILLING_REASON_AUTOMATIC_PENDING_INVOICE_ITEM_INVOICE = 'automatic_pending_invoice_item_invoice';
```
remi-stripe: I am not convinced I want that constant for now. Up to you, but I'd focus on just the core ones until we solve what to do for API versions in the future that remove values

lib/Refund.php
remi-stripe: Yeah I'd do an override to play it safe, even if the name was horrible. But I'm fine with risking it

lib/Review.php
```diff
@@ -27,12 +27,21 @@ class Review extends ApiResource
     use ApiOperations\All;
     use ApiOperations\Retrieve;
 
+    const CLOSED_REASON_APPROVED = 'approved';
```
remi-stripe: same as earlier, I wouldn't add those

lib/Review.php
```diff
@@ -27,12 +27,21 @@ class Review extends ApiResource
...
+    const CLOSED_REASON_DISPUTED = 'disputed';
+    const CLOSED_REASON_REFUNDED = 'refunded';
+    const CLOSED_REASON_REFUNDED_AS_FRAUD = 'refunded_as_fraud';
+
+    const OPENED_REASON_MANUAL = 'manual';
+    const OPENED_REASON_RULE = 'rule';
```
remi-stripe: same

ob-stripe:
ptal @remi-stripe

PR #879, opened 2020-02-14
Make `\Stripe\Collection` implement `\Countable`

### LINKED ISSUE #807: Wrong invoices count (always 4)
I have an issue with following:

\Stripe\Stripe::setApiKey($_ENV['STRIPE_KEY']);
$invoices = \Stripe\Invoice::all(['limit' => 100]);

$invoices->count() always returns 4 even if when I iterate over results it clearly shows more records. count($invoices->data) returns correct number of 100.

ob-stripe:
r? @brandur-stripe @richardm-stripe 
cc @stripe/api-libraries 

Make `\Stripe\Collection` implement `\Countable`, so that calling `count($page)` will return the number of objects in the page rather than the number of keys on the page object.

Fixes #807.

### DIFF
--- a/lib/Collection.php
+++ b/lib/Collection.php
@@ -10,7 +10,7 @@ namespace Stripe;
  * @property bool $has_more
  * @property \Stripe\StripeObject[] $data
  */
-class Collection extends StripeObject implements \IteratorAggregate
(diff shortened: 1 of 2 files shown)


ob-stripe:
Released as 7.25.0.

PR #888, opened 2020-02-27
Prepare for `strict_types`

ob-stripe:
cc @stripe/api-libraries 

Some changes in preparation of [strict typing](https://www.php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration.strict) once we drop support for PHP 5.x.

I'll add inline comments to explain the different changes.

### DIFF
--- a/lib/ApiResource.php
+++ b/lib/ApiResource.php
@@ -43,8 +43,6 @@ abstract class ApiResource extends StripeObject
             ($v instanceof ApiResource)) {
             $v->saveWithParent = true;
         }
-
-        return $v;
     }
 
     /**
--- a/lib/BankAccount.php
+++ b/lib/BankAccount.php
@@ -89,7 +89,7 @@ class BankAccount extends ApiResource
                "'bank_account_id')` or `Account::retrieveExternalAccount(" .
                "'account_id', 'bank_account_id')`.";
 
-        throw new Exception\BadMethodCallException($msg, null);
+        throw new Exception\BadMethodCallException($msg);
     }
 
     /**
@@ -107,7 +107,7 @@ class BankAccount extends ApiResource
                '$updateParams)` or `Account::updateExternalAccount(' .
                "'account_id', 'bank_account_id', \$updateParams)`.";
 
-        throw new Exception\BadMethodCallException($msg, null);
+        throw new Exception\BadMethodCallException($msg);
     }
 
     /**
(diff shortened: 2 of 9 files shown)


lib/ApiResource.php
```diff
@@ -43,8 +43,6 @@ public function __set($k, $v)
             ($v instanceof ApiResource)) {
             $v->saveWithParent = true;
         }
-
-        return $v;
```
ob-stripe: This was a bug: the `__set()` magic method should not return anything.

lib/BankAccount.php
```diff
@@ -89,7 +89,7 @@ public static function retrieve($_id, $_opts = null)
                "'bank_account_id')` or `Account::retrieveExternalAccount(" .
                "'account_id', 'bank_account_id')`.";
 
-        throw new Exception\BadMethodCallException($msg, null);
+        throw new Exception\BadMethodCallException($msg);
```
ob-stripe: The second argument to [`Exception.__construct`](https://www.php.net/manual/en/exception.construct.php) is optional, but when present should be an integer. `null` is not a valid value.

lib/Util/DefaultLogger.php
```diff
@@ -8,15 +8,22 @@
...
     {
         if (\count($context) > 0) {
             throw new \Stripe\Exception\BadMethodCallException('DefaultLogger does not currently implement context. Please implement if you need it.');
         }
-        \error_log($message, $this->messageType, $this->destination);
+
+        if (null === $this->destination) {
+            \error_log($message, $this->messageType);
+        } else {
+            \error_log($message, $this->messageType, $this->destination);
```
ob-stripe: Likewise, the third argument to [`error_log()`](https://www.php.net/manual/en/function.error-log.php) is optional, but should be a string when present.

lib/Util/Util.php
```diff
@@ -314,7 +314,7 @@ public static function flattenParamsList($value, $calculatedKey)
      */
     public static function urlEncode($key)
     {
-        $s = \urlencode($key);
+        $s = \urlencode((string) $key);
```
ob-stripe: Need to explicitly cast here because [`urlencode()`](https://www.php.net/manual/en/function.urlencode.php) expects strings.

tests/Stripe/CollectionTest.php
```diff
@@ -19,7 +19,7 @@ final class CollectionTest extends \PHPUnit\Framework\TestCase
     public function setUpFixture()
     {
         $this->fixture = Collection::constructFrom([
-            'data' => [['id' => 1]],
+            'data' => [['id' => '1']],
```
ob-stripe: Updated this test to use strings instead of integers for IDs, as Stripe resource IDs should always be strings.

PR #893, opened 2020-03-12
Fix multiple PHPDocs

remi-stripe:
Codegen for openapi e77fb69

r? @ob-stripe 
cc @stripe/api-libraries


lib/BankAccount.php
```diff
@@ -89,7 +89,7 @@ public static function retrieve($_id, $_opts = null)
                "'bank_account_id')` or `Account::retrieveExternalAccount(" .
                "'account_id', 'bank_account_id')`.";
 
-        throw new Exception\BadMethodCallException($msg);
+        throw new Exception\BadMethodCallException($msg, null);
```
remi-stripe: Is this normal?

ob-stripe: No, I need to fix the overrides.

Issue #896, opened 2020-03-18
Support PHP 7.4


justin-schroeder:
We’re running into issues getting the stripe SDK to work in PHP 7.4. The error produced is:

```
Trying to access array offset on value of type int /var/task/vendor/stripe/stripe-php/lib/StripeObject.php:433
```

This makes sense, because line 433 is using an array accessor in `StripeObject::toArray()` to determine if a property is public or private before serializing to an array:

```php
if ('_' === $k[0]) ...
```

However, in PHP 7.4 you should not use array notation to access the first character of a number like  in PHP 7.3.

```php
// In PHP 7.3
$x = 123;
echo $x[0]; // outputs NULL

// In PHP 7.4
$x = 123;
echo $x[0]; // throws Notice Trying to access array offset on value of type int in 
```

For environments that convert error/notices into exceptions (like Laravel) this causes an exception to be thrown.

ob-stripe:
Thanks for the bug report @justin-schroeder, I was able to reproduce it easily.

Out of curiosity, how did you discover it? Are you using integers as metadata keys?

justin-schroeder:
@ob-stripe yep. I suppose we're an odd one out in that sense ha. Sometimes we use numeric keys to "fake" a sortable array of properties.

ob-stripe:
Just to be sure there's not another issue at play here, this can only happen if you call `toArray()` after setting metadata with integer keys but before calling `save()`, right? The API should always return strings for keys, even if integers were passed on the way in.

justin-schroeder:
@ob-stripe actually no, the code was a pretty simple controller method (context is Laravel):

```
    public function getPlans(Request $request)
    {
        $all_plans = \Stripe\Plan::all();
        return response($all_plans);
    }
```

Some of the plans happen to have metadata with numeric keys.

ob-stripe:
Just released stripe-php 7.27.3 with your fix, thanks again.

Would you mind sharing your account ID and one of the plan IDs so we can take a closer look at what happens on our end?

justin-schroeder:
Awesome, thanks for the quick fix.

acct_1G1bAyAwqhErLVlI

plan_GqkWIob7obsABb
plan_GqkW2BWGqt1BEZ

PR #897, opened 2020-03-18
Fixes issue #896

justin-schroeder:
In this PR:

- Fixes issue introduced in PHP 7.4 with accessing an integer as an array (`123[0]` now throws a notice)

### DIFF
--- a/lib/StripeObject.php
+++ b/lib/StripeObject.php
@@ -443,7 +443,7 @@ class StripeObject implements \ArrayAccess, \Countable, \JsonSerializable
         };
 
         return \array_reduce(\array_keys($this->_values), function ($acc, $k) use ($maybeToArray) {
-            if ('_' === $k[0]) {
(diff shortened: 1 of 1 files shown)


ob-stripe:
Thanks for the fix @justin-schroeder. FWIW I think the @stripe-ci bot is a bit too aggressive here: it wanted you to create the PR from a feature branch in your fork rather than from the `master` branch. I don't think that's necessary for PRs that come from external forks, this should probably only apply to our internal repos anyway.

I've opened a PR based on your fix here: #899. I added a test to make sure we're covering the bug, and explicitly converted the key to string to comply with PHP 7.1's strict types (we're not yet using them because we're still supporting older PHP versions at this time, but we will eventually).

justin-schroeder:
Great thanks @ob-stripe. Yeah I couldn't quite figure out what the bot was asking me to do, and decided to bail so there weren't 15 closed PRs ha.

PR #899, opened 2020-03-18
Convert keys to strings in `StripeObject::toArray()`

### LINKED ISSUE #896: Support PHP 7.4

ob-stripe:
r? @cjavilla-stripe @remi-stripe 
cc @stripe/api-libraries @justin-schroeder

Convert keys to strings in `StripeObject::toArray()`.

Fixes #896.

### DIFF
--- a/lib/StripeObject.php
+++ b/lib/StripeObject.php
@@ -443,7 +443,7 @@ class StripeObject implements \ArrayAccess, \Countable, \JsonSerializable
         };
 
         return \array_reduce(\array_keys($this->_values), function ($acc, $k) use ($maybeToArray) {
-            if ('_' === $k[0]) {
+            if ('_' === \substr((string) $k, 0, 1)) {
                 return $acc;
             }
             $v = $this->_values[$k];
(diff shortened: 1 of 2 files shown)


ob-stripe:
Released as 7.27.3.

PR #901, opened 2020-03-24
Add strict parsing for RequestOptions

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Add strict parsing option to `RequestOptions::parse()`. When enabled, strict parsing will throw an `InvalidArgumentException` when:
- a string is passed (to deprecate passing per-request API keys as strings, and instead force the use of arrays like `['api_key' => 'sk_...']`)
- an array with unexpected keys is passed

Strict parsing is disabled by default, but is enabled when used from `StripeClient`. This makes this change fully backwards compatible for existing request methods on models, but forces strict parsing when using request methods on services.

### DIFF
--- a/lib/Util/RequestOptions.php
+++ b/lib/Util/RequestOptions.php
@@ -2,22 +2,30 @@
 
 namespace Stripe\Util;
 
-use Stripe\Exception;
-
 class RequestOptions
 {
     /**
-     * @var array a list of headers that should be persisted across requests
+     * @var array<string> a list of headers that should be persisted across requests
      */
     public static $HEADERS_TO_PERSIST = [
         'Stripe-Account',
         'Stripe-Version',
     ];
 
+    /** @var array<string, string> */
     public $headers;
+
+    /** @var null|string */
     public $apiKey;
+
+    /** @var null|string */
     public $apiBase;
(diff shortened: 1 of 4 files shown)


lib/Util/RequestOptions.php
```diff
@@ -86,27 +101,46 @@ public static function parse($options)
...
                 $headers['Idempotency-Key'] = $options['idempotency_key'];
+                unset($options['idempotency_key']);
             }
             if (\array_key_exists('stripe_account', $options)) {
                 $headers['Stripe-Account'] = $options['stripe_account'];
+                unset($options['stripe_account']);
             }
             if (\array_key_exists('stripe_version', $options)) {
                 $headers['Stripe-Version'] = $options['stripe_version'];
+                unset($options['stripe_version']);
```
remi-stripe: can you add a clear test proving that `unset` can't mutate the input array? I know we discussed this and it's safe but that always makes me uneasy

ob-stripe: I've verified it manually, but I'd rather not add a test for it since it's how PHP itself works. We'd be testing the language, not the stripe-php library.

PR #903, opened 2020-03-25
Change StripeClient initialization

ob-stripe:
cc @stripe/api-libraries 

Change the `StripeClient` constructor to expose a simpler API.

The new constructor takes a single argument, which can be a string (the API key) or an array with various configuration values.

In most cases, users will use the string form:
```php
$stripe = new \Stripe\StripeClient('sk_test_123');
```

If needed, it's easy to pass "advanced" configuration settings:
```php
$stripe = new \Stripe\StripeClient([
    'api_key' => 'sk_test_123',
    'client_id' => 'ca_123',
]);
```

The configuration array approach is used by other PHP libraries, including Guzzle, the most popular PHP HTTP client library (https://github.com/guzzle/guzzle/blob/1cdd69a0cd361ee8b6ba5584e45db80c0ea8b31c/src/Client.php#L51). The main downside is that there is no standard PHPDoc syntax for defining the shape of the array, but I think PHPStan and other static analyzer tools are working on something like this, and it's not _that_ bad to just use freeform text to expose the various configuration keys and their expected types.

### DIFF
--- a/lib/BaseStripeClient.php
+++ b/lib/BaseStripeClient.php
@@ -29,21 +29,42 @@ class BaseStripeClient implements StripeClientInterface
     private $filesBase;
 
     /**
-     * Initializes a new instance of the {@link StripeClient} class.
+     * Initializes a new instance of the {@link BaseStripeClient} class.
      *
-     * @param null|string $apiKey the API key used by the client to make requests
-     * @param null|string $clientId the client ID used by the client in OAuth requests
-     * @param null|string $apiBase The base URL for Stripe's API. Defaults to {@link DEFAULT_API_BASE}.
-     * @param null|string $connectBase The base URL for Stripe's OAuth API. Defaults to {@link DEFAULT_CONNECT_BASE}.
-     * @param null|string $filesBase The base URL for Stripe's Files API. Defaults to {@link DEFAULT_FILES_BASE}.
+     * The constructor takes a single argument. The argument can be a string, in which case it
+     * should be the API key. It can also be an array with various configuration settings.
+     *
+     * Configuration settings include the following options:
+     *
+     * - api_key (string): the Stripe API key, to be used in regular API requests.
+     * - client_id (string): the Stripe client ID, to be used in OAuth requests.
+     * - api_base (string): the base URL for regular API requests. Defaults to
+     *   {@link DEFAULT_API_BASE}. Changing this should rarely be necessary.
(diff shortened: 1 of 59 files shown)


lib/BaseStripeClient.php
```diff
@@ -29,21 +29,42 @@ class BaseStripeClient implements StripeClientInterface
...
+     * Configuration settings include the following options:
+     *
+     * - api_key (string): the Stripe API key, to be used in regular API requests.
+     * - client_id (string): the Stripe client ID, to be used in OAuth requests.
+     * - api_base (string): the base URL for regular API requests. Defaults to
+     *   {@link DEFAULT_API_BASE}. Changing this should rarely be necessary.
+     * - connect_base (string): the base URL for OAuth requests. Defaults to
+     *   {@link DEFAULT_CONNECT_BASE}. Changing this should rarely be necessary.
+     * - files_base (string): the base URL for file creation requests. Defaults to
+     *   {@link DEFAULT_FILES_BASE}. Changing this should rarely be necessary.
```
remi-stripe: you're missing `stripe_account` right?

tests/Stripe/Service/ApplePayDomainServiceTest.php
```diff
@@ -23,7 +23,7 @@ final class ApplePayDomainServiceTest extends \PHPUnit\Framework\TestCase
      */
     protected function setUpService()
     {
-        $this->client = new \Stripe\StripeClient('sk_test_123', null, MOCK_URL);
+        $this->client = new \Stripe\StripeClient(['api_key' => 'sk_test_123', 'api_base' => MOCK_URL]);
```
remi-stripe: Minor but as I see this change, can we make this a global variable? If we ever want to add something or change the default key or whatever it would make the PR more readable

lib/BaseStripeClient.php
ob-stripe: I haven't added any new configuration options yet, like `stripe_version` or `stripe_account`. I'll do that in a followup PR.

tests/Stripe/Service/ApplePayDomainServiceTest.php
ob-stripe: I plan on cleaning up the service tests later by factorizing code in a `ServiceTestHelper` trait or something like that.

PR #908, opened 2020-04-01
Update README

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Update README file to use `StripeClient`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,8 +53,8 @@ If you use Composer, these dependencies should be handled automatically. If you 
 Simple usage looks like:
 
 ```php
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -125,7 +125,7 @@ end up there instead of `error_log`:
 You can access the data from the last API response on any object via `getLastResponse()`.
 
 ```php
-$customer = \Stripe\Customer::create([
+$customer = $stripe->customers->create([
```
remi-stripe, line 128: shouldn't we recreate the client each time? That seems important to show right?

PR #920, opened 2020-04-22
Add support for the `Session` resource and APIs on the `BillingPortal` namespace

remi-stripe:
r? @richardm-stripe 
cc @stripe/api-libraries


lib/Util/Util.php
```diff
@@ -57,6 +57,7 @@ public static function convertToStripeObject($resp, $opts)
             \Stripe\Balance::OBJECT_NAME => \Stripe\Balance::class,
             \Stripe\BalanceTransaction::OBJECT_NAME => \Stripe\BalanceTransaction::class,
             \Stripe\BankAccount::OBJECT_NAME => \Stripe\BankAccount::class,
+            \Stripe\BillingPortal\Session::OBJECT_NAME => \Stripe\BillingPortal\Session::class,
```
remi-stripe, line 60: Flagging this wasn't codegen-ed yet and I had to manually fix it and we should prioritize. I would have missed it if I had not added tests.

richardm-stripe: ack

PR #921, opened 2020-04-22
Fix build error on master

remi-stripe:
This is a build error we have on master for some reason so doing a PR to fix this.

r? @ob-stripe 
cc @stripe/api-libraries


ob-stripe:
Closing in favor of #922, the `\get_class($this)` -> `static::class` change is mildly scary and I don't have time to thoroughly test it atm, so going to pin the existing version for now and we can do an explicit upgrade later.

PR #922, opened 2020-04-22
Pin PHP-CS-Fixer version

ob-stripe:
Pin PHP-CS-Fixer to a specific version to avoid builds suddenly failing when a new version is released.


ob-stripe:
Not sure why the checks aren't appearing in GitHub, but the build is successful: https://travis-ci.org/github/stripe/stripe-php/builds/678275185

PR #924, opened 2020-04-29
Add support for the `Price` resource and APIs

remi-stripe:
Codegen for openapi f75fd88

r? @ob-stripe 
cc @stripe/api-libraries @areagle-stripe


lib/Source.php
```diff
@@ -89,7 +89,7 @@ public function detach($params = null, $opts = null)
 
         $id = $this['id'];
         if (!$id) {
-            $class = \get_class($this);
+            $class = static::class;
```
ob-stripe, line 92: I pinned PHP-CS-Fixer's version to avoid this change for the time being. Can you reinstall your dependencies (`rm -rf vendor/ composer.lock && composer update`) and reset the changes to `Source.php` and `StripeObject.php`?

remi-stripe: Fixed, PTAL

PR #926, opened 2020-05-01
[wip] Accept null values for params

ob-stripe:
Accept null values for parameters in service methods, and send a null string to the API (instead of just ignoring them as is the case in resource methods).

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,8 +53,8 @@ If you use Composer, these dependencies should be handled automatically. If you 
 Simple usage looks like:
 
 ```php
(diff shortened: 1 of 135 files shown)


remi-stripe:
Is this one stale because it was merged/rebuilt in https://github.com/stripe/stripe-php/pull/771 ?

ob-stripe:
Yep, this was implemented in https://github.com/stripe/stripe-php/pull/771/commits/b48e9dd8ec27388772dc84195c0f31e293c2674b.

PR #930, opened 2020-05-11
Services: null as emptystring

richardm-stripe:
r? @ob-stripe 
For resource-style PHP methods, we ignore any `null` value that is sent inside `$params`. For service-style PHP methods this is no longer necessary because we do not support `->save`, and we should understand that a null param is an instruction to unset the field in question -- which our API supports through sending an empty string. This pull request changes `AbstractServiceMethod` so that all incoming null params are translated to empty strings.

### DIFF
--- a/tests/Stripe/Service/AbstractServiceTest.php
+++ b/tests/Stripe/Service/AbstractServiceTest.php
@@ -26,6 +26,14 @@ final class AbstractServiceTest extends \PHPUnit\Framework\TestCase
         $this->service = new \Stripe\Service\CouponService($this->client);
     }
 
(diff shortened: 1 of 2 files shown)


tests/Stripe/Service/AbstractServiceTest.php
```diff
@@ -26,6 +26,14 @@ public function setUpMockService()
         $this->service = new \Stripe\Service\CouponService($this->client);
     }
 
+    public function testNullGetsEmptyStringified()
```
richardm-stripe, line 29: Not too happy with this testing strategy. Open to recommendations.

PR #933, opened 2020-05-15
Services bugfix: handle null properly on arrays

richardm-stripe:
r? @ob-stripe 
cc @stripe/api-libraries 
With PHP services, sending `null` on a method's params is translated to an empty string when it is sent to the API, to cause the API to unset that field. This was only working properly at the top level. This PR causes it to work correctly for nested fields.

### DIFF
--- a/tests/Stripe/Service/AbstractServiceTest.php
+++ b/tests/Stripe/Service/AbstractServiceTest.php
@@ -57,4 +57,16 @@ final class AbstractServiceTest extends \PHPUnit\Framework\TestCase
 
         $this->service->retrieve(' ');
     }
+
+    public function testFormatParams()
+    {
+        $result = \Stripe\Service\AbstractService::formatParams(['foo' => null]);
+        static::assertTrue('' === $result['foo']);
+        static::assertTrue(null !== $result['foo']);
+
+        $result = \Stripe\Service\AbstractService::formatParams(['foo' => ['bar' => null, 'baz' => 1]]);
+        static::assertTrue('' === $result['foo']['bar']);
+        static::assertTrue(null !== $result['foo']['bar']);
+        static::assertTrue(1 === $result['foo']['baz']);
+    }
 }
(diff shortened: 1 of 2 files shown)


tests/Stripe/Service/AbstractServiceTest.php
```diff
@@ -57,4 +57,22 @@ public function testRetrieveThrowsIfIdNullIsWhitespace()
...
+        $result = \Stripe\Service\AbstractService::formatParams(['foo' => null]);
+        static::assertTrue('' === $result['foo']);
+        static::assertTrue(null !== $result['foo']);
+
+        $result = \Stripe\Service\AbstractService::formatParams(['foo' => ['bar' => null, 'baz' => 1]]);
+        static::assertTrue('' === $result['foo']['bar']);
+        static::assertTrue(null !== $result['foo']['bar']);
+        static::assertTrue(1 === $result['foo']['baz']);
+
+        $result = \Stripe\Service\AbstractService::formatParams(['foo' => ["bar", null, null, "baz"]]);
```
remi-stripe, line 72: What about passing 
```
[
  'email' => null,
  'invoice_settings' => [
    'default_payment_method' => 'null',
  ],
]
```

As in, this method didn't seem recurring or able to handle non top-level params

richardm-stripe: I don't quite follow -- this PR adds a recursive call https://github.com/stripe/stripe-php/pull/933/files#diff-d6b577df0148860a08e6dfb3025bd0d0R52

and this test case covers non top-level params -- everything is nested one level under "foo".

richardm-stripe:
ptal @remi-stripe

richardm-stripe:
Thanks! Made private
ptal @ob-stripe

ob-stripe:
Formatting error is causing CI to fail, ptal @richardm-stripe

richardm-stripe:
Fixed, sorry
ptal @ob-stripe

PR #938, opened 2020-05-22
Remove extra array_keys() call.

ellisio:
This seems redundant, since the variable isn't being used.


ob-stripe:
LGTM. Assigning to @richardm-stripe for merging and releasing.

PR #941, opened 2020-05-25
Add pending update constants

driesvints:
I wasn't sure where to send this to but thought the subscription object was probably the right place. This PR adds the different values for the `payment_behavior` field to accommodate pending updates: https://stripe.com/docs/billing/subscriptions/pending-updates

### DIFF
--- a/lib/Subscription.php
+++ b/lib/Subscription.php
@@ -63,6 +63,10 @@ class Subscription extends ApiResource
     const STATUS_TRIALING = 'trialing';
     const STATUS_UNPAID = 'unpaid';
 
+    const ALLOW_INCOMPLETE = 'allow_incomplete';
+    const ERROR_IF_INCOMPLETE = 'error_if_incomplete';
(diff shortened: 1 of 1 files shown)


remi-stripe:
@driesvints Sorry for the delay on that one and dropping the ball. We'll investigate this week and either merge or come back to you with changes. We need to do some work internally as stripe-php is automatically generated from the openapi spec so changes like this need to be added as "overrides" internally in our tooling first.

driesvints:
@remi-stripe no worries at all! Definitely not urgent. Thanks for all your work 👍

remi-stripe:
@driesvints We just released a fix in 7.39.0. We did our version as it needed proper prefixing for the parameter name and it's now  `PAYMENT_BEHAVIOR_ALLOW_INCOMPLETE` and similar! Sorry again for the delay on that one!

driesvints:
Awesome! Thanks again :)

PR #942, opened 2020-05-26
fix autopagination for Service

### LINKED ISSUE #939: [bug] StripeClient subscriptions & autoPagingIterator - missing parameters
Hi there.

I'm migrating from 6.41 to 7.36.
Problem occurred when using autoPagingIterator on a subscriptions collection coming from a StripeClient. I didn't try on other types.

## Usages

```
// legacy code with singleton
$subscriptions = Stripe\Subscription::all([
    'status' => 'all',
    'limit' => 100,
]);

/** @var Stripe\Subscription $subscription */
foreach ($subscriptions->autoPagingIterator() as $subscription) {
    // subscription processing
}
```

```
// StripeClient version
$subscriptions = $this->client->subscriptions->all([
    'status' => 'all',
    'limit' => 100,
]);

/** @var Stripe\Subscription $subscription */
foreach ($subscriptions->autoPagingIterator() as $subscription) {
    // subscription processing
}
```

## Problem

The last subscription status from first iteration is `cancelled`.
When using StripeClient version, I get the following error : "No such subscription: sub_XXXX"

## Explanation

StripeClient autoPagingIterator does not forward `$param` from the initial request.

See `$param` argument values from the `ApiRequestor::request()` function.

### Legacy version

```
// first iteration
array:2 [
    "status" => "all"
    "limit" => 100
]

// second iteration
array:3 [
    "status" => "all"
    "limit" => 100
    "starting_after" => "sub_XXXX"
]
```

### StripeClient version

```
// first iteration
array:2 [
    "status" => "all"
    "limit" => 100
]

// second iteration
// note that status & limit are missing.
array:1 [
    "starting_after" => "sub_XXXX"
]
```

Edit : Explanation legacy / StripeClient have been inverted. Problem is still the same.

pepin-stripe:
Should fix #939.

Briefly, `autoPagingIterator` expects the `filters` property on `Stripe\Collection` to be set from `$params`.

This was happening before in `\ApiOperations\All::all` https://github.com/stripe/stripe-php/blob/master/lib/ApiOperations/All.php#L33, but isn't done in
`BaseStripeClient#request`

https://github.com/stripe/stripe-php/blob/master/lib/BaseStripeClient.php#L129-L140.

This PR introduces `BaseStripeClient#requestCollection` (and `AbstractService#requestCollection`, to call it), which simply calls `#request` and then `->setFilters($params)`, and then we updated the "SubscriptionService" to use it.

ptal @ob-stripe to see if this seems right (feel free to bikeshed on naming)

I prefer adding a separate #requestCollection method rather than adding the call conditionally to `#request` -- since `->all` should always return a `\Stripe\Collection`, it seems better that it call a method guaranteed to return the narrower `\Stripe\Collection` rather than the broader `\Stripe\Object`

### DIFF
--- a/lib/BaseStripeClient.php
+++ b/lib/BaseStripeClient.php
@@ -151,16 +151,16 @@ class BaseStripeClient implements StripeClientInterface
      */
     public function requestCollection($method, $path, $params, $opts)
     {
-         $obj = $this->request($method, $path, $params, $opts);
-         if (!($obj instanceof \Stripe\Collection)) {
-             $received_class = \get_class($obj);
-             $msg = "Expected to receive `Stripe\Collection` object from Stripe API. Instead received `{$received_class}`.";
-             throw new \Stripe\Exception\UnexpectedValueException
-             ($msg);
-         }
-         $obj->setFilters($params);
+        $obj = $this->request($method, $path, $params, $opts);
+        if (!($obj instanceof \Stripe\Collection)) {
+            $received_class = \get_class($obj);
+            $msg = "Expected to receive `Stripe\\Collection` object from Stripe API. Instead received `{$received_class}`.";
 
-         return $obj;
+            throw new \Stripe\Exception\UnexpectedValueException($msg);
+        }
+        $obj->setFilters($params);
+
+        return $obj;
     }
 
     /**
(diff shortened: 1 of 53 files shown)


pepin-stripe:
ptal @richardm-stripe

richardm-stripe:
Ok, will add codegen for this and release tomorrow.

richardm-stripe:
codegenned,
ptal @ob-stripe again

PR #961, opened 2020-06-25
[WIP] Codegen more enum values

richardm-stripe:
Up until now, we generate constants for enums that appear at the top-level of stripe responses and put them at the top of the corresponding resource class. We do not generate constants corresponding to

* values of enums that appear above the top level. For example `account.capabilities.card_issuing` is an `enum("active", "inactive", "pending")`, which is not generated according to the current scheme.
* values of enums that appear not on any responses, but are to be passed in on the request. For example,  `Subscription#create.payment_behavior` (as reported in #941).

This PR adds those enums.

**Q. What about cases when the values of the request enum differs from the response enum, or other request enums on the resource?**
A. The enum values represent the superset of all enums sharing that name.

TODO: This doesn't yet include enums "nullable" or "emptyStringable" enums, or arrays of enums (if those exist)


lib/Account.php
```diff
@@ -42,24 +42,119 @@ class Account extends ApiResource
     use ApiOperations\NestedResource;
     use ApiOperations\Update;
 
+    const AU_BECS_DEBIT_PAYMENTS_ACTIVE = 'active';
```
remi-stripe, line 45: I don't think we should do those this way since they are just identical. If I had done this right originally, it should be a list of `STATUS_*` constants inside the `Capability` class instead right?

lib/Account.php
```diff
@@ -42,24 +42,119 @@ class Account extends ApiResource
...
-    const CAPABILITY_STATUS_PENDING = 'pending';
+    const CARD_ISSUING_ACTIVE = 'active';
+    const CARD_ISSUING_INACTIVE = 'inactive';
+    const CARD_ISSUING_PENDING = 'pending';
+
+    const CARD_PAYMENTS_ACTIVE = 'active';
+    const CARD_PAYMENTS_INACTIVE = 'inactive';
+    const CARD_PAYMENTS_PENDING = 'pending';
+
+    const CODE_INVALID_ADDRESS_CITY_STATE_POSTAL_CODE = 'invalid_address_city_state_postal_code';
```
remi-stripe, line 66: I really don't think we should add something like this to the codebase, but I'm obviously biased against all those constants in the first place. 

IMO those only make sense the day we start supporting real enums and API versions. Because right now, any value disappearing would be a breaking change or require an override

lib/Account.php
```diff
@@ -42,24 +42,119 @@ class Account extends ApiResource
...
+    const CODE_VERIFICATION_FAILED_ADDRESS_MATCH = 'verification_failed_address_match';
+    const CODE_VERIFICATION_FAILED_BUSINESS_IEC_NUMBER = 'verification_failed_business_iec_number';
+    const CODE_VERIFICATION_FAILED_DOCUMENT_MATCH = 'verification_failed_document_match';
+    const CODE_VERIFICATION_FAILED_ID_NUMBER_MATCH = 'verification_failed_id_number_match';
+    const CODE_VERIFICATION_FAILED_KEYED_IDENTITY = 'verification_failed_keyed_identity';
+    const CODE_VERIFICATION_FAILED_KEYED_MATCH = 'verification_failed_keyed_match';
+    const CODE_VERIFICATION_FAILED_NAME_MATCH = 'verification_failed_name_match';
+    const CODE_VERIFICATION_FAILED_OTHER = 'verification_failed_other';
+
+    const INTERVAL_DAILY = 'daily';
```
remi-stripe, line 105: I don't think this name is right. It should be something like `SETTINGS_PAYOUTS_SCHEDULE_INTERVAL_DAILY` instead? Back to "let's not add any of those" in my mind

lib/Account.php
```diff
@@ -42,24 +42,119 @@ class Account extends ApiResource
...
+
+    const JCB_PAYMENTS_ACTIVE = 'active';
+    const JCB_PAYMENTS_INACTIVE = 'inactive';
+    const JCB_PAYMENTS_PENDING = 'pending';
+
+    const LEGACY_PAYMENTS_ACTIVE = 'active';
+    const LEGACY_PAYMENTS_INACTIVE = 'inactive';
+    const LEGACY_PAYMENTS_PENDING = 'pending';
+
+    const STRUCTURE_GOVERNMENT_INSTRUMENTALITY = 'government_instrumentality';
```
remi-stripe, line 118: same as above, it should be prefixed by the full path to avoid naming conflicts no?

lib/Account.php
```diff
@@ -42,24 +42,119 @@ class Account extends ApiResource
...
+    const STRUCTURE_PRIVATE_PARTNERSHIP = 'private_partnership';
+    const STRUCTURE_PUBLIC_COMPANY = 'public_company';
+    const STRUCTURE_PUBLIC_CORPORATION = 'public_corporation';
+    const STRUCTURE_PUBLIC_PARTNERSHIP = 'public_partnership';
+    const STRUCTURE_SOLE_PROPRIETORSHIP = 'sole_proprietorship';
+    const STRUCTURE_TAX_EXEMPT_GOVERNMENT_INSTRUMENTALITY = 'tax_exempt_government_instrumentality';
+    const STRUCTURE_UNINCORPORATED_ASSOCIATION = 'unincorporated_association';
+    const STRUCTURE_UNINCORPORATED_NON_PROFIT = 'unincorporated_non_profit';
+
+    const TAX_REPORTING_US_1099_K_ACTIVE = 'active';
```
remi-stripe, line 134: Minor but why the different color?

lib/Card.php
```diff
@@ -46,6 +46,9 @@ class Card extends ApiResource
     use ApiOperations\Delete;
     use ApiOperations\Update;
 
+    const ACCOUNT_HOLDER_TYPE_COMPANY = 'company';
+    const ACCOUNT_HOLDER_TYPE_INDIVIDUAL = 'individual';
```
remi-stripe, line 50: This is incorrect, there's no such thing on `Card`

richardm-stripe:
^ was interesting as a proof of concept, will re-visit at a later time when the codegen infrastructure is more mature

PR #963, opened 2020-06-27
Suggestion: Add Decoding Charset option to Webhook::constructEvent

RealityRipple:
Because `json_decode()` always returns strings in UTF-8 and because PHP doesn't seamlessly convert between encodings, developers may run into issues with special-character names or addresses sent via Webhook. If mbstring is installed (and PHP is 7.2 or newer), it'd be wonderful to allow `mb_convert_encoding()` to be called before `\lib\Webhook.php`'s `Event::constructFrom()` call. An optional parameter with the desired charset should take care of everything. This pull request is just an example of what I'm suggesting.

### DIFF
--- a/lib/Webhook.php
+++ b/lib/Webhook.php
@@ -39,7 +39,7 @@ abstract class Webhook
             throw new Exception\UnexpectedValueException($msg);
         }
 
-        if (function_exists('mb_convert_encoding') && $charset !== 'UTF-8') {
+        if (version_compare(PHP_VERSION, '7.2.0') >= 0 && function_exists('mb_convert_encoding') && $charset !== 'UTF-8') {
             $deData = mb_convert_encoding($data, $charset);
             if ($deData !== null && $deData !== $data)
                 $data = $deData;
--- a/lib/Webhook.php
+++ b/lib/Webhook.php
@@ -18,13 +18,15 @@ abstract class Webhook
      * @param string $secret secret used to generate the signature
      * @param int $tolerance maximum difference allowed between the header's
      *  timestamp and the current time
+     * @param string $charset use a character set other than UTF-8 after decoding
+     *  the JSON payload (requires mbstring)
      *
      * @throws Exception\UnexpectedValueException if the payload is not valid JSON,
      * @throws Exception\SignatureVerificationException if the verification fails
      *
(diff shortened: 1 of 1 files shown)


lib/Webhook.php
```diff
@@ -37,6 +39,12 @@ public static function constructEvent($payload, $sigHeader, $secret, $tolerance
             throw new Exception\UnexpectedValueException($msg);
         }
 
+        if (function_exists('mb_convert_encoding') && $charset !== 'UTF-8') {
+            $deData = mb_convert_encoding($data, $charset);
+            if ($deData !== null && $deData !== $data)
```
RealityRipple, line 44: I'm not sure what the best methodology for this line is, but you get the gist.

remi-stripe:
@RealityRipple Thanks a lot for taking the time to open a PR and showcase what you're after. We're going to look into this internally to figure out what the best way to handle this is

RealityRipple:
This isn't a direct contribution. It's an example of a suggested feature. If you still want me to sign, please let me know.

richardm-stripe:
Hi @RealityRipple, sorry for the long delay. I've thought about this a bit.

My first thought is that, since this involves version-specific and plugin/specific behavior it would probably be better to leave this up to the caller, and have an interface like
```php
public static function constructEvent($payload, $sigHeader, $secret, $tolerance = self::DEFAULT_TOLERANCE, $decoder) {
  ...
  return Event::constructFrom($decoder ? $decoder($data) : $data);
}
```

and then the caller would do something like

```php
$myDecoder = function($data)
{
  $converted = \mb_convert_encoding($data, 'ISO-2022-JP');
  if ($converted !== null) {
    return $converted;
  }
  return $data;
};
constructEvent($payload, $sigHeader, $secret, 300, $myDecoder);
```

On the other hand, defining `$myDecoder` isn't really much less than expecting the user to define `myConstructEvent`, and just calling `WebhookSignature::verifyHeader`, `\json_decode`, and whatever else they want before calling `Event::constructFrom`. (Which is probably what you are doing, yes?)

For now, I think I'd like to stick with the approach of keeping `constructEvent` as simple as possible and encouraging users who need more control to reimplement to suit their needs.

PR #964, opened 2020-06-30
Add OAuthService

### LINKED ISSUE #958: OAuth with client/services pattern
I'm using the new client/services pattern (StripeClient) but as far as I can tell, the OAuth doesn't support this pattern yet, which means I still have to do this to get a token:
```PHP
Stripe::setApiKey($apiKey);
$response = OAuth::token([
    'grant_type' => 'authorization_code',
    'code' => $_REQUEST['code']
]);
```
Will this be updated to work with the StripeClient?

richardm-stripe:
r? @ob-stripe 
cc @stripe/api-libraries 

Fixes #958 and adds `$stripe->oauth` with `authorizeUrl`, `token`, and `deauthorize` methods.

I chose to introduce `->requestConnect` into the StripeClientInterface, rather than overload `->request`, because to my intuition this seems cleaner, but I'm interested in second opinions.

### DIFF
--- a/lib/BaseStripeClient.php
+++ b/lib/BaseStripeClient.php
@@ -17,7 +17,7 @@ class BaseStripeClient implements StripeClientInterface
     private $config;
 
     /** @var \Stripe\Util\RequestOptions */
-    private $defaultOpts;
+    protected $defaultOpts;
 
     /**
      * Initializes a new instance of the {@link BaseStripeClient} class.
--- a/lib/BaseStripeClient.php
+++ b/lib/BaseStripeClient.php
@@ -17,7 +17,7 @@ class BaseStripeClient implements StripeClientInterface
     private $config;
 
     /** @var \Stripe\Util\RequestOptions */
-    protected $defaultOpts;
+    private $defaultOpts;
 
     /**
      * Initializes a new instance of the {@link BaseStripeClient} class.
--- a/lib/Service/OAuthService.php
+++ b/lib/Service/OAuthService.php
@@ -16,7 +16,7 @@ class OAuthService extends \Stripe\Service\AbstractService
      */
     protected function requestConnect($method, $path, $params, $opts)
     {
-        $opts = \Stripe\Util\RequestOptions::parse($opts);
+        $opts = $this->_parseOpts($opts);
         $opts->apiBase = $this->_getBase($opts);
 
         return $this->request($method, $path, $params, $opts);
@@ -34,7 +34,7 @@ class OAuthService extends \Stripe\Service\AbstractService
     {
         $params = $params ?: [];
 
-        $opts = \Stripe\Util\RequestOptions::parse($opts);
+        $opts = $this->_parseOpts($opts);
         $base = $this->_getBase($opts);
 
         $params['client_id'] = $this->_getClientId($params);
@@ -65,6 +65,16 @@ class OAuthService extends \Stripe\Service\AbstractService
         return $this->requestConnect('post', '/oauth/token', $params, $opts);
     }
 
+    /**
+     * Disconnects an account from your platform.
+     *
+     * @param null|array $params
+     * @param null|array $opts
+     *
+     * @throws \Stripe\Exception\OAuth\OAuthErrorException if the request fails
+     *
+     * @return StripeObject object containing the response from the API
+     */
     public function deauthorize($params = null, $opts = null)
     {
         $params = $params ?: [];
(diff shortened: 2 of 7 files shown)


lib/Util/RequestOptions.php
```diff
@@ -21,6 +21,9 @@ class RequestOptions
     /** @var null|string */
     public $apiBase;
 
+    /** @var null|string */
+    public $connectBase;
```
ob-stripe, line 25: I'm a little bit on the fence about this. `RequestOptions` contains options that affects a specific request. A specific request goes to a specific endpoint. IMO there should be a single parameter that contains the base URL for the request.

It's what we did in stripe-dotnet (where it is more appropriately called [`BaseUrl`](https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Services/_common/RequestOptions.cs#L26-L32)), and what I had in mind when I designed the PHP services, though the only existing use case is file creation requests: https://github.com/stripe/stripe-php/blob/master/lib/Service/FileService.php#L53-L56.

lib/Service/OAuthService.php
```diff
@@ -0,0 +1,74 @@
...
+     *
+     * @return string the URL to Stripe's OAuth form
+     */
+    public function authorizeUrl($params = null, $opts = null)
+    {
+        $params = $params ?: [];
+
+        $base = ($opts && \array_key_exists('connect_base', $opts)) ? $opts['connect_base'] : $this->client->getConnectBase();
+
+        $clientId = $this->_getClientId($params);
```
ob-stripe, line 21: It doesn't look like this variable is used.

lib/Service/OAuthService.php
```diff
@@ -0,0 +1,74 @@
...
+     * @param null|array $params
+     * @param null|array $opts
+     *
+     * @throws \Stripe\Exception\OAuth\OAuthErrorException if the request fails
+     *
+     * @return StripeObject object containing the response from the API
+     */
+    public function token($params = null, $opts = null)
+    {
+        return $this->requestConnect('post', $this->buildPath('/oauth/token'), $params, $opts);
```
ob-stripe, line 44: We should fill the `client_secret` parameter with the API key if it's not explicitly provided.

```suggestion
        $params = $params ?: [];
        if (!\array_key_exists('client_secret', $params)) {
            $params['client_secret'] = $this->client->getApiKey();
        }
        return $this->requestConnect('post', $this->buildPath('/oauth/token'), $params, $opts);
```

lib/Service/OAuthService.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public function deauthorize($params = null, $opts = null)
+    {
+        return $this->requestConnect('post', $this->buildPath('/oauth/deauthorize'), $params, null);
```
ob-stripe, line 49: We should fill the `client_id` parameter with the client ID if it's not explicitly provided.

```suggestion
        $params = $params ?: [];
        $params['client_id'] = $this->_getClientId($params);
        return $this->requestConnect('post', $this->buildPath('/oauth/deauthorize'), $params, null);
```

richardm-stripe:
ptal @ob-stripe -- I've addressed all your comments & moved 'connectRequest' out into OAuthService. Thanks for the careful review.

lib/BaseStripeClient.php
```diff
@@ -17,7 +17,7 @@ class BaseStripeClient implements StripeClientInterface
     private $config;
 
     /** @var \Stripe\Util\RequestOptions */
-    private $defaultOpts;
+    protected $defaultOpts;
```
ob-stripe, line 20: Is this necessary? You don't seem to be using it anywhere else in the PR.

lib/Service/OAuthService.php
```diff
@@ -0,0 +1,108 @@
...
+     * @throws \Stripe\Exception\OAuth\OAuthErrorException if the request fails
+     *
+     * @return StripeObject object containing the response from the API
+     */
+    public function token($params = null, $opts = null)
+    {
+        $params = $params ?: [];
+        $params['client_secret'] = $this->_getClientSecret($params);
+
+        return $this->requestConnect('post', $this->buildPath('/oauth/token'), $params, $opts);
```
ob-stripe, line 66: You don't need to use `buildPath()` since the path is constant.

```suggestion
        return $this->requestConnect('post', '/oauth/token', $params, $opts);
```

lib/Service/OAuthService.php
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    public function deauthorize($params = null, $opts = null)
+    {
+        $params = $params ?: [];
+        $params['client_id'] = $this->_getClientId($params);
+
+        return $this->requestConnect('post', $this->buildPath('/oauth/deauthorize'), $params, $opts);
```
ob-stripe, line 74: Same here.

```suggestion
        return $this->requestConnect('post', '/oauth/deauthorize', $params, $opts);
```

lib/Service/OAuthService.php
```diff
@@ -0,0 +1,108 @@
...
+     * @param null|array $params
+     * @param null|array $opts
+     *
+     * @return string the URL to Stripe's OAuth form
+     */
+    public function authorizeUrl($params = null, $opts = null)
+    {
+        $params = $params ?: [];
+
+        $base = ($opts && \array_key_exists('connect_base', $opts)) ? $opts['connect_base'] : $this->client->getConnectBase();
```
ob-stripe, line 39: I think we should use `api_base` instead of `connect_base` here. It would change the API from the existing `\Stripe\OAuth::authorizeUrl` static method, but it would make it consistent with the other two methods in the service. wdyt?

richardm-stripe: Agreed.

richardm-stripe:
ptal @ob-stripe 

1. Removed unnecessary calls to ->buildPath for constant paths.
2. Changed $defaultOptions back to private.
3. Changed `->authorizeUrl` to use `api_base` instead of `connect_base` -- I refactored this into `$opts->apiBase = $this->_getBase()` function.
4. I made `->_getBase()` throw an exception if `connect_base` is passed. Might save a user a few cycles when switching from \Stripe\OAuth.

lib/Service/OAuthService.php
```diff
@@ -0,0 +1,121 @@
...
+     */
+    public function token($params = null, $opts = null)
+    {
+        $params = $params ?: [];
+        $params['client_secret'] = $this->_getClientSecret($params);
+
+        return $this->requestConnect('post', '/oauth/token', $params, $opts);
+    }
+
+    public function deauthorize($params = null, $opts = null)
```
ob-stripe, line 68: This method should have a PHPDoc block.

lib/Service/OAuthService.php
```diff
@@ -0,0 +1,121 @@
...
+
+        return $this->client->getApiKey();
+    }
+
+    private function _getBase($opts)
+    {
+        // Throw an exception for the convenience of anybody migrating to
+        // \Stripe\Service\OAuthService from \Stripe\OAuth, where `connect_base`
+        // was the name of the parameter that behaves as `api_base` does here.
+        if (isset($opts->connect_base)) {
```
ob-stripe, line 113: I'm not sure this will work. `\Stripe\RequestOptions` does not have a `connect_base` property, even if there is a `connect_base` key in the array passed to `RequestOptions::parse()`, so I think this condition will always be false. Mind adding a test for it?

lib/Service/OAuthService.php
richardm-stripe: Added!

lib/Service/OAuthService.php
richardm-stripe: Oops! Good catch. Fixed now.

richardm-stripe:
ptal @ob-stripe fixed the $opts['connect_base'] check and added docblocks.

PR #965, opened 2020-07-02
add `Stripe-Account` to the event object for connect account

baijunyao:
For [connect account](https://stripe.com/connect), The `Stripe-Account` header is missing when converting a webhook payload to a `\Stripe\Event` object, so the following code will report an error.
```php
    $event = Webhook::constructEvent($payload, $signature, $secret);

    /** @var \Stripe\Invoice $stripe_invoice */
    $stripe_invoice = $event->data->object;

    $stripe_invoice->finalizeInvoice();
```
Error message:
```json
{
    "title": "Unprocessable Entity",
    "detail": {
        "id": [
            "No such invoice: in_xxx"
        ]
    },
    "status": 422,
    "data": [],
    "doc_url": "https:\/\/stripe.com\/docs\/error-codes\/resource-missing"
}
```

### DIFF
--- a/lib/Webhook.php
+++ b/lib/Webhook.php
@@ -37,6 +37,8 @@ abstract class Webhook
             throw new Exception\UnexpectedValueException($msg);
         }
 
-        return Event::constructFrom($data);
+        $opts = isset($data['account']) ? ['stripe_account' => $data['account']] : null;
+
+        return Event::constructFrom($data, $opts);
     }
 }
--- a/tests/Stripe/WebhookTest.php
+++ b/tests/Stripe/WebhookTest.php
@@ -2,6 +2,8 @@
 
 namespace Stripe;
 
+use ReflectionClass;
+
 /**
  * @internal
  * @covers \Stripe\Webhook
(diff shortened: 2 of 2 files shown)


richardm-stripe:
Hi @baijunyao, thank you for the contribution, sorry for letting this sit around so long.

I don't think we want to merge this -- in my view, it is appropriate that users should specify explicitly in the `$opts` when they want to send a request with a specific stripe_account header, even when the object was attained via deserialization from a webhook event triggered by a particular account: i.e.
```php
$event = Webhook::constructEvent($payload, $signature, $secret);
/** @var \Stripe\Invoice $stripe_invoice */
$stripe_invoice = $event->data->object;
$stripe_invoice->finalizeInvoice(null, ['stripe_account' => $event->account]);
```
I know that it's *usually* the case that the account that triggered the webhook event will be the account that you want to set the `stripe_account` header for, but in theory this might not be the case, and some object expanded on the webhook endpoint should be queried with your main stripe account. It is better to be explicit.

Also, with the introduction of [StripeClient](https://github.com/stripe/stripe-php/wiki/Migration-to-StripeClient-and-services-in-7.33.0) in 7.33.0, the preferred way to finalize an invoice from a webhook would be something like
```php
$stripe = new \Stripe\StripeClient('sk_test_xxxx');
...
$event = Webhook::constructEvent($payload, $signature, $secret);
/** @var \Stripe\Invoice $stripe_invoice */
$stripe_invoice = $event->data->object;
$stripe->invoices->finalizeInvoice(
  $stripe_invoice->id,
  ["stripe_account" => $event->account]
);
```
and there is no way to do this sort of persistence because `$stripe->invoices->finalizeInvoice` doesn't have access to the deserialized invoice.

PR #971, opened 2020-07-14
Adds enum values for `billing_address_collection` on Checkout `Session`

cjavilla-stripe:
* Adds enum values for `billing_address_collection` on Checkout `Session`

r? @remi-stripe 
cc @stripe/api-libraries


cjavilla-stripe:
@richardm-stripe I think this might be dependent on the oauth PR. Can you let me know once thats merged into autogen and I'll regenerate?

PR #988, opened 2020-07-31
add delete customer

Edu-Vin:
Customers can be deleted using <code>Stripe\Customer::delete('cus_Hiu5Sh7Bjvzczc', [])</code>  based on the api documentation.

### DIFF
--- a/lib/Customer.php
+++ b/lib/Customer.php
@@ -77,6 +77,23 @@ class Customer extends ApiResource
         $this->refreshFrom(['discount' => null], $opts, true);
     }
 
+    /**
+     * @param string $id The ID of customer to be deleted.
+     * @param array|null $params
+     * @param array|string|null $options
+     *
+     * @return Obj
+     */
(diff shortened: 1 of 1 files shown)


remi-stripe:
@Edu-Vin Thanks for the PR! Unfortunately we can't add a static method like this to the Customer object because it conflicts with the instance version.

Instead, you should use the Client/Service infrastructure which we shipped recently that lets you delete a customer directly without having to fetch it first. This is documented here: https://github.com/stripe/stripe-php/wiki/Migration-to-StripeClient-and-services-in-7.33.0

It's also the way documented here: https://stripe.com/docs/api/customers/delete

```
$stripe = new \Stripe\StripeClient(
  'sk_test_123'
);
$stripe->customers->delete(
  'cus_123',
  []
);
```

I'm going to close this out since the future of the library is those services which are actively maintained. We even automatically generate the code from our openapi spec!

PR #991, opened 2020-08-07
Fix invalid return types in OAuthService

ostrolucky:


richardm-stripe:
Hi @ostrolucky, thanks for the contribution!

PR #994, opened 2020-08-13
Nullable balance_transactions on issuing disputes

richardm-stripe:
r? @ob-stripe (feel free to change reviewer)
cc @stripe/api-libraries

* `Issuing.dispute.balance_transactions` is now nullable.

### DIFF
--- a/lib/Issuing/Dispute.php
+++ b/lib/Issuing/Dispute.php
@@ -16,7 +16,7 @@ namespace Stripe\Issuing;
  *
  * @property string $id Unique identifier for the object.
  * @property string $object String representing the object's type. Objects of the same type share the same value.
- * @property \Stripe\BalanceTransaction[] $balance_transactions List of balance transactions associated with this dispute.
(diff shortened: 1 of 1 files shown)


lib/Issuing/Dispute.php
```diff
@@ -16,7 +16,7 @@
  *
  * @property string $id Unique identifier for the object.
  * @property string $object String representing the object's type. Objects of the same type share the same value.
- * @property \Stripe\BalanceTransaction[] $balance_transactions List of balance transactions associated with this dispute.
+ * @property null|\Stripe\BalanceTransaction[] $balance_transactions List of balance transactions associated with this dispute.
```
remi-stripe, line 19: Why doesn't that one have `bank_name` for ACSS like the other one @richardm-stripe

remi-stripe: ah nervemind it's only top-level properties that's why

Issue #992, opened 2020-08-09
Are response headers returned by HTTP client expected to be flat or not?


ostrolucky:
Hi, I'm building [PSR-18 adapter for stripe-php](https://github.com/psr18-adapter/stripe-php) and it's unclear if [Stripe\HttpClient\ClientInterface::request](https://github.com/stripe/stripe-php/blob/fd57205d3e3a1dccab80538654128d5c46a1f5ac/lib/HttpClient/ClientInterface.php#L21) method should return flat list of headers, or PSR-7 style headers where headers with multiple values are grouped under same key as another array value. So basically, should list of headers be `array<string, string>`, or `array<string, list<string>>`?

My confusion lies because your test cases assert for flat values https://github.com/stripe/stripe-php/blob/fd57205d3e3a1dccab80538654128d5c46a1f5ac/tests/Stripe/HttpClient/CurlClientTest.php#L339 but your code expects multi dimensional list https://github.com/stripe/stripe-php/blob/fd57205d3e3a1dccab80538654128d5c46a1f5ac/lib/ApiRequestor.php#L382

And as far as I can see, the code given is the only time your library happens to read response headers.

ctrudeau-stripe:
Hello @ostrolucky, 

In CurlClient which implements ClientInterface::request the rheaders variable originates from [there](https://github.com/stripe/stripe-php/blob/fd57205d3e3a1dccab80538654128d5c46a1f5ac/lib/HttpClient/CurlClient.php#L294-L305). And its elements will always be a single string. 

ApiRequestor rheaders is a completely separate variable. 

Let us know if you have any other questions.

ctrudeau-stripe:
Hello @ostrolucky, as far as we can tell it is always flat. This alerted us to a bug in our telemetry reporting. Thank you for pointing it out!

ostrolucky:
Alright. And what's the expectation when server returns multiple values for same header name? Eg.
```
Set-Cookie: PHPSESSID=abc; path=/
Set-Cookie: PHPSESSID=def; path=/
Set-Cookie: PHPSESSID=ghi; path=/
```

richardm-stripe:
The current behavior seems to be that it sets the value equal to the *last* header specified.

ostrolucky:
Alright thx, I'll make my adapter to behave same way

PR #996, opened 2020-08-14
Fixing telemetry

ctrudeau-stripe:
User pointed out [an issue](https://github.com/stripe/stripe-php/issues/992) where we discovered that we weren't sending telemetry correctly since #781. This removes that and checks to see if the request ID is a string. 

r? @ob-stripe 
cc? @stripe/api-libraries

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -377,9 +377,11 @@ class ApiRequestor
             $hasFile
         );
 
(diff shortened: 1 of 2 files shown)


ob-stripe:
I'm not entirely sure how I managed to mess this up so badly, but it sure looks like I did 😭 

The fix looks correct to me, but I would recommend spying on actual traffic with a tool like `tcpflow` to make sure the telemetry header is as expected.

PR #1004, opened 2020-08-28
Fix PHPDocs

remi-stripe:
Codegen for openapi 6c45bc9

r? @richardm-stripe 
cc @stripe/api-libraries


remi-stripe:
it's now stale so closing and will re-gen.

PR #1009, opened 2020-09-08
FIX: delete spaces from disable_functions string

flinebux:
Sometime disable_functions config consist spaces at string.

For example my config is "dl, apache_child_terminate, apache_setenv, define_syslog_variables, eval, exec, fp, fput, ftp_connect, ftp_exec, ftp_get, ftp_login, ftp_nb_fput, ftp_put, ftp_raw, ftp_rawlist, highlight_file, ini_alter, ini_get_all, ini_restore, inject_code, mysql_pconnect, openlog, passthru, php_uname, phpAds_remoteInfo, phpAds_XmlRpc, phpAds_xmlrpcDecode, phpAds_xmlrpcEncode, popen, posix_getpwuid, posix_kill, posix_mkfifo, posix_setpgid, posix_setsid, posix_setuid, posix_setuid, posix_uname, proc_close, proc_get_status, proc_nice, proc_terminate, shell_exec, syslog, system, xmlrpc_entity_decode".

With config as at example we get error: "php_uname() has been disabled for security reasons", because we didnt trim spaces from the string.

PR #1015, opened 2020-09-21
Trim spaces when parsing php.ini disablefunctions

richardm-stripe:
Fixes #1009

r? @ctrudeau-stripe 
cc @stripe/api-libraries 

Summary: we were throwing exceptions by trying to call \php_uname on systems where \php_uname isn't supported because it was disabled in the php.ini, and we were not detecting the disabling correctly. I tested this by manually executing the function on my machine and setting various values of `disable_functions` inside my own /etc/php.ini.

Thanks for the contribution in #1009 @flinebux. I took a slightly different approach because I wanted to try and add a test as well.

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -258,6 +258,27 @@ class ApiRequestor
     /**
      * @static
      *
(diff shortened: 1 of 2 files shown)


ctrudeau-stripe (approved):
LGTM thanks for adding a test!

PR #1019, opened 2020-09-25
PHPDoc improvements

richardm-stripe:
Our PHPDocs for a couple "polymorphic groups" have used `\Stripe\StripeObject` as a fallback, but our codegen is now capable of providing a more specific union type in a couple of these instances.
r? @ctrudeau-stripe 
cc @stripe/api-libraries


ctrudeau-stripe (approved):
LGTM

PR #1025, opened 2020-10-10
Type-hint `AbstractService` constructor

angsmugmug:
To enable DI container loading of individual Stripe Service classes.

**Details**

For the usage pattern where individual Stripe Service classes are utilized, e.g.:

```
class MyCustomPaymentService
{
    public function __construct(\Stripe\Service\PaymentIntentService $stripePaymentIntents) { ... }
}
```

When a DI container is used to load `MyCustomPaymentService`, a configured `Stripe\StripeClientInterface` client cannot be loaded due to the lack of type-hinting on the `AbstractService` constructor.


richardm-stripe:
Hi @angsmugmug, thanks for the contribution! Would you mind running `make fmt` so that the build will pass CI and merging becomes possible?

Also would you mind detailing your use case a little bit more? So far we don't use any type-hints, opting instead for type annotations in PHPDoc. I have some hesitation about this -- we don't consider the shape of the `stripe-php` classes under reflection to be part of the public interface of the library, and if your code requires a type hint in order to work properly, that means it's dependent on the shape of our library under reflection, which I'm not sure we want to encourage, especially if it's possible to accomplish the same sort of thing with a wrapper. In any case I would be curious to know a little bit more about how you are using this and what it enables for you.

angsmugmug:
Hi @richardm-stripe, cleaned up via `make fmt`.

Yes, the usage pattern we were trying was to utilize the individual service classes in `stripe-php`, such as `\Stripe\Service\PaymentIntentService`, as the dependency of our own wrapper classes...

```php
// Pattern A
class MyCustomPaymentService
{
    public function __construct(\Stripe\Service\PaymentIntentService $stripePaymentIntentService) { ... }
}

class MyCustomCustomerService
{
    public function __construct(\Stripe\Service\CustomerService $stripeCustomerService) { ... }
}
```

... as opposed to using the `\Stripe\StripeClient` client class as the dependency... 

```php
// Pattern B
class MyCustomPaymentService
{
    public function __construct(\Stripe\StripeClient $stripeClient)
    {
        $stripePaymentIntentService = $stripeClient->paymentIntents;
        ...
    }
}

class MyCustomCustomerService
{
    public function __construct(\Stripe\StripeClient $stripeClient)
    {
        $stripeCustomerService = $stripeClient->customers;
        ...
    }
}
```

It looks like the `\Stripe\StripeClient` class acts as a single entry point for the services, by utilizing the `\Stripe\Service\CoreServiceFactory` to instantiate the service class dynamically.

The reason we wanted to go with pattern (A) was our desire to create service wrapper classes around the `stripe-php` services we are utilizing.  These wrapper classes would then be consumed by our application code, creating a boundary between `stripe-php` and application code.  Pattern (A) seemed to match that approach best since it was a one-to-one mapping between service wrapper and `stripe-php` service class.  This meant that the responsibility of Stripe API communication was left to the `\Stripe\StripeClient`, thus not utilizing `\Stripe\StripeClient` as a service locator/factory/single entry point class.

Toronto00 (approved):

richardm-stripe:
Hello, sorry for the long silence on this. We are considering adding type hints to this library as part of a general effort, and are considering the tradeoffs of that separately. I'm closing this PR though, since it's unlikely that we will merge a PR adding type hints to a specific place outside of a general effort.

If you do see this, I am still curious about your use case though. Can you explain slightly more about how these wrappers create a boundary between stripe-php and your application and what this boundary protects you from?

In some ways, it seems like splitting off services from StripeClient is *reducing* the boundary between your application and stripe-php, because it is making you dependent on very specific details of `stripe-php`'s guts, in this case the presence/absence of a type hint inside the argument to the constructor of the `AbstractService` class, and on the specific types of the classes returned by e.g. `$stripe->paymentIntents`. What benefit do you achieve from splitting off the services that counterweighs this?

PR #1027, opened 2020-10-14
Warn if opts are in params

vcheung-stripe:
### Summary
Warn the user if they mistakenly put options into the `$params` argument to a method, similar to the [Stripe Node warning](https://github.com/stripe/stripe-node/blob/master/lib/utils.js#L128-L132). For example, we warn when the user includes an `api_key` in `$params`.

### Motivation
https://jira.corp.stripe.com/browse/DX-6065

### Testing
Unit tests

### DIFF
--- a/tests/Stripe/BaseStripeClientTest.php
+++ b/tests/Stripe/BaseStripeClientTest.php
@@ -178,4 +178,27 @@ final class BaseStripeClientTest extends \PHPUnit\Framework\TestCase
         $client = new BaseStripeClient(['api_key' => 'sk_test_client', 'api_base' => MOCK_URL]);
         $client->requestCollection('get', '/v1/charges/ch_123', [], []);
     }
+
+    public function testRequestWithOptsInParamsWarns()
+    {
+        $this->expectException(\PHPUnit_Framework_Error_Warning::class);
+
+        $client = new BaseStripeClient([
+            'api_key' => 'sk_test_client',
+            'stripe_account' => 'acct_123',
+            'api_base' => MOCK_URL,
+        ]);
+        $charge = $client->request(
+            'get',
+            '/v1/charges/ch_123',
+            [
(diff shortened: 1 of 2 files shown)


vcheung-stripe:
First time writing PHP, so not sure if I'm following the right conventions or even putting code in the write place. Would love nitpicks.

tests/Stripe/BaseStripeClientTest.php
```diff
@@ -178,4 +178,27 @@ public function testRequestCollectionThrowsForNonList()
         $client = new BaseStripeClient(['api_key' => 'sk_test_client', 'api_base' => MOCK_URL]);
         $client->requestCollection('get', '/v1/charges/ch_123', [], []);
     }
+
+    public function testRequestWithOptsInParamsWarns()
+    {
+        $this->expectException(\PHPUnit_Framework_Error_Warning::class);
```
richardm-stripe, line 184: Any way to assert on the error message?

richardm-stripe (approved):
Fantastic!

richardm-stripe:
Since you asked for nits: sometimes I prefer a `foreach` vs. `\array_filter` in PHP when I would do the opposite in e.g. Javascript because lambdas have a heavy syntax. I also lean towards variable interpolation rather than `\sprintf`, but there are precedents for both `\array_filter` and `\sprintf` already in stripe-php and would not necessarily change what you have.

vcheung-stripe:
> Since you asked for nits: sometimes I prefer a `foreach` vs. `\array_filter` in PHP when I would do the opposite in e.g. Javascript because lambdas have a heavy syntax. I also lean towards variable interpolation rather than `\sprintf`, but there are precedents for both `\array_filter` and `\sprintf` already in stripe-php and would not necessarily change what you have.

Will keep what I have, but good to know.

PR #1038, opened 2020-11-06
Update PHPDocs

remi-stripe:
Codegen for openapi 83680bb

r? @ctrudeau-stripe 
cc @stripe/api-libraries


lib/ErrorObject.php
```diff
@@ -23,6 +23,9 @@
  *    returned on a request involving a PaymentIntent.
  * @property PaymentMethod $payment_method The PaymentMethod object for errors
  *    returned on a request involving a PaymentMethod.
+ * @property string $payment_method_type If the error is specific to the type
```
ob-stripe, line 26: This also applies to other existing fields, but presumably this is nullable?

ob-stripe (approved):

lib/ErrorObject.php
remi-stripe: Yeah the others are nullable too and it's because this one is not automatically generated but I'm tracking internally

PR #1039, opened 2020-11-06
Bump PHP-CS-Fixer to 2.16.5

ob-stripe:
r? @remi-stripe 

Bump PHP-CS-Fixer to 2.16.5 and run `make fmt`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
     "php-coveralls/php-coveralls": "^2.1",
     "squizlabs/php_codesniffer": "^3.3",
     "symfony/process": "~3.4",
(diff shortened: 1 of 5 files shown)


remi-stripe (approved):
Approving but are you sure about the `\get_class` change being 100% safe?

ob-stripe:
> Approving but are you sure about the \get_class change being 100% safe?

As much as I can be. It's only used in error messages so the absolute worst case is that we get an error while returning an error :) (But I don't think that's likely.)

ob-stripe:
Oh, `Account.php` and `Source.php` are codegen'd so we'll need to update the manual overrides I guess. cc @richardm-stripe @ctrudeau-stripe

PR #1050, opened 2020-11-19
Added proration behavior constants

nickdnk:
Hello

I just added some constants that I couldn't find anywhere. I'd say these belong on Subscription, but a case could be made for SubscriptionSchedule as well.

### DIFF
--- a/lib/Subscription.php
+++ b/lib/Subscription.php
@@ -58,6 +58,10 @@ class Subscription extends ApiResource
     const PAYMENT_BEHAVIOR_ERROR_IF_INCOMPLETE = 'error_if_incomplete';
     const PAYMENT_BEHAVIOR_PENDING_IF_INCOMPLETE = 'pending_if_incomplete';
 
+    const PRORATION_BEHAVIOR_ALWAYS_INVOICE = 'always_invoice';
+    const PRORATION_BEHAVIOR_CREATE_PRORATIONS = 'create_prorations';
+    const PRORATION_BEHAVIOR_NONE = 'none';
+
     const STATUS_ACTIVE = 'active';
     const STATUS_CANCELED = 'canceled';
     const STATUS_INCOMPLETE = 'incomplete';


remi-stripe:
@nickdnk Thanks for submitting! The stripe-php is now automatically generated from our openapi spec so we will have to make some changes internally before we can merge the change which will take a bit of time but I'll flag!

nickdnk:
> @nickdnk Thanks for submitting! The stripe-php is now automatically generated from our openapi spec so we will have to make some changes internally before we can merge the change which will take a bit of time but I'll flag!

I thought this was only for phpdocs, but okay sure. No rush.

remi-stripe:
No the entire library is not automatically generated so every class, methods and constants now come from the openapi spec!

nickdnk:
> No the entire library is not automatically generated so every class, methods and constants now come from the openapi spec!

Do you accept PRs in the OpenAPI document then? 😬

remi-stripe:
> Do you accept PRs in the OpenAPI document then? 😬
The openapi spec is automatically generated from our code, otherwise it would defeat the purpose!

nickdnk:
> > Do you accept PRs in the OpenAPI document then? 😬
> > The openapi spec is automatically generated from our code, otherwise it would defeat the purpose!

Well, yes and no. Generating libraries in multiple languages from one OpenAPI doc _would_ serve a purpose, but generating it from code is obviously the best solution. It was more of a joke anyway.

I suppose "my work here is done" then. Or do you still want PRs such as these to bring stuff to attention?

remi-stripe:
@nickdnk We definitely still accept your PRs and someone internally has made the changes to match your PR right now so that we can merge it soon. You can also file an issue instead to ask for extra constants that you want if you don't want to do the PR and we would do it for you!

While we do constants, we mostly matched what you had already done! Constants are still dangerous since they can disappear/change in API versions and the stripe-php library supports all API versions. In practice it'd mean losing constants in minor version (which we don't want) or releasing a major each time we remove a constant (which is costly). For that reason, we don't generate constants for every single enum values today though we likely will in the future!

(summary: please keep the PRs coming if you're willing :)

nickdnk:
> @nickdnk We definitely still accept your PRs and someone internally has made the changes to match your PR right now so that we can merge it soon. You can also file an issue instead to ask for extra constants that you want if you don't want to do the PR and we would do it for you!
> 
> While we do constants, we mostly matched what you had already done! Constants are still dangerous since they can disappear/change in API versions and the stripe-php library supports all API versions. In practice it'd mean losing constants in minor version (which we don't want) or releasing a major each time we remove a constant (which is costly). For that reason, we don't generate constants for every single enum values today though we likely will in the future!
> 
> (summary: please keep the PRs coming if you're willing :)

Okay.

I'd make the case that PHPDoc and deprecation notices should be able to clear up any problems with constants. If you upgrade your API version on Stripe and this has a breaking change of an enumerated value, I'd expect any serious developer to check their code and make the necessary changes, updating the Stripe PHP library in the progress. But I also know we don't live in a perfect world. I like constants because it makes it easier to find uses of enumerated values when refactoring code while avoiding bugs due to typos.

richardm-stripe:
Merging this! Thanks so for the contribution @nickdnk.

I've confirmed that `git remote add nickdnk https://github.com/nickdnk/stripe-php && git diff nickdnk/master` displays no diff for me against the latest generated code.

richardm-stripe:
Released in [7.65.0](https://github.com/stripe/stripe-php/releases/tag/v7.65.0).

PR #1063, opened 2020-12-15
Multiple API changes

ctrudeau-stripe:
Codegen for openapi cee0955.
r? @brandur-stripe
cc @stripe/api-libraries

## Changelog
* Updates PHP docstrings to match the latest version of the API
* Adds support for 'account_requirement' enum value on `File.purpose`;


brandur-stripe (approved):
Thanks! LGTM.

PR #1086, opened 2021-02-03
Update CA cert bundle URL

FPCSJames:
The curl domain has changed from curl.haxx.se to curl.se. Calling update_certs.php with the old CA bundle URL may result in the text of the 301 response being downloaded instead of the bundle file, breaking communication with Stripe.


richardm-stripe (approved):
Thanks for the contribution. Confirmed that the redirect exists, so we should merge this.

PR #1089, opened 2021-02-09
API Updates

richardm-stripe:
Codegen for openapi 04fba1e.
r? @richardm-stripe
cc @stripe/api-libraries

## Changelog
* Added support for `afterpay_clearpay_payments` on `Account.capabilities`
* Added support for `payment_settings` on `Invoice`


remi-stripe (approved):
Technically we are supposed to have constants for all capabilities in https://github.com/stripe/stripe-php/blob/master/lib/Account.php#L55 but they are ~all missing so not urgent

PR #1090, opened 2021-02-11
CI: github actions pipeline

richardm-stripe:
@ob-stripe had put together a prototype GHA pipeline, this PR builds upon that. I think this is ~production-ready.

- Stopped using the PHPStan docker image. This was breaking because the latest docker image there uses PHP 8 which we cannot use in CI yet because our version of PHPUnit is too old.

- Removed the dependency of `make phpstan` on stripe mock.

- Added a step to persist the composer cache. Should speed up CI runs.

- Changed
```diff
- on: [push, pull_request]
+ on: 
+   push:
+     branches: [ master ]
+   pull_request:
+     branches: [ master ]
```

The previous thing results in running every job twice when you push to any PR. The new thing runs on every push to a pull request against master, and on the master branch after a push.


.github/workflows/ci.yml
```diff
@@ -41,11 +41,12 @@ jobs:
...
-          - "5.6"
-          - "7.0"
-          - "7.1"
-          - "7.2"
-          - "7.3"
+          # - "5.6"
+          # - "7.0"
+          # - "7.1"
+          # - "7.2"
+          # - "7.3"
```
richardm-stripe, line 48:
```suggestion
          blueohueloaiheohaundieoa
```

ob-stripe:
> The previous thing results in running every job twice when you push to any PR.

This is the existing behavior with Travis. I agree that running jobs twice is unnecessary, but we definitely want to run the jobs once for every PR, otherwise how do we know that the PR is safe to merge?

richardm-stripe:
@ob-stripe I think this configuration does that. `on.pull_request.branches.master` this guarantees that CI will have run on the latest commit for every pull request (whose base is master) and then `on.push.master` makes sure that after we merge to master, CI runs.

ob-stripe (approved):
LGTM!

PR #1094, opened 2021-02-19
API Updates

richardm-stripe:
Codegen for openapi 403fdb8.
r? @richardm-stripe
cc @stripe/api-libraries

## Changelog
* Added support for new resource `BillingPortal.Configuration`
* Added support for `configuration` and `on_behalf_of` on `Session#create` and `BillingPortal.Session`
* `WebhookEndpoint#create.enabled_events[]` and `WebhookEndpoint#update.enabled_events[]` added new enum members: `billing_portal.configuration.created and billing_portal.configuration.updated`


tests/Stripe/Service/BillingPortal/ConfigurationServiceTest.php
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * @internal
+ * @covers \Stripe\Service\BillingPortal\ConfigurationService
+ */
+final class ConfigurationServiceTest extends \PHPUnit\Framework\TestCase
+{
+    use \Stripe\TestHelper;
+
+    const TEST_RESOURCE_ID = 'cs_123';
```
remi-stripe, line 13: id should be bpc_123

remi-stripe (approved):
one nitpick and I would have recommended adding test for the not service version but it's alll codegen so I won't block

tests/Stripe/Service/BillingPortal/ConfigurationServiceTest.php
richardm-stripe, line 13:
```suggestion
    const TEST_RESOURCE_ID = 'bpc_123';
```

PR #1103, opened 2021-03-29
Run CI on php8?

richardm-stripe:


richardm-stripe:
Closing for now, might revive later.

PR #1116, opened 2021-04-22
Fix property annotation on Invoice object

driesvints:
The current annotation types are invalid and not parsed by IDE's like PhpStorm. I've edited the ones I could find in the `Invoice.php` file that I was currently editing but more could be present in the rest of the codebase.

Here's a screenshot of PhpStorm not recognising this syntax:

Screenshot 2021-04-22 at 22 14 41

### DIFF
--- a/lib/Invoice.php
+++ b/lib/Invoice.php
@@ -45,7 +45,7 @@ namespace Stripe;
  * @property string $object String representing the object's type. Objects of the same type share the same value.
  * @property null|string $account_country The country of the business associated with this invoice, most often the business creating the invoice.
  * @property null|string $account_name The public name of the business associated with this invoice, most often the business creating the invoice.
- * @property null|(string|\Stripe\TaxId)[] $account_tax_ids The account tax IDs associated with the invoice. Only editable when the invoice is a draft.
+ * @property null|string[]|\Stripe\TaxId[] $account_tax_ids The account tax IDs associated with the invoice. Only editable when the invoice is a draft.
  * @property int $amount_due Final amount due at this time for this invoice. If the invoice's total is smaller than the minimum charge amount, for example, or if there is account credit that can be applied to the invoice, the <code>amount_due</code> may be 0. If there is a positive <code>starting_balance</code> for the invoice (the customer owes money), the <code>amount_due</code> will also take that into account. The charge that gets generated for the invoice will be for the amount specified in <code>amount_due</code>.
  * @property int $amount_paid The amount, in %s, that was paid.
  * @property int $amount_remaining The amount remaining, in %s, that is due.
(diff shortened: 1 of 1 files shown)


driesvints:
Failing tests seem unrelated.

remi-stripe:
@driesvints Thanks for the PR! Technically the phpdoc we have is correct and follows the [official spec](https://docs.phpdoc.org/3.0/guide/references/phpdoc/types.html#arrays). This means that this is likely something missing in PHPStorm instead. Might be worth sending them a bug report.

We're looking into improving this on our end though since we never return both a string and an object in the array so your alternative is correct in this case (though we have cases where it wouldn't be). But those files are automatically generated on our end from our own code/openapi spec so it'll require some changes that will affect other files too. Does that make sense?

driesvints:
I see, thanks for making me aware of this! I'll close this one off.

driesvints:
@remi-stripe JetBrains have made it clear that they won't support this syntax. By knowing that the largest PHP IDE won't support this syntax would you re-consider changing to their recommended syntax (the one from this PR)?

https://youtrack.jetbrains.com/issue/WI-60112#focus=Comments-27-4856351.0-0

remi-stripe:
@driesvints We're looking into potentially changing this internally but we won't get to it for a while I think as it's been like this for a while now and it hasn't been raised before. It's on our radar though.

That response from Jetbrains is surprising though. Using `null|int[]|string[]` is not technically correct. That prevents you from having `['a', 'b', 33, 'c', 756]` for example. We have cases in our API where we return an array of polymorphic objects like `[<Card>, <BankAccount>,<Card>]` in which case we can't switch to `null|Card[]|BankAccount[]` as the types would be incorrect. It seems like a misunderstanding of the ask on their end if they think it's just cosmetic when really it changes the types entirely and the corresponding potential static analysis. I'd recommend pushing a bit more on them to ensure they understand the scope.

driesvints:
@remi-stripe fully understandable and definitely a correct point you raise. I'll respond to the issue.

driesvints:
@remi-stripe the person who replied on the issue notes that a different syntax is possible atm: https://youtrack.jetbrains.com/issue/WI-20193#focus=Comments-27-4438701.0-0

```php
 * @property null|array<string, \Stripe\TaxId> $account_tax_ids
```

I believe that would achieve the same thing and it might be more adopted amongst IDE's?

remi-stripe:
@driesvints Does it work on PHPStorm already if you do this? Technically we could do that and hope most IDEs follow this. Ultimately using `array<.....>` feels more wordy and identical to what we're currently doing though which is in the spec.

driesvints:
Seems to work! I agree that I think it's best that PhpStorm also supports the other syntax but since they're not going to do that it might be best to switch to the syntax supported by the most used IDE in the ecosystem? 

Thanks for your patience btw! 🙂 

Screenshot 2021-04-23 at 17 46 50

remi-stripe:
I'll keep the PR as closed, but we are looking into a potential improvement in the future so we're tracking the recommendation you have here since it seems to work fine for PHPStorm

driesvints:
Thanks @remi-stripe

PR #1143, opened 2021-06-18
[WIP] binary streaming support

richardm-stripe:


lib/HttpClient/CurlClient.php
```diff
@@ -275,16 +275,193 @@ public function request($method, $absUrl, $headers, $params, $hasFile)
...
+        $headerCallback = function ($curl, $header_line) use (&$rheaders) {
+            // Ignore the HTTP request line (HTTP/1.1 200 OK)
+            if (false === \strpos($header_line, ':')) {
+                return \strlen($header_line);
+            }
+            list($key, $value) = \explode(':', \trim($header_line), 2);
+            $rheaders[\trim($key)] = \trim($value);
+
+            return \strlen($header_line);
+        };
```
dcr-stripe, line 330: Wonder if we can make a function that generates this callback given a `headers` param. Would be nice if we could avoid duplicating with the code below.

lib/HttpClient/CurlClient.php
```diff
@@ -275,16 +275,193 @@ public function request($method, $absUrl, $headers, $params, $hasFile)
...
+                    // Don't expose the $curl handle to the user, and don't require them to
+                    // return the length of $data.
+                    \call_user_func_array($readBodyChunk, [$data]);
+
+                    return \strlen($data);
+                };
+            }
+            $succeeded = false;
+
+            $shouldRetry = $this->shouldRetry($errno, $rcode, $rheaders, $numRetries);
```
dcr-stripe, line 411: We're missing some code here from the original that I wonder if we need to keep.. chiefly:

```
            if (!$this->getEnablePersistentConnections()) {
                $this->closeCurlHandle();
            }
```

Do we need to respect that case when streaming? I know you reset it below.

tests/Stripe/HttpClient/CurlClientTest.php
```diff
@@ -352,4 +353,120 @@ public function testSetRequestStatusCallback()
...
+    {
+        $this->stopTestServer();
+    }
+
+    // This is a pretty flaky/uncertain way to try and get an
+    // http server to deliver the body in separate "chunks".
+    //
+    // It seems to work but feel free to just skip or delete
+    // if this flakes.
+    public function testExecuteRequestWithRetriesCallsWriteFunctionWithChunks()
```
dcr-stripe, line 370: I wonder if there's something we can do to annotate these tests as more "heavy" or integration tests? I imagine their runtime is longer than standard unit tests. No real proposed change here, just commenting in case you know of something.

tests/TestServer.php
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+     * Makes a directory in a temporary path containing only an `index.php` file with
+     * the specified content ($code).
+     *
+     * @param mixed $code
+     */
+    private function makeTemporaryServerDirectory($code)
+    {
+        $dir = \sys_get_temp_dir() . \DIRECTORY_SEPARATOR . 'stripe-php-test-server';
```
dcr-stripe, line 35: Thoughts on adding an extra qualifier to this to make sure we don't accidentally clash throughout multiple test runs (eg. timestamp)? Can also help inspect the generated code to make sure all looks fine.

tests/TestServer.php
```diff
@@ -0,0 +1,163 @@
...
+        $indexPhp = $dir . \DIRECTORY_SEPARATOR . 'index.php';
```
dcr-stripe, line 36: Tiny Nit: Elsewhere we use PHP as the capitalization. Rename to indexPHP to be consistent (Eg. isPHPTestServerRequestLogLine)

lib/HttpClient/CurlClient.php
```diff
@@ -275,16 +275,193 @@ public function request($method, $absUrl, $headers, $params, $hasFile)
...
+        };
+
+        while (true) {
+            list($headerCallback, $writeCallback) = $this->useHeadersToDetermineWriteCallback($opts, $determineWriteCallback);
+            $opts[\CURLOPT_HEADERFUNCTION] = $headerCallback;
+            $opts[\CURLOPT_WRITEFUNCTION] = $writeCallback;
+
+            $shouldRetry = false;
+            $rbody = null;
+            $succeeded = null;
```
dcr-stripe, line 438: We don't actually seem to use `$succeeded` anywhere - am I missing something? Maybe we no longer need?

tests/Stripe/HttpClient/CurlClientTest.php
```diff
@@ -352,4 +353,120 @@ public function testSetRequestStatusCallback()
...
+    {
+        $chunk1 = 'First, bytes';
+        $chunk2 = 'more bytes';
+        $chunk3 = 'final bytes';
+        $serverCode = <<<EOF
+<?php
```
dcr-stripe, line 376: I'm trying to think through if there are better ways we can provide the code than a raw string... but haven't figured anything out yet. Maybe we can have the test code be in fixture files? Mainly I'm just worried that we might have some kind of syntax error or something in here that ends up hiding an issue. Not sure what would be better though...

dcr-stripe (commented):
Overall I think this looks good and makes sense. Left a few comments. Thanks for putting together the test infrastructure! Curious what @ob-stripe thinks as well.

lib/HttpClient/CurlClient.php
richardm-stripe: Hmm. I do like a little bit of duplication because the control flow is weird and imo harder to follow if it's not inline -- so I took half your suggestion and moved the *body* into a helper but left the weird control flow inline.

tests/TestServer.php
```diff
@@ -0,0 +1,163 @@
...
+        $pipes = [];
+        $this->serverProc = \proc_open(
+            $command,
+            [2 => ['pipe', 'w']],
+            $pipes
+        );
+
+        $pid = \proc_get_status($this->serverProc)['pid'];
+
+        // echo "Started test server on pid $pid\n";
```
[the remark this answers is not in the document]
richardm-stripe: Removed.

lib/HttpClient/CurlClient.php
richardm-stripe: Great point. I think so, I added support for this which required me to shuffle around where `$errno` and `$message` were set. I also added some test cases.

lib/HttpClient/CurlClient.php
richardm-stripe: Great catch! Yes no longer needed!

tests/Stripe/HttpClient/CurlClientTest.php
richardm-stripe: I added a step to run `php -l` on the raw strings - this should catch syntax errors.

Don't want to do the fixtures, as that involves separating the test logic in the server from the test logic in the test and makes things harder to read/change.

It's a bit better in the similar [stripe-node](https://github.com/stripe/stripe-node/blob/7e1b9f7c5a17e4fee00c1eee43e83f6a26c29a31/test/StripeResource.spec.js#L189) [test server](https://github.com/stripe/stripe-node/blob/9b58b95e2fbd0296ceb43aa6b1854785edbc3561/testUtils/index.js#L14) where you can control server and client from the same process. Don't know how to accomplish anything similar in PHP.

tests/Stripe/HttpClient/CurlClientTest.php
richardm-stripe: PHPUnit has "test suites" https://phpunit.readthedocs.io/en/9.5/organizing-tests.html, but I don't think it's time for that yet.

I tuned the retry timeouts to bring the test suite down from 4.5s to 1.5s (still mildly distressing, it was sub 1s before)

tests/TestServer.php
richardm-stripe: Good thought - I'd like to solve the problem when it arises.

dcr-stripe (approved):
LGTM! Thanks Richard! 🚢

ob-stripe (approved):
LGTM as well.

Not a huge fan of having to spin up a new process in tests for the test server, but it works and I can't think of a better way.

PR #1145, opened 2021-06-25
API Updates

richardm-stripe:
Codegen for openapi d12f3f1.
r? @richardm-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `boleto` on `PaymentMethod`.
* Add support for `il_vat` as a member of the `TaxID.Type` enum.


dcr-stripe (approved):
LG! 🎉

PR #1151, opened 2021-07-06
Add `mode` constants into Checkout\Session

TeleginS:
Hi guys. 
I would like to add these constants. I don't really want to keep them on my side in the project.
according [Checkout Session API doc](https://stripe.com/docs/api/checkout/sessions/object#checkout_session_object-mode)


richardm-stripe:
Hello @TeleginS  - thanks for the contribution!

PR #1159, opened 2021-07-13
[RFC] Add PHPDoc generics to Collection

franmomu:
Hi, this is just to see if it's fine, then I'll continue. Nowadays many tools support this generic `@template` annotation, this would improve the developer experience allowing to specify which types a `Collection` holds.

I've only used in `InvoiceService` to see how would look like.

### DIFF
--- a/lib/Collection.php
+++ b/lib/Collection.php
@@ -9,6 +9,9 @@ namespace Stripe;
  * @property string $url
  * @property bool $has_more
  * @property \Stripe\StripeObject[] $data
+ *
+ * @template TStripeObject of StripeObject
+ * @template-implements \IteratorAggregate<TStripeObject>
  */
 class Collection extends StripeObject implements \Countable, \IteratorAggregate
 {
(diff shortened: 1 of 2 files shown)


remi-stripe:
@franmomu Thanks for the example, this is not something I had seen before. I'll assign to @richardm-stripe to look into the change and see if it's something we'd want to add. Since the library is automatically generated from code on our end, we'd be able to tweak this for the whole library if we end up aligning!

richardm-stripe:
@franmomu - I think this is a great idea.

When you say
> Nowadays many tools support this generic

I am curious which tools you are referring to? I am aware of PHPStan and Psalm but I wonder if others should be on our radar.

franmomu:
> I am curious which tools you are referring to? I am aware of PHPStan and Psalm but I wonder if others should be on our radar.

I was thinking about those two, [Phan](https://github.com/phan/phan) and [the new phpstorm coming version](https://blog.jetbrains.com/phpstorm/2021/07/phpstorm-2021-2-beta/).

So, based on:

> Since the library is automatically generated from code on our end, we'd be able to tweak this for the whole library if we end up aligning!

is this something you should do? or can I help with something

PR #1166, opened 2021-07-27
API Updates

richardm-stripe:
Codegen for openapi 1d88c6a.
r? @richardm-stripe
cc @stripe/api-libraries

## Changelog
* Add support for new value `redacted` on enum `Review.closed_reason`


dcr-stripe:
Going to close this out since we're re-generating this and it's a no-op.

PR #1175, opened 2021-08-24
API Updates

richardm-stripe:
Codegen for openapi 51de6c5.
r? @dcr-stripe
cc @stripe/api-libraries


remi-stripe:
No change so likely a simple close

PR #1178, opened 2021-09-08
API Updates

dcr-stripe:
Codegen for openapi 95a1ab8.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `livemode` on `Reporting.ReportType`


dcr-stripe:
Not sure if we want to merge this or wait for next release since it's just adding a const.

PR #1186, opened 2021-10-06
Add generated test suite.

dcr-stripe:
### Notify

r? @richardm-stripe 

### Summary

Adds auto-generated test suite. This test suite tries to invoke each method against stripe-mock.


dcr-stripe:
This isn't quite ready yet based on the PHPStan results. Will re-create once we're good to go.

PR #1199, opened 2021-11-05
Run tests for PHP 8

hivokas:


hivokas:
Well, I suppose it's not that easy because the codebase doesn't support PHP 8 :(

remi-stripe:
@hivokas Correct it's unfortunately not easy to add support for this. All the tools we use are either locked on older versions or require dropping older versions of PHP to support PHP 8. While it's something we really want to offer, it'd mean dropped PHP 5.6 and PHP 7.0 which are unfortunately still widely used by developers using Stripe today.

hivokas:
@remi-stripe I've found this branch in this repo: https://github.com/stripe/stripe-php/commit/84c077c39c84bd867bfd98c56034faf5b667ea66. 

I believe test cases can be updated to be supported both by PHP 5.6 and PHP 8, both by PHPUnit 5.7 and PHPUnit 9 by not using asserts that are missing either in PHPUnit 5.7 or PHPUnit 9. Then, `composer.json` can be updated to use PHPUnit `^5.7|^9.0` which will allow GitHub Actions CI workflow to use PHPUnit 5.7 for older PHP versions, PHPUnit 9 for PHP 8. Does it make sense?

remi-stripe:
cc @richardm-stripe can you have a look at the PR and see whether this is a viable path forward?

driesvints:
@remi-stripe I still think that just branching to a new major version is the best way to go to modernize the SDK and drop a bunch of old PHP versions. You can simultaneously support the new major version and the old one so nobody gets left behind.

lucasmichot:
Definitely agree with @driesvints
cf. https://www.php.net/supported-versions.php

richardm-stripe:
This was subsumed by #1209, which began running CI for both PHP8 and PHP8.1.

hivokas:
This is great @richardm-stripe! Thanks!

PR #1201, opened 2021-11-11
API Updates

richardm-stripe:
Codegen for openapi b428029.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
* Add support for new resource `ShippingRate`
* Add support for `shipping_options` and `shipping_rate` on `Checkout.Session`


dcr-stripe (approved):
LGTM!

PR #1206, opened 2021-11-19
API Updates

richardm-stripe:
Codegen for openapi 08ff630.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
* Add support for new value `jct` on enum `TaxRate.tax_type`


dcr-stripe (approved):
LGTM

PR #1209, opened 2021-11-29
Support testing against PHP 8 and 8.1.

dcr-stripe:
## Summary
This PR:
* Runs CI for PHP 8.0 and PHP 8.1.
* Changes `composer.json` to require (as a devDependency) EITHER phpunit 5.7, which is compatible with PHP <=7.2; OR phpunit 9.0, which is compatible with PHP 8.
  * Some functions in phpunit 9.0 don't exist in phpunit 5.7. To keep our test suite compatible with both, we extend the PHPUnit `TestCase` class and add some "shims" over top of them.
* Fixes the flaky StripeTelemetryTest
* Updates the php-cs-fixer task to run on PHP8
  * Towards that end, upgrade php-cs-fixer to the recent version 3.2.1 (`|| 2.17.1`  is provided as a fallback so composer install doesn't break on old versions of PHP)
  * Make small formatting/style changes associated with that upgrade.

fixes #1200 #1199


richardm-stripe:
@dcr-stripe will you take a look at this?

.github/workflows/ci.yml
```diff
@@ -19,22 +19,28 @@ jobs:
...
   phpstan:
     name: PHPStan
 
     runs-on: ubuntu-latest
 
+    strategy:
+      fail-fast: false
+      matrix:
+        php-version:
+          - "7.4"
```
dcr-stripe, line 35: Should we run this on 8.1 as well?

composer.json
```diff
@@ -21,11 +21,10 @@
     "ext-mbstring": "*"
   },
   "require-dev": {
-    "phpunit/phpunit": "^5.7",
+    "phpunit/phpunit": "^5.7 || ^9.0",
     "php-coveralls/php-coveralls": "^2.1",
     "squizlabs/php_codesniffer": "^3.3",
-    "symfony/process": "~3.4",
```
dcr-stripe, line 27: Confirming this was just a stale require?

dcr-stripe (commented):
This looks good to me overall! Thanks for all your hard work on this Richard! 🎉 

(I can't approve since it's my own PR)

.github/workflows/ci.yml
richardm-stripe: Yes, I have that in a forthcoming PR. There are some warnings I need to address first.

composer.json
richardm-stripe: Good call. We actually use it in `StripeMock.php` which apparently attempts to boot up stripe-mock if you aren't already running it?

In our CI we always boot up stripe-mock in a separate step, and I always boot stripe-mock up manually, so I went ahead and removed this file.

dcr-stripe: LGTM - good find!

hivokas:
This is awesome, thanks for your work!

PR #1211, opened 2021-11-30
Update PHPStan

richardm-stripe:
## Notify
r? @dcr-stripe 
## Summary
* Upgrades PHPStan to the latest version, for compatibility with php8
* Moves phpstan  into `composer.json` as a devDependency  rather than being installed manually by the Makefile.
* Adds a step to the CI jobs to remove phpstan from composer.json from all jobs but the PHPStan job.
* Adds some annotations to suppress new PHPStan errors.
* Fixes a couple return types.
* Moves some squelches from the "baseline" file into inline squelches as comments -- a new feature in the latest PHPStan and my preferred way to squelch.
* Removes coveralls.


Makefile
```diff
@@ -4,10 +4,6 @@ export PHPSTAN_VERSION := 0.12.59
 vendor: composer.json
```
dcr-stripe, line 4: Probably want to remove the PHPSTAN_VERSION export above.

composer.json
```diff
@@ -24,7 +24,8 @@
     "phpunit/phpunit": "^5.7 || ^9.0",
     "php-coveralls/php-coveralls": "^2.1",
     "squizlabs/php_codesniffer": "^3.3",
-    "friendsofphp/php-cs-fixer": "3.2.1 || 2.17.1"
+    "friendsofphp/php-cs-fixer": "3.2.1 || 2.17.1",
+    "phpstan/phpstan": "^1.2"
```
dcr-stripe, line 28: Will this cause issues installing the package with dev dependencies on earlier versions if we're saying it's not compatible with them?

lib/Customer.php
```diff
@@ -78,6 +78,8 @@ public function deleteDiscount($params = null, $opts = null)
         $url = $this->instanceUrl() . '/discount';
         list($response, $opts) = $this->_request('delete', $url, $params, $opts);
         $this->refreshFrom(['discount' => null], $opts, true);
+
+        return $this;
```
dcr-stripe, line 82: This will need a codegen change right?

lib/Quote.php
```diff
@@ -66,8 +66,6 @@ class Quote extends ApiResource
      * @param null|array|string $opts
      *
      * @throws \Stripe\Exception\ApiErrorException if the request fails
-     *
-     * @return \Stripe\File the created file
```
dcr-stripe, line 70: Flagging this will also need a codegen change.

composer.json
richardm-stripe: Yes. I think we should go with a policy of `composer install` is only supported on php 7.3+. I think this policy is fine because the only users of `composer install` are
* CI - which we have already handled with the `composer remove` workaround in this PR.
* maintainers/contributors to stripe-php:
  * I am happy to be restricted to php7.3+ when doing everyday development on stripe-php
  * When I am explicitly trying to test compatibility things with <=php7.2 it will be definitely an inconvenience to not be able to `composer install` without manually working around, and beyond that it will be inconvenient to not be able to utilize `phpstan/php-cs-fixer`, but I think this inconvenience is far outweighed by the additional complexity it would take to continuously maintain compatibility with multiple versions of PHPStan and php-cs-fixer.

dcr-stripe: Sounds good - let's leave a comment explaining this and steps for running <php 7.2 so that this doesn't come back to haunt us down the road once we've forgotten :)

dcr-stripe (approved):
LGTM! Let's:
- update the PR description to mention coverall removal
- update the PR title to no longer be WIP.

PR #1182, opened 2021-09-18
Fix PHP 8.1 compatibility with new interface types

kylekatarnls:

PR #1213, opened 2021-12-01
[PHP 8.1] Add missing ReturnTypeWillChange

spaze:
These were originally added in #1182 (now closed) but missed in #1211 (the one replacing the former PR).


spaze:
I've now diffed  #1182 & #1211 and these should be all that were missed.

richardm-stripe:
Thanks for the contribution @spaze!

PR #1220, opened 2021-12-15
API Updates

yejia-stripe:
Codegen for openapi 64b4f72.
r? @richardm-stripe
cc @stripe/api-libraries

## Changelog
* Add support for new resource `PaymentIntentTypeSpecificPaymentMethodOptionsClient`


richardm-stripe (approved):
Approved, this is just docstrings though without any new fields, so we should merge but not release.

PR #1228, opened 2021-12-24
Set subscription pause behavior constants

yaroslawww:
For convenience and consistency, it is useful to have constants for pause behavior.


richardm-stripe:
Hello @yaroslawww. We can't merge this PR right now, as our code-generator only describes types that appear top-level resources `Subscription.*` and this enum appears two levels down `Subscription.pause_collection.behavior`. We'd like to support this eventually, but would likely do this as a library-wide effort rather than adding individual enums.

Would you be willing to re-open this as a "feature request"-labeled issue vs. a PR?

PR #1233, opened 2022-01-19
API Updates

richardm-stripe:
Codegen for openapi 4d2f301.
r? @yejia-stripe
cc @stripe/api-libraries

## Changelog
* Add support for new resource `PaymentLink`
* Add support for `payment_link` on `Checkout.Session`


remi-stripe (approved):
LGTM but always wary of not testing methods like `allLineItems` since it's easy to break. But I know it's codegen-ed and unlikely

remi-stripe (approved):

PR #1239, opened 2022-02-03
API Updates

yejia-stripe:
Codegen for openapi 10fe9f5.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
* Add `REASON_EXPIRED_UNCAPTURED_CHARGE` enum value on `Refund`.


dcr-stripe (changes requested):
The fixer doesn't seem to have run here (hence why there's so many unchanged files).

From stripe-php, can you run:

```
composer clean
composer install
 ./vendor/bin/php-cs-fixer fix -v --using-cache=no
```

That will reset your local install to bring it up to date with the latest CS fixer version and should now work! Let me know if you're having issues.

dcr-stripe (approved):
This looks good - thanks Yejia! Maybe let's change the changelog to something like:

`Add REASON_EXPIRED_UNCAPTURED_CHARGE enum value on Refund.` 

to be more accurate?

PR #1242, opened 2022-02-11
Fix some PHPDoc in tests

franmomu:
I was curious about raising phpstan level to 2 and 1696 errors appeared, with these changes errors are down to 76.


richardm-stripe:
Thank you for the contribution!

PR #1244, opened 2022-02-23
API Updates

yejia-stripe:
Codegen for openapi 2775931.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
* Add support for new values `bbpos_wisepad3` and `stripe_m2` on enum `Terminal.Reader.device_type`


yejia-stripe:
@pakrym-stripe No, I think this update is just a no-op for PHP because we don't generate all the individual fields of `payment_method_options` under [`PaymentIntent`](https://github.com/stripe/stripe-php/blob/master/lib/PaymentIntent.php). Closing this PR

yejia-stripe:
Ah sorry you're right, we should merge, but not release (based on previous updates). r? @dcr-stripe

PR #1246, opened 2022-03-01
Add ServiceNavigatorTrait to support namespace/resource homonyms

kamil-stripe:
Extracts `__get` method from `AbstractServiceFactory` into a trait can be used across `AbstractService` and `AbstractServiceFactory`. The method helps reading inaccessible, non-existent in our case, properties which will be utilized in future SDK releases for services beneath a namespace named identically to a resource.

### DIFF
--- a/lib/Service/ServiceNavigatorTrait.php
+++ b/lib/Service/ServiceNavigatorTrait.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Stripe\Service;
+
+/**
+ * Trait for creatable resources. Adds a `create()` static method to the class.
+ *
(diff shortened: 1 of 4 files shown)


lib/Service/ServiceNavigatorTrait.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace Stripe\Service;
+
+/**
+ * Trait for creatable resources. Adds a `create()` static method to the class.
```
kamil-stripe, line 6: Needs to be updated

richardm-stripe (approved):
LGTM, let's merge when we release the first service with auxiliary services

PR #1249, opened 2022-03-08
Add missing constant for payment_behavior

LeonMelis:
For subscriptions, there is also `payment_behavior=default_incomplete`, of which the constant was missing.

See: https://stripe.com/docs/api/subscriptions/create#create_subscription-payment_behavior
And: https://stripe.com/docs/api/subscriptions/update#update_subscription-payment_behavior

Also note that the API documentation could really use some improving here. The difference between 'default_incomplete' and 'allow_incomplete' in the scope of updating subscriptions is unclear to me.

### DIFF
--- a/lib/Subscription.php
+++ b/lib/Subscription.php
@@ -57,6 +57,7 @@ class Subscription extends ApiResource
     use ApiOperations\Retrieve;
     use ApiOperations\Update;
 
+    const PAYMENT_BEHAVIOR_DEFAULT_INCOMPLETE = 'default_incomplete';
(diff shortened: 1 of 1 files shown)


remi-stripe:
@LeonMelis Thanks a lot for catching and submitting a PR. Can I ask you to order the constant alphabetically with the other 3? It should go after `PAYMENT_BEHAVIOR_ALLOW_INCOMPLETE` instead

LeonMelis:
Sure. It should now be in alphabetical order

PR #1262, opened 2022-04-12
Add support for updating the version of the repo

pakrym-stripe:
Encodes the logic for version updating as a Makefile. Simplifies what the generation tools needs to know about the repo.
The plan is to have the same interface across all the repos.

Sample usage:

```
make update-version VERSION=1.2.5
```

Also adds a target that codegen will use to format all our repos:

```
make codegen-format
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,31 +1,7 @@
-export PHPDOCUMENTOR_VERSION := v3.0.0
+.PHONY: codegen-format update-version
+update-version:
+	@echo "$(VERSION)" > VERSION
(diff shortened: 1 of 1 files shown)


richardm-stripe (changes requested):
unintended deletes?

pakrym-stripe:
>unintended deletes?

Yep, fixed.

pakrym-stripe:
rl? @richardm-stripe

pakrym-stripe:
re-r? @richardm-stripe

richardm-stripe (approved):

PR #1266, opened 2022-04-18
Codegen for openapi a8928d0

richardm-stripe:
## Summary

Makes corrections to https://github.com/stripe/stripe-php/pull/1265 which was merged but not released. 

We're not releasing `Customer.allFundingInstructions` yet, and we're exposing `$stripe->customers->createFundingInstructions` on the non-services interface only.

### DIFF
--- a/lib/Service/CustomerService.php
+++ b/lib/Service/CustomerService.php
@@ -40,22 +40,6 @@ class CustomerService extends \Stripe\Service\AbstractService
     }
 
     /**
-     * Retrieve all applicable funding instructions for a customer cash balance.
-     *
-     * @param string $parentId
-     * @param null|array $params
-     * @param null|array|\Stripe\Util\RequestOptions $opts
-     *
-     * @throws \Stripe\Exception\ApiErrorException if the request fails
-     *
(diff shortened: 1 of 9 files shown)


lib/Service/CustomerService.php
```diff
@@ -142,17 +126,17 @@ public function createBalanceTransaction($parentId, $params = null, $opts = null
...
      *
-     * @param string $parentId
+     * @param string $id
      * @param null|array $params
      * @param null|array|\Stripe\Util\RequestOptions $opts
      *
      * @throws \Stripe\Exception\ApiErrorException if the request fails
      *
-     * @return \Stripe\funding_instructions
+     * @return \Stripe\Customer
```
kamil-stripe, line 135: Does it return Customer?

kamil-stripe (approved):

lib/Service/CustomerService.php
richardm-stripe: Ah good catch.

richardm-stripe, line 135:
```suggestion
     * @return \Stripe\FundingInstructions
```

PR #1268, opened 2022-04-20
Add $card property specific documentation in PaymentMethod object.

arouze:
Hi,
Just a small documentation improvment based on the [documentation](https://stripe.com/docs/api/payment_methods/object?lang=php#payment_method_object-card).

### DIFF
--- a/lib/PaymentMethod.php
+++ b/lib/PaymentMethod.php
@@ -26,7 +26,7 @@ namespace Stripe;
  * @property \Stripe\StripeObject $bancontact
  * @property \Stripe\StripeObject $billing_details
  * @property \Stripe\StripeObject $boleto
- * @property \Stripe\StripeObject $card
+ * @property null|\Stripe\Card $card If this is a card PaymentMethod, this hash contains the user’s card details.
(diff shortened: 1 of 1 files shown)


lib/PaymentMethod.php
```diff
@@ -26,7 +26,7 @@
  * @property \Stripe\StripeObject $bancontact
  * @property \Stripe\StripeObject $billing_details
  * @property \Stripe\StripeObject $boleto
- * @property \Stripe\StripeObject $card
+ * @property null|\Stripe\Card $card If this is a card PaymentMethod, this hash contains the user’s card details.
```
remi-stripe, line 29: @arouze This change is incorrect. The `Card` class is entirely separate in the stripe-php library and represents an older resource in the API, the [Card object](https://stripe.com/docs/api/cards/object) for a different API. While they share some properties, they are unrelated and this change wouldn't be something we could merge.

remi-stripe:
@arouze Going to close this PR as the change is not correct and what we currently have is correct, hope it's okay!

PR #1274, opened 2022-04-28
Fix PHPDoc on Discount for nullable properties

remi-stripe:
r? @stripe/api-library-reviewers 

Reported by a user!


dcr-stripe (approved):
LGTM. Thanks Remi!

Issue #1237, opened 2022-01-27
Deprecation warning when using PHP 8.1


felipeEddy:
There are new rules with the PHP 8.1 update, some fixes must be done in order to prevent errors and warnings. Example below:

![image](https://user-images.githubusercontent.com/49731840/151442149-0b98ed22-e4ed-4cf6-89ac-f8fdddac43d0.png)

I'm current using version **v7.112.0**.

benhath:
2.4.4 supports PHP 8.1 - setup:upgrade fails with:

Unable to apply data patch StripeIntegration\Payments\Setup\Patch\Data\Patch001WebhooksConfiguration for module StripeIntegration_Payments. Original exception message: Deprecated Functionality: preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated in /Users/***/Development/dev.local/vendor/stripe/module-payments/Model/Config.php on line 295

```php
    public function decrypt($key)
    {
         if (!preg_match('/^[A-Za-z0-9_]+$/', $key))
            $key = $this->encryptor->decrypt($key);

        return trim($key);
    }
```

SachinBahukhandi:
Are PR's being accepted on this?

dcr-stripe:
Hi @felipeEddy , thank you for the report here. I'd love to get these warnings resolved. How are they surfacing for you? This particular case is only possible if calling the magic `__debugInfo()` method on a RequestOptions object (eg. by using `var_dump`) which does not have an API key set. I can do an isolated fix for this, but I'd like to make sure these warnings are all resolved and not just this one.

moonbas3:
Hey @dcr-stripe, simply loading the bindings in an 8.1 environment will spit out a lot of notices that need to be taken care of. Here are some other examples:

`[15-Jun-2022 06:29:06 UTC] PHP Deprecated:  Return type of Stripe\StripeObject::offsetGet($k) should either be compatible with ArrayAccess::offsetGet(mixed $offset): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /mnt/.../stripe/libs/stripe/lib/StripeObject.php on line 215
[...]

PR #1289, opened 2022-05-20
fix: Update RequestOptions#redactedApiKey to stop exploding null.

dcr-stripe:
r? @kamil-stripe 

## Summary

Updates `RequestOptions#redactedApiKey` to handle the null case properly, otherwise this shows a warning in PHP 8.1 as passing null to explode is deprecated.

## Motivation

https://github.com/stripe/stripe-php/issues/1237

### DIFF
--- a/lib/Util/RequestOptions.php
+++ b/lib/Util/RequestOptions.php
@@ -154,8 +154,13 @@ class RequestOptions
         throw new \Stripe\Exception\InvalidArgumentException($message);
     }
 
+    /** @return string */
     private function redactedApiKey()
     {
+        if (is_null($this->apiKey)) {
+            return '';
+        }
+
         $pieces = \explode('_', $this->apiKey, 3);
         $last = \array_pop($pieces);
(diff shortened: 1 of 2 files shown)


lib/Util/RequestOptions.php
```diff
@@ -154,8 +154,13 @@ public static function parse($options, $strict = false)
         throw new \Stripe\Exception\InvalidArgumentException($message);
     }
 
+    /** @return string */
     private function redactedApiKey()
     {
+        if (null === $this->apiKey) {
```
kamil-stripe, line 160: Yoda

kamil-stripe, line 160: Yoda

kamil-stripe (approved):

lib/Util/RequestOptions.php
dcr-stripe: Heh it's how the formatter did it 😆

PR #1292, opened 2022-05-22
Fix: Run `friendsofphp/php-cs-fixer` as installed with `composer`

localheinz:
This pull request

- [x] runs `friendsofphp/php-cs-fixer` as installed with `composer`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,10 +18,29 @@ jobs:
     steps:
       - uses: actions/checkout@master
 
-      - name: PHP-CS-Fixer
-        uses: docker://oskarstark/php-cs-fixer-ga:3.4.0
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -18,10 +18,29 @@ jobs:
     steps:
       - uses: actions/checkout@master
 
-      - name: PHP-CS-Fixer
-        uses: docker://oskarstark/php-cs-fixer-ga:3.4.0
```
localheinz, line 22: Compare with `composer.json`:

https://github.com/stripe/stripe-php/blob/47cfc4046e2b4b8d861e18dfcf6e74eea8f99374/composer.json#L26

dcr-stripe:
Thanks for the PR @localheinz and sorry for the delay in responding! Our current workflow uses 3.4.0 explicitly since 3.5.0 doesn't support PHP 7.1 (which we support in CI) and composer fails to install in that case.

PR #1293, opened 2022-05-22
Install deps in the install step of CI

szepeviktor:
Without `--no-update` `composer remove` installs all dependencies.


szepeviktor:
Thank you.

szepeviktor:
@richardm-stripe Please ping me if you want to make CI UI more readable.

kép

PR #1304, opened 2022-06-14
Allow passing PSR-3 loggers to setLogger as they are compatible

Seldaek:
IMO you should really just use PSR-3 and require psr/log as it is anyway part of almost every PHP project at this point.. but at least this allows setting things up without getting static analysis warnings.


dcr-stripe:
Thank you for the contribution @Seldaek and sorry about the delay! I'll discuss this with the team and get back to you but I think this makes a ton of sense!

dcr-stripe:
Hi @Seldaek - this looks good to me! I've brought your branch up to head - there's a small linter error:

```

     /**
-     * @param Util\LoggerInterface|\Psr\Log\LoggerInterface $logger the logger to which the library
+     * @param \Psr\Log\LoggerInterface|Util\LoggerInterface $logger the logger to which the library
      *   will produce messages
      */
```

Do you mind applying this change, and then I can give this an approval and merge?

Thank you for the contribution!

Seldaek:
Sure, done.

dcr-stripe (commented):
LGTM! Thank you for your contribution @Seldaek , it'll go out with the next release!

PR #1322, opened 2022-07-12
API Updates

yejia-stripe:
Codegen for openapi v164.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
* Change type of `source_type` on `Transfer` from nullable string to string (comment-only change)


dcr-stripe (approved):
Empty changelog. Let's fix that and then LGTM!

PR #1324, opened 2022-07-18
API Updates

pakrym-stripe:
Codegen for openapi v165.
r? @kamil-stripe
cc @stripe/api-libraries

## Changelog
* Add support for new value `blik` on enum `PaymentLink.payment_method_types[]`
* Add support for `blik` on `PaymentMethod`
* Add support for new value `blik` on enum `PaymentMethod.type`


kamil-stripe (approved):
Should we add `Invoice.upcomingLines` and `Source.allSourceTransactions` to the changelog?

PR #1344, opened 2022-08-01
API Updates

pakrym-stripe:
Codegen for openapi v174.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `shipping_cost` and `shipping_details` on `Checkout.Session`


dcr-stripe (approved):
LGTM, though we probably want to tweak the changelog to remove unnecessary entries that are captured by the larger breaking changes PR.

PR #1350, opened 2022-08-17
Ability to inject service instances to the factory

cangelis:
Instances resolved externally can be passed to the factory. This is useful for the mock services that are created in unit tests.

Example use case (using Laravel's `mock` method which uses Mockery under the hood);

```php
$client = app(StripeClient::class);

$customerService = $this->mock(CustomerService::class, function ($mock) {
    $mock->shouldReceive('retrieve')->andReturn(Customer::constructFrom('......'));
});

$client->getFactory()->setServiceInstance(CustomerService::class, $customerService);

// returns the mock
$client->customers;
```


lib/Service/AbstractServiceFactory.php
```diff
@@ -56,4 +56,17 @@ public function __get($name)
 
         return null;
     }
+
+    /**
+     * Set the instance of a service that is instantiated externally
+     * This is useful for the cases where a mock service is created in the tests
+     * so it can be injected to the factory
```
richardm-stripe, line 63:
```suggestion
     * so it can be injected to the factory.
```

lib/StripeClient.php
```diff
@@ -82,4 +82,14 @@ public function __get($name)
 
         return $this->coreServiceFactory->__get($name);
     }
+
+    /**
+     * Returns the factory that is used for resolving the services
```
richardm-stripe, line 87:
```suggestion
     * Returns the factory that is used for resolving the services.
```

pakrym-stripe:
Hey @cangelis, sorry for a delayed response. We were thinking about the problem of mocking StripeClient properties and came up with alternative solution of [making the `__get` itself mockable by exposing it as a named method](https://github.com/stripe/stripe-php/pull/1381). Can you check if the solution form the [above PR](https://github.com/stripe/stripe-php/pull/1381) works for you case?

pakrym-stripe:
I've closed this as superseded by #1381. Please file an issue if we can improve service mocking in other ways.

Issue #1353, opened 2022-08-22
PaymentIntent "last_payment_error" property is not an ErrorObject


YonniRouelle:
### Describe the bug

I'm retrieving some PaymentIntent with errors and I can't use Stripe\ErrorObject property because the `last_payment_error` property of a PaymentIntent is an instance of Stripe\StripeObject

### To Reproduce

1. Retrieve a `PaymentIntent` from Stripe which has a `last_payment_error` property defined
2. Dump the `PaymentIntent` retrieved
3. Check the `last_payment_error` property type
4. The `last_payment_error` property is an instance of `Stripe\StripeObject`

### Expected behavior

I expect the property to be an instance of `Stripe\ErrorObject`

### Code snippets

```PHP
$paymentIntent = $stripeClient->paymentIntents->retrieve('payment_intent_id');

dump($paymentIntent->last_payment_error instanceof Stripe\ErrorObject); // false
```

### OS

Ubuntu 22.04.1 LTS

### PHP version

PHP 8.1

### Library version

stripe-php v9.0.0

### API version

2022-08-01

dcr-stripe:
Thanks for the report @YonniRouelle - sorry you're experiencing issues here!

I believe this is because we aren't specifying the valid error types in https://github.com/stripe/stripe-php/blob/19218fbe/lib/Util/ObjectTypes.php#L12, and therefore these objects don't get transformed from a plain `Stripe\StripeObject` to a `Stripe\ErrorObject`.

dcr-stripe:
Following up here - I submitted https://github.com/stripe/stripe-php/pull/1361 to actually fix the type hints to reflect reality.

The team will investigate handling these fields more consistently across all SDKs to provide a better user experience.

Thanks for the patience!

kissifrot:
Hello, a side effect of this change is we cannot use code such as `$paymentIntent->last_payment_error->message` anymore as static analysis tools such as phpstan will complain about missing `message` not being part of the class anymore.
However [API docs](https://stripe.com/docs/api/errors#errors-payment_intent-last_payment_error) sill mention those fields.

What is the workaround for this?

PR #1357, opened 2022-08-23
fix: Transform error StripeObjects to ErrorObject.

dcr-stripe:
r? kamil-stripe

## Summary

Include the error types in the object type mapping and start transforming API objects returned with these types to `ErrorObject`.

## Motivation

Fixes https://github.com/stripe/stripe-php/issues/1353

We explicitly state in the type annotation that these are `ErrorObject`s but they weren't getting transformed appropriately.


kamil-stripe (approved):

dcr-stripe:
Closing this out - we opted for fixing the type hints (https://github.com/stripe/stripe-php/pull/1361) and then re-visiting this as part of the next major to handle this consistently across all SDKs.

PR #1360, opened 2022-08-24
Add test coverage using Coveralls

anniel-stripe:
## Summary
r? @pakrym-stripe 

Add coveralls CI step. Test coverage data will be published to https://coveralls.io/github/stripe/stripe-php and displayed as a badge in the README.

Example build: https://coveralls.io/jobs/105038768


.coveralls.github-actions.yml
```diff
@@ -0,0 +1,3 @@
+service_name: github-actions
```
pakrym-stripe, line 1: Can we pass these as direct parameter to the coveralls as parameters?
```
./vendor/bin/php-coveralls -v --coverage_clover='./dist/tests.clover' --json_path='./dist/coveralls-upload.json'
```
This will allows us to have fewer random files in the repo root.

.github/workflows/ci.yml
```diff
@@ -135,3 +135,11 @@ jobs:
         run: |
           php --version
           ./build.php $env:AUTOLOAD
+
+      - name: Coveralls
```
pakrym-stripe, line 139: nit: This is the third pattern of running coveralls in addition to a separate job and the parallel matrix with finish job. I think it's the best one but we need to look into consistency a bit.

pakrym-stripe (approved):
One minor comment.

.github/workflows/ci.yml
anniel-stripe: I agree, I plan to stick with this pattern and go back to change the rest of the libraries to follow it.

.github/workflows/ci.yml
```diff
@@ -137,7 +137,7 @@ jobs:
           ./build.php $env:AUTOLOAD
 
       - name: Coveralls
-        run: vendor/bin/php-coveralls -c .coveralls.github-actions.yml -v
+        run: ./vendor/bin/php-coveralls -v --coverage_clover='./dist/tests.clover' --json_path='./dist/coveralls-upload.json'
```
pakrym-stripe: do we need service_name too?

pakrym-stripe (approved):

.github/workflows/ci.yml
anniel-stripe: I actually don't think we do, the command seems to work fine in CI without it?

PR #1369, opened 2022-09-23
Add SingletonRetrieve that can be used in generated code

pakrym-stripe:
Move `_singletonRetrieve` into a trait that can be used similarly to other standard method traits.


yejia-stripe (approved):
nice, lgtm!

PR #1372, opened 2022-10-06
API Updates

kamil-stripe:
Codegen for openapi v201.
cc @stripe/api-libraries


richardm-stripe (approved):
Just copy changes, so release unnecessary, but approving.

PR #1291, opened 2022-05-22
Fix: Configure finder for `friendsofphp/php-cs-fixer`

localheinz:
This pull request

- [x] configures the finder for `friendsofphp/php-cs-fixer`

Running

```shell
vendor/bin/php-cs-fixer fix
```

on current `master` yields

```
Loaded config default from "/Users/am/Sites/stripe/stripe-php/.php-cs-fixer.php".
Using cache file ".php-cs-fixer.cache".

In Finder.php line 611:

  You must call one of in() or append() methods before iterating over a Finder.


fix [--path-mode PATH-MODE] [--allow-risky ALLOW-RISKY] [--config CONFIG] [--dry-run] [--rules RULES] [--using-cache USING-CACHE] [--cache-file CACHE-FILE] [--diff] [--format FORMAT] [--stop-on-violation] [--show-progress SHOW-PROGRESS] [--] [<path>...]
```

PR #1379, opened 2022-10-17
Do not run Coveralls in PR-s

szepeviktor:
when the secret token is not available

From https://github.com/stripe/stripe-php/pull/1291#issuecomment-1281188329


anniel-stripe (approved):

szepeviktor:
Thank you, Anniel!

PR #1380, opened 2022-10-17
Fix: Run `phpstan` job with PHP 8.1 only

localheinz:
This pull request

- [x] runs the `phpstan` job with PHP 8.1 only

💁‍♂️ I doubt that `phpstan/phpstan` will find different issues on different versions of PHP.


localheinz:
@richardm-stripe 

Really, does every push need approval from a maintainer?

szepeviktor:
> I doubt that `phpstan/phpstan` will find different issues on different versions of PHP.

This can be ultimately resolved by adding PHP version to PHPStan configuration file.
e.g. `phpVersion: 70400`

pakrym-stripe:
Thank you for your contribution, @localheinz. Sorry for the delay.

localheinz:
Thank you, @pakrym-stripe!

PR #1385, opened 2022-10-24
retrievePaymentMethod's return type is wrong.

yuki-yasumura:
### summary
`retrievePaymentMethod`'s return type is wrong.

```
++ @return \Stripe\Customer
-- @return \Stripe\PaymentMethod
```


richardm-stripe:
Hi @yuki-yasumura. This change looks correct to me, unfortunately this code is generated so it would just be reverted if we merged as-is. We can merge this PR after fixing the underlying bug in our generator

yuki-yasumura:
@richardm-stripe

Thank you very much.

I see, so that is the background.
Fixing a fundamental bug in the generator, please merge!

PR #1388, opened 2022-11-01
Remove AlipayAccount

anniel-stripe:
## Summary
Remove AlipayAccount code as AlipayAccount was removed in [9.0.0](https://github.com/stripe/stripe-php/blob/ef98d5b4/CHANGELOG.md#900---2022-08-02).

## Changelog
* Remove AlipayAccount resource file that should have been removed in v9.0.0.

PR #1392, opened 2022-11-07
Next major release changes

pakrym-stripe:
# Changelog

## 10.0.0 - 2022-11-15

Breaking changes that arose during code generation of the library that we postponed for the next major version. For changes to the Stripe products, read more at https://stripe.com/docs/upgrades#2022-11-15.

"⚠️" symbol highlights breaking changes.

### Deprecated
* Mark `resource.save` as deprecated (#1382). Prefer the static update method that doesn't require retrieval of the resource to update it.
```PHP
// before
$resource = Price::retrieve(self::TEST_RESOURCE_ID);
$resource->metadata['key'] = 'value';
$resource->save();

// after
$resource = Price::update('price_123', [
    'metadata' => ['key' => 'value'],
]);
```

### ⚠️ Removed
- Removed deprecated `Sku` resource and service (#1377)
- Removed deprecated `Orders` resource and service (#1375)
- Removed deprecated `Product` field from the `LineItem` (#1375)
- Removed deprecated `AlipayAccount` resource (#1388)
- Removed `charges` field on `PaymentIntent` and replace it with `latest_charge`. (TODO)


lib/Refund.php
```diff
@@ -9,6 +9,11 @@
  * been created but not yet refunded. Funds will be refunded to the credit or debit
  * card that was originally charged.
  *
+ * Stripe Tax users with recurring payments and invoices can create <a
+ * href="https://stripe.com/docs/api/credit_notes">Credit Notes</a>, which reduce
+ * overall tax liability because tax is correctly recalculated and apportioned to
+ * the related invoice.
+ *
```
ramya-stripe, line 16: Why is this comment part of "breaking changes"?

pakrym-stripe: Responded offline.

PR #1402, opened 2022-11-23
Add valid now value to trial_end

Seldaek:
As per https://stripe.com/docs/api/subscriptions/update#update_subscription-trial_end


remi-stripe:
Hey @Seldaek! The PHPdocs describe the **property** itself that can be returned by our API. The `now` value is only accepted as a **parameter** to our API. You pass `trial_end: 'now'` and we end the trial immediately. This means that `trial_end: 'now'` is never returned back in the API and instead would be the current timestamp.

That change is not something that we would merge so I'm going to close but let me know if you have any questions!

PR #1407, opened 2022-12-08
API Updates

pakrym-stripe:
Codegen for openapi v213.
cc @stripe/api-libraries


richardm-stripe (approved):
LGTM but no need to release

PR #1413, opened 2022-12-22
API Updates

richardm-stripe:
Codegen for openapi v216.
cc @stripe/api-libraries

## Changelog
Change `CheckoutSession.cancel_url` to be nullable.


pakrym-stripe (approved):
Changelog might need manual editing.

PR #1424, opened 2023-01-19
API Updates

pakrym-stripe:
Codegen for openapi v219.
cc @stripe/api-libraries

## Changelog

* Added `REFUND_CREATED`, `REFUND_UPDATED` event definitions.


anniel-stripe (approved):
Does the Changelog need to be updated to include the new refund statuses?

PR #1425, opened 2023-01-19
Fix Stripe::setAccountId parameter type

villfa:
The account ID is a nullable property but, according the PHPDoc, the related setter method doesn't accept `null` as an argument (while in fact passing `null` works fine).
I think it's correct to accept `null` to be able to reset this static property.

Updating the PHPDoc will fix the PHPStan errors of the form `Parameter #1 $accountId of method Stripe\Stripe::setAccountId() expects string, null given.`.

### DIFF
--- a/lib/Stripe.php
+++ b/lib/Stripe.php
@@ -185,7 +185,7 @@ class Stripe
     }
 
     /**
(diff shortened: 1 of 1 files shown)


pakrym-stripe:
Thank you, @villfa!

PR #1431, opened 2023-01-30
Fix: Do not use unbounded version constraint for `actions/checkout`

localheinz:
This pull request

- [x] avoids using an unbounded version constraint for `actions/checkout`


pakrym-stripe:
Thank you, @localheinz !

localheinz:
Thank you, @pakrym-stripe!

PR #1435, opened 2023-01-30
Fix: Keep rules sorted

localheinz:
This pull request

- [x] keeps rules sorted in `.php-cs-fixer.php`


pakrym-stripe:
@localheinz `PHP-CS-Fixer` seems to be unhappy with this PR.

PR #1437, opened 2023-01-30
Enhancement: Use `phive` to install `friendsofphp/php-cs-fixer`

localheinz:
This pull request

- [x] uses [`phive`](https://phar.io) to install `friendsofphp/php-cs-fixer`


pakrym-stripe:
Thank you for the contribution @localheinz. 
I see a binary file check in as part of this PR. Is this required?

PR #1446, opened 2023-03-01
stripe->customers->retrievePaymentMethod returns the wrong class (type hint)

highstrike:
changed from the customer object to the payment method object type hint


pakrym-stripe:
@highstrike, thank you for your contribution!

highstrike:
👍🏻

PR #1450, opened 2023-03-09
API Updates

anniel-stripe:
Codegen for openapi v249.
cc @stripe/api-libraries

## Changelog
* Add support for `cancellation_details` on `Subscription`
* Fix return types on custom methods (extends https://github.com/stripe/stripe-php/pull/1446)


richardm-stripe:
Looks like there's some fixed docstrings too? Worth mentioning in the changelog. There was probably a generator fix for custom method return types.

richardm-stripe (approved):

PR #1488, opened 2023-04-27
Increment PHPStan to strictness level 2

richardm-stripe:
PHPStan has "rule levels" https://phpstan.org/user-guide/rule-levels that determine which sorts of type errors they detect. I was surprised to discover we were on rule level 1. This PR adds type fixes and error suppressions to get us to rule level 2.

Almost everything is docstring-only changes, with the exception of

[Fix unsafe static:: references](https://github.com/stripe/stripe-php/pull/1488/commits/6ef895a85b7abfee80c311e07e8649340ce191dd). PHPStan doesn't like `static::` references to private methods. I'm not 100% sure why but I think it's because
```php
<?php
class Foo {
    private static function abc(){
        return "hello";
    }
    public static function xyz() {
        return static::abc();
    }
}
class Bar extends Foo {
    private static function abc() {
        return "world";
    }
}
print Bar::xyz();
```
will result in an error like
```
Fatal error: Uncaught Error: Call to private method Bar::abc() from scope Foo 
```

In any case, doing `self::` instead makes it so that the private method defined in the parent class is always called, and prevents subclasses from re-implementing it, which I think is what is intended, anyway.


lib/Account.php
```diff
@@ -127,6 +129,7 @@ public function serializeParameters($force = false)
         return $update;
     }
 
+    // TODO (MAJOR): Remove legal_entity/additional_owners logic.
```
anniel-stripe, line 132:
```suggestion
```

lib/Service/AbstractService.php
```diff
@@ -72,22 +72,28 @@ private static function formatParams($params)
...
     protected function request($method, $path, $params, $opts)
     {
-        return $this->getClient()->request($method, $path, static::formatParams($params), $opts);
+        return $this->getClient()->request($method, $path, self::formatParams($params), $opts);
     }
 
     protected function requestStream($method, $path, $readBodyChunkCallable, $params, $opts)
     {
-        return $this->getStreamingClient()->requestStream($method, $path, $readBodyChunkCallable, static::formatParams($params), $opts);
+        // TODO (MAJOR): Add this method to StripeClientInterface
```
anniel-stripe, line 80: Also in this file
```suggestion
```

lib/Service/AbstractService.php
```diff
@@ -72,22 +72,28 @@ private static function formatParams($params)
...
+        // @phpstan-ignore-next-line
+        return $this->getStreamingClient()->requestStream($method, $path, $readBodyChunkCallable, self::formatParams($params), $opts);
     }
 
     protected function requestCollection($method, $path, $params, $opts)
     {
-        return $this->getClient()->requestCollection($method, $path, static::formatParams($params), $opts);
+        // TODO (MAJOR): Add this method to StripeClientInterface
```
anniel-stripe, line 87:
```suggestion
```

lib/Service/AbstractService.php
```diff
@@ -72,22 +72,28 @@ private static function formatParams($params)
...
+        // @phpstan-ignore-next-line
+        return $this->getClient()->requestCollection($method, $path, self::formatParams($params), $opts);
     }
 
     protected function requestSearchResult($method, $path, $params, $opts)
     {
-        return $this->getClient()->requestSearchResult($method, $path, static::formatParams($params), $opts);
+        // TODO (MAJOR): Add this method to StripeClientInterface
```
anniel-stripe, line 94:
```suggestion
```

anniel-stripe (approved):

PR #1502, opened 2023-05-23
Generate error codes

pakrym-stripe:
Generate error code constants.

### DIFF
--- a/lib/ErrorObject.php
+++ b/lib/ErrorObject.php
@@ -41,10 +41,6 @@ class ErrorObject extends StripeObject
      * Possible string representations of an error's code.
      *
      * @see https://stripe.com/docs/error-codes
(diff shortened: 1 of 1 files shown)


lib/ErrorObject.php
```diff
@@ -41,7 +41,13 @@ class ErrorObject extends StripeObject
      * Possible string representations of an error's code.
      *
      * @see https://stripe.com/docs/error-codes
+     *
+     * @param mixed $values
+     * @param mixed $opts
+     * @param mixed $partial
```
anniel-stripe, line 47: What are these for?

anniel-stripe (commented):
~nit about the param docstring~

anniel-stripe (approved):

PR #1508, opened 2023-05-29
Webhook: improves validation and simplifies code

drhino:
Compatible with php5.6 and passes validation.
The second commit updates the test to allow for stricter validation (cfr. expecting a sha256 and modern timestamp).

More info about the updated test:
- `bad_signature` in the test is not a valid sha256 hash because it contains an underscore and does not equal to 64 chars.
- The last test uses `12345` which is far away from the current time. The improved validation requires an unsigned integer with 10 digits.

Whats changed:
- Now uses regex to extract the values.
- Moved the timestamp-check so theres no need to break the foreach.

P.S: git shows more changes than actually happened because the timestamp code moved.
Exceptions and messages are left unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,7 +246,7 @@ background terminal ([stripe-mock's README][stripe-mock] also contains
 instructions for installing via Homebrew and other methods):
 
 ```bash
-go get -u github.com/stripe/stripe-mock
+go install github.com/stripe/stripe-mock@latest
 stripe-mock
 ```
 
--- a/tests/Stripe/WebhookTest.php
+++ b/tests/Stripe/WebhookTest.php
@@ -79,7 +79,7 @@ final class WebhookTest extends \Stripe\TestCase
         $this->expectException(\Stripe\Exception\SignatureVerificationException::class);
         $this->expectExceptionMessage('No signatures found matching the expected signature for payload');
 
-        $sigHeader = $this->generateHeader(['signature' => 'bad_signature']);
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -246,7 +246,7 @@ background terminal ([stripe-mock's README][stripe-mock] also contains
 instructions for installing via Homebrew and other methods):
 
 ```bash
-go get -u github.com/stripe/stripe-mock
+go install github.com/stripe/stripe-mock@latest
```
anniel-stripe, line 249: Happy to accept this fix to the README, thanks for catching!

tests/Stripe/WebhookTest.php
```diff
@@ -79,7 +79,7 @@ public function testNoValidSignatureForPayload()
         $this->expectException(\Stripe\Exception\SignatureVerificationException::class);
         $this->expectExceptionMessage('No signatures found matching the expected signature for payload');
 
-        $sigHeader = $this->generateHeader(['signature' => 'bad_signature']);
+        $sigHeader = $this->generateHeader(['signature' => \hash('sha256', 'bad_signature')]);
```
anniel-stripe, line 82: We don't need this to be a valid hash, since it's meant to be a "bad_signature" anyways :)

anniel-stripe:
Hi @drhino , I don't think we can accept this PR in its current state. The current code is intended to explicitly show the logic for webhook verification, and changing to regexes makes it less readable. Is there an issue you encountered that prompted this change?

anniel-stripe (approved):
README changes LGTM!

PR #1509, opened 2023-06-01
Webhook: `getTimestamp()`: fixes `is_numeric()` issues 

drhino:
`is_numeric()` allows for invalid timestamps, for example: a negative integer

```php
<?php
// the following pass the validation with `is_numeric()`
$tests = array(
    "0x539",
    "02471",
    "9.1",
    " 42",
    "58635272821786587286382824657568871098287278276543219876543",
    "0.0",
    ".0",
    "0.",
    "1.8e617",
    "1337e0",
    "0E1",
    "2E1",
    "1e-2",
    "-1.3e3",
    " +3",
    " -1.1",
    "000",
    "-1",
);
?>
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,7 +246,7 @@ background terminal ([stripe-mock's README][stripe-mock] also contains
 instructions for installing via Homebrew and other methods):
 
 ```bash
-go get -u github.com/stripe/stripe-mock
+go install github.com/stripe/stripe-mock@latest
 stripe-mock
 ```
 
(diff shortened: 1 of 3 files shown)


richardm-stripe:
Hello @drhino, thank you for the submission. I'm closing this PR though as I don't see the need to defend against these values. The Stripe API should never produce values that look like this, and if an attacker submits abnormal values for the timestamp, it won't pass the signature check and should be rejected.

Feel free to re-open and provide details about a scenario where this would be helpful!

PR #1523, opened 2023-07-07
[Next Major] Pin to latest API version

richardm-stripe:
## Changelog
  **⚠️ ACTION REQUIRED: the breaking change in this release likely affects you ⚠️**

  In this release, Stripe API Version `2023-08-16` (the latest at time of release) will be sent by default on all requests.
  The previous default was to use your [Stripe account's default API version](https://stripe.com/docs/development/dashboard/request-logs#view-your-default-api-version).

  To successfully upgrade to stripe-php v11, you must either

  1. **(Recommended) Upgrade your integration to be compatible with API Version `2023-08-16`.**

     Please read the API Changelog carefully for each API Version from `2023-08-16` back to your [Stripe account's default API version](https://stripe.com/docs/development/dashboard/request-logs#view-your-default-api-version). Determine if you are using any of the APIs that have changed in a breaking way, and adjust your integration accordingly. Carefully test your changes with Stripe [Test Mode](https://stripe.com/docs/keys#test-live-modes) before deploying them to production.

     You can read the [v11 migration guide](https://github.com/stripe/stripe-php/wiki/Migration-guide-for-v11) for more detailed instructions.
  2. **(Alternative option) Specify a version other than `2023-08-16` when initializing `stripe-php`.**

     If you were previously initializing stripe-php without an explicit API Version, you can postpone modifying your integration by specifying a version equal to your [Stripe account's default API version](https://stripe.com/docs/development/dashboard/request-logs#view-your-default-api-version). For example:

     ```diff
       // if using StripeClient
     - $stripe = new \Stripe\StripeClient('sk_test_xyz');
     + $stripe = new \Stripe\StripeClient([
     +   'api_key' => 'sk_test_xyz',
         'stripe_version' => '2020-08-27',
     + ]);

       // if using the global client
       Stripe.apiKey = "sk_test_xyz";
     + Stripe::setApiVersion('2020-08-27');
     ```

     If you were already initializing stripe-php with an explicit API Version, upgrading to v11 will not affect your integration.

     Read the [v11 migration guide](https://github.com/stripe-php/wiki/Migration-guide-for-v11) for more details.

  Going forward, each major release of this library will be *pinned* by default to the latest Stripe API Version at the time of release.

  That is, instead of upgrading stripe-php and separately upgrading your Stripe API Version through the Stripe Dashboard. whenever you upgrade major versions of stripe-php, you should also upgrade your integration to be compatible with the latest Stripe API version.


lib/BaseStripeClient.php
```diff
@@ -18,6 +18,7 @@ class BaseStripeClient implements StripeClientInterface, StripeStreamingClientIn
         'api_key' => null,
         'client_id' => null,
         'stripe_account' => null,
+        // Note, even if null, ApiRequestor will default this to Stripe::$apiVersion
```
pakrym-stripe, line 21: isn't that wrong? Should StripeClient inherit global settings?

pakrym-stripe (changes requested):
qq

lib/BaseStripeClient.php
richardm-stripe: We can change the behavior so that StripeClient initialized without an explicit stripe_version sends requests using the pinned version instead of `Stripe::$apiVersion` but that change is breaking.

richardm-stripe: e.g.
```php
Stripe::setApiVersion("2020-XX-YYYY");
$stripe = new StripeClient("sk_...");
$stripe->customers->create(...);
```
will use a different API version if we change the fallback behavior.

Are you suggesting we make that change now?

pakrym-stripe (approved):

lib/BaseStripeClient.php
richardm-stripe: I opened https://github.com/stripe/stripe-php/pull/1551 in response to our convo here. I'm not sure we should merge it along with the major.

PR #1526, opened 2023-07-11
Add constants for payment intent cancellation reasons

javespi:
Add the different values available for cancellation reason field when a payment intent is cancelled.

Values retrieved from API documentation:
- https://stripe.com/docs/api/payment_intents/cancel#cancel_payment_intent-cancellation_reason

Similar approach as it was already done for statuses:
- https://github.com/stripe/stripe-php/pull/655


richardm-stripe:
Hello @javespi, thank you for the pull request.

The reason these constants aren't already in the library is because they exist on the request-side but not the response side. Presently, the library only exposes constants that appear on the response side.

I've opened up a feature request in #1530 but this means I'm not going to merge your pull-request as-is, if we add request-side constants that would happen library-wide and I don't want to make exceptions in case it is inconsistent with the eventual design.

richardm-stripe:
Thanks for the contribution! I've fixed the underlying problem in our library generator, so this enum will be added once https://github.com/stripe/stripe-php/pull/1529/commits/e2b090c499c181d4cba156595f7241fe6fa0c6be#diff-5239e074340aff1a427e536f1944bba3ff00d356fafb24a988da1505c7236da7 is merged for our next release.

I'll merge your PR, even though it will get wiped by the generated version, so you get committer credit!

richardm-stripe (approved):

javespi:
Thank you @richardm-stripe for the effort! Which stack are you using for the library generator? I guess based on OpenAPI descriptions.

I'm curious about it; it would be worth to explore for some APIs I am currently developing!

PR #1551, opened 2023-08-15
Warn if StripeClient + global API version

richardm-stripe:
## Motivation
From discussion on [version pinning](https://github.com/stripe/stripe-php/pull/1523/files#r1293999062). Right now, the behavior of `StripeClient` is that when you don't specify `stripe_version`, the fallback is to the global, mutable `Stripe::$apiVersion`. This is regrettable, doesn't match what we are doing in other libraries, and makes reasoning about the behavior of the library more complicated.

We're not ready to change this behavior -- especially not alongside the introduction of version pinning -- but as a start we should discourage users from relying on this behavior, and encourage them to either avoid `Stripe::$apiVersion` altogether if they are exclusively using `StripeClient`, or to set a version explicitly when initializing `StripeClient` if they still have code using the global client.

## Summary
This PR causes the library to emit a deprecation warning (once) via [`\trigger_error(..., \E_USER_DEPRECATION)`](https://www.php.net/manual/en/function.trigger-error.php) if the user sends a request from `StripeClient` without a hardcoded version and it appears they have interacted with `Stripe::$apiVersion`.


richardm-stripe:
Turns out I was wrong about the fallback behavior of StripeClient.

PR #1556, opened 2023-08-18
Merge master into beta

pakrym-stripe:


lib/ApiRequestor.php
```diff
@@ -406,12 +408,6 @@ function ($key) use ($params) {
         }
         $defaultHeaders = $this->_defaultHeaders($myApiKey, $clientUAInfo);
 
-        if ('preview' === $apiMode && !isset($headers['Stripe-Version'])) {
```
pakrym-stripe, line 409: This logic moved to BaseStripeClient

lib/BaseStripeClient.php
```diff
@@ -186,7 +186,10 @@ public function rawRequest($method, $path, $params = null, $opts = [])
             $headers['Stripe-Context'] = $opts['stripe_context'];
             unset($opts['stripe_context']);
         }
-        $opts = $this->defaultOpts->merge($opts, true);
+
+        $defaultRawRequestOpts = $this->defaultOpts->merge(['stripe_version' => \Stripe\Util\ApiVersion::PREVIEW], true);
```
anniel-stripe, line 190: Shouldn't we only set `'stripe_version'` to the preview version if `'preview' === $apiMode`?

pakrym-stripe: Yes, I'll add another test

anniel-stripe (approved):

PR #1562, opened 2023-08-28
Nicer ApiErrorException::__toString()

### LINKED ISSUE #1561: Provide option to retain full stack traces in ApiErrorException
### Is your feature request related to a problem? Please describe.

The `Stripe\ApiErrorException` class overrides the default `Throwable::__toString()` implementation to supplement the exception message with request status and ID information when it's available. This added information is great! However, it means that the default behavior that renders a stack trace is lost, which makes tracking down and debugging errors significantly more challenging.

This can be exacerbated when using a PSR-3 error logger, providing it to [`exception` context value](https://www.php-fig.org/psr/psr-3/#13-context). Such loggers (as well as non-PSR loggers) often rely on `__toString()`, since the default behavior handles all sorts of fiddly formatting (adding line numbers, checking for and displaying previous exceptions, etc).

Since the Stripe API exceptions override that method, it means that a common setup becomes significantly less valuable and harder to use.

### Describe the solution you'd like

Ideally, I'd prefer the original `Throwable::__toString()` behavior in place by default - though I recognize that this could be disruptive to people that are expecting the current behavior. 

Given that, an option for the same would be really beneficial. I could envision this as an option passed when creating a `new Stripe\StripeClient` or (somewhat less ideal, but possibly more straightforward to implement) something like `Stripe\ApiErrorException::setDisplayStackTraces(bool)`

### Describe alternatives you've considered

If loggers are built in-house, it's possible to modify them to avoid reliance on `__toString()` and instead rebuild equivalent logic based on the other accessors on `Throwable`s. However, in addition to being a whole lot of redundant work, we'd still have to special-case the Stripe exception to re-add the data provided in the current implementation (if it's so desired).

For third-party loggers, including those bundled with frameworks, this isn't really a practical option.

### Additional context

_No response_

richardm-stripe:
Fixes #1561 

Changes ApiErrorException::__toString() to have more useful information (a stack trace). This should be a safe change: `__toString()` is for human consumption and not intended to be consumed programatically.

Before this change:
```
(Status 200) (Request req_test) message
```
After this change:
```
Error sending request to Stripe: (Status 200) (Request req_test) message
Mock_ApiErrorException_21f57d2b: message in /Users/richardm/stripe/stripe-php/lib/Exception/ApiErrorException.php:38
Stack trace:
#0 /Users/richardm/stripe/stripe-php/tests/Stripe/Exception/ApiErrorExceptionTest.php(17): Stripe\Exception\ApiErrorException::factory('message', 200, '{"error": {"cod...', Array, Array, 'some_code')
#1 /Users/richardm/stripe/stripe-php/tests/Stripe/Exception/ApiErrorExceptionTest.php(45): Stripe\Exception\ApiErrorExceptionTest->createFixture()
#2 /Users/richardm/stripe/stripe-php/vendor/phpunit/phpunit/src/Framework/TestCase.php(1608): Stripe\Exception\ApiErrorExceptionTest->testToString()
#3 /Users/richardm/stripe/stripe-php/vendor/phpunit/phpunit/src/Framework/TestCase.php(1214): PHPUnit\Framework\TestCase->runTest()
#4 /Users/richardm/stripe/stripe-php/vendor/phpunit/phpunit/src/Framework/TestResult.php(728): PHPUnit\Framework\TestCase->runBare()
#5 /Users/richardm/stripe/stripe-php/vendor/phpunit/phpunit/src/Framework/TestCase.php(964): PHPUnit\Framework\TestResult->run(Object(Stripe\Exception\ApiErrorExceptionTest))
#6 /Users/richardm/stripe/stripe-php/vendor/phpunit/phpunit/src/Framework/TestSuite.php(684): PHPUnit\Framework\TestCase->run(Object(PHPUnit\Framework\TestResult))
#7 /Users/richardm/stripe/stripe-php/vendor/phpunit/phpunit/src/TextUI/TestRunner.php(653): PHPUnit\Framework\TestSuite->run(Object(PHPUnit\Framework\TestResult))
#8 /Users/richardm/stripe/stripe-php/vendor/phpunit/phpunit/src/TextUI/Command.php(144): PHPUnit\TextUI\TestRunner->run(Object(PHPUnit\Framework\TestSuite), Array, Array, true)
#9 /Users/richardm/stripe/stripe-php/vendor/phpunit/phpunit/src/TextUI/Command.php(97): PHPUnit\TextUI\Command->run(Array, true)
#10 /Users/richardm/stripe/stripe-php/vendor/phpunit/phpunit/phpunit(98): PHPUnit\TextUI\Command::main()
#11 /Users/richardm/stripe/stripe-php/vendor/bin/phpunit(123): include('/Users/richardm...')
#12 {main}
```


anniel-stripe (approved):

adaviloper:
I'm unsure if this change is why, but after upgrading from a version before this release, my Stripe exceptions are now seemingly being split by the new line character being added after including the entire stack trace.

I think this is breaking my logger since it's no longer receiving the format it's expecting. 

This is more or less a shot in the dark, but does this at all make sense as a side-effect here or should I start looking into other solutions?

adaviloper:
image

This is what I've been getting now after upgrading along with the rest of the stack trace which I've cropped out. You can see that the error being thrown is from the `CardException` class; not shown is that the reason is for an incorrect security code on the card.

It seems that at some point the way the exceptions are coerced into strings is including the entire stack trace and affecting the way DataDog is interpretting the error message.

prathmesh-stripe:
@adaviloper What version of `dogstatd` are you currently using? And what does your datadog agent configuration file?

adaviloper:
@prathmesh-stripe We're on 7.61.0. Are there any keys that are of particular interest? I'm not able to post the entire file.

adaviloper:
@prathmesh-stripe Appreciate the information. I'll go ahead and create an issue!

PR #1565, opened 2023-08-30
Enhancement: Use PHP 8.2 as leading PHP version

localheinz:
This pull request

- [x] uses PHP 8.2 as the leading PHP version when running jobs on GitHub Actions

💁‍♂️ PHP 8.1 is the oldest and PHP 8.2 is the newest PHP version with [active support](https://www.php.net/supported-versions.php) that we currently use in jobs on GitHub Actions. Both versions are way off from the lowest PHP version that `stripe/stripe-PHP` aims to be compatible with, so why not use the newest PHP version?


richardm-stripe:
Thanks for the contribution @localheinz! These are all very helpful and much appreciated.

richardm-stripe (approved):

PR #1566, opened 2023-08-30
Fix: Remove `squizlabs/php_codesniffer`

localheinz:
This pull request

- [x] removes `squizlabs/php_codesniffer`

💁‍♂️ This dependency does not appear to be used anywhere in this project.

PR #1567, opened 2023-08-31
Fix: Rename jobs

localheinz:
This pull request

- [x] renames GitHub Actions jobs to reveal the intent, not the tool used

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,8 +19,8 @@ on:
       - feature/**
 
 jobs:
-  php-cs-fixer:
-    name: PHP-CS-Fixer
+  coding-standards:
+    name: Coding Standards
 
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -19,8 +19,8 @@ on:
       - feature/**
 
 jobs:
-  php-cs-fixer:
-    name: PHP-CS-Fixer
+  coding-standards:
+    name: Coding Standards
```
localheinz, line 23: This jobs enforces coding standards.

It currently runs `friendsofphp/php-cs-fixer` in a step, but it could use an additional or a different tool, for example, `squizlabs/php_codesniffer` (see #1566).

.github/workflows/ci.yml
```diff
@@ -32,8 +32,8 @@ jobs:
         with:
           args: --format=txt --diff --dry-run --using-cache=no --verbose
 
-  phpstan:
-    name: PHPStan
+  static-code-analysis:
+    name: Static Code Analysis
```
localheinz, line 36: This jobs runs a static code analysis.

It currently runs `phpstan/phpstan` in a step, but it could use an additional or a different tool, for example, `vimeo/psalm`.

pakrym-stripe:
Hi @localheinz, thank you for your contribution.

I chatted with the team and we decided that it's quite useful for us to know which tool is used. With so many repositories it's hard to remember that `Static Code Analysis` means `php stan`

PR #1582, opened 2023-09-22
Generate Discount, SourceTransaction and use sections in more places

pakrym-stripe:
Auto generate Discount and SourceTransaction resources.


lib/Discount.php
```diff
@@ -1,23 +1,28 @@
...
+ * @property null|string|\Stripe\Customer $customer The ID of the customer associated with this discount.
+ * @property null|int $end If the coupon has a duration of <code>repeating</code>, the date that this discount will end. If the coupon has a duration of <code>once</code> or <code>forever</code>, this attribute will be null.
+ * @property null|string $invoice The invoice that the discount's coupon was applied to, if it was applied directly to a particular invoice.
+ * @property null|string $invoice_item The invoice item <code>id</code> (or invoice line item <code>id</code> for invoice line items of type='subscription') that the discount's coupon was applied to, if it was applied directly to a particular invoice item or invoice line item.
+ * @property null|string|\Stripe\PromotionCode $promotion_code The promotion code applied to create this discount.
  * @property int $start Date that the coupon was applied.
  * @property null|string $subscription The subscription that this coupon is applied to, if it is applied to a particular subscription.
  */
-class Discount extends StripeObject
+class Discount extends ApiResource
```
richardm-stripe, line 25: Is this correct/intentional?

This causes `Discount` to have e.g. a `.refresh()` method and an `.instance_url` method and I'm not sure that's intended.

richardm-stripe (approved):
Everything LGTM unless the `Discount extends StripeObject` -> `Discount extends ApiResource` was unintentional.

lib/Discount.php
pakrym-stripe: Yes, all resources have ApiResource base class in PHP. See [UsageRecordSummary](https://github.com/stripe/stripe-php/blob/7d0a90772fc1c179e370971264318208533324b9/lib/UsageRecordSummary.php#L16) for example.

PR #1612, opened 2023-11-27
Report usage of .save and StripeClient

richardm-stripe:
## Changelog
* Reports uses of the deprecated `.save` and of `StripeClient` in `X-Stripe-Client-Telemetry`. (You can disable telemetry via `\Stripe\Stripe::setEnableTelemetry(false);`, see the [README](https://github.com/stripe/stripe-php/blob/master/README.md#telemetry).)

## Details
I've added `usage` to several infrastructure methods:
  - Several underscored methods in `lib/ApiOperations/Request.php`. (Since they are underscored, they should not be considered part of the public interface). I've included a default value there for expedience sake, though. Most callsites for these methods have no usage to report.
  - `request` and `requestStream` in `ApiRequestor`. These are public, but I have added a default value so that it is non-breaking.
  - The private `_requestRaw` and `_requestRawStreaming` in `ApiRequestor`. No default value here, but they are private.
  - The `RequestTelemetry` constructor which is public (but I doubt it has much external use), but I've added a default value there.
  - I've modified `BaseStripeClient` to always pass in through `["stripe_client"]` as usage to `ApiRequestor`, and modified `->save` in `ApiOperations` to pass in `["save"]` through to the helpers in `Request.php`.

## Testing
I modified the existing test case for telemetry to use a method that would trigger "save" usage to be included. It mocks at the HTTP Client level, which is fine, because all the logic for telemetry is thankfully implemented at a higher level of abstraction in `ApiRequestor`.


anniel-stripe (approved):

lib/RequestTelemetry.php
```diff
@@ -9,18 +9,24 @@
...
      * @param string $requestId the request's request ID
      * @param int $requestDuration the request's duration in milliseconds
+     * @param string[] $usage names of tracked behaviors associated with this request
      */
-    public function __construct($requestId, $requestDuration)
+    public function __construct($requestId, $requestDuration, $usage = [])
     {
         $this->requestId = $requestId;
         $this->requestDuration = $requestDuration;
+        $this->usage = $usage;
```
anniel-stripe, line 30: Should we omit usage if empty here too?

PR #1630, opened 2024-01-11
Add paginated requests helper function and use in Search and All

helenye-stripe:
For the purpose of de-magicking PHP search / all. This simplifies what the codegen will have to generate per resource (e.g. for search the body will just call the function _requestPage.)

### DIFF
--- a/lib/ApiOperations/All.php
+++ b/lib/ApiOperations/All.php
@@ -19,19 +19,8 @@ trait All
      */
     public static function all($params = null, $opts = null)
     {
-        self::_validateParams($params);
         $url = static::classUrl();
 
-        list($response, $opts) = static::_staticRequest('get', $url, $params, $opts);
-        $obj = \Stripe\Util\Util::convertToStripeObject($response->json, $opts);
-        if (!($obj instanceof \Stripe\Collection)) {
-            throw new \Stripe\Exception\UnexpectedValueException(
-                'Expected type ' . \Stripe\Collection::class . ', got "' . \get_class($obj) . '" instead.'
-            );
-        }
-        $obj->setLastResponse($response);
-        $obj->setFilters($params);
-
-        return $obj;
+        return static::_requestPage($url, \Stripe\Collection::class, $params, $opts);
     }
 }
--- a/lib/ApiOperations/Request.php
+++ b/lib/ApiOperations/Request.php
@@ -18,9 +18,9 @@ trait Request
     {
         if ($params && !\is_array($params)) {
             $message = 'You must pass an array as the first argument to Stripe API '
-               . 'method calls.  (HINT: an example call to create a charge '
-               . "would be: \"Stripe\\Charge::create(['amount' => 100, "
-               . "'currency' => 'usd', 'source' => 'tok_1234'])\")";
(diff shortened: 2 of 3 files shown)


helenye-stripe:
Also assigning one additional person to gather more feedback on whether we want this or not!

anniel-stripe (approved):
Makes sense to me!

lib/ApiOperations/All.php
```diff
@@ -19,19 +19,8 @@ trait All
...
-        if (!($obj instanceof \Stripe\Collection)) {
-            throw new \Stripe\Exception\UnexpectedValueException(
-                'Expected type ' . \Stripe\Collection::class . ', got "' . \get_class($obj) . '" instead.'
-            );
-        }
-        $obj->setLastResponse($response);
-        $obj->setFilters($params);
-
-        return $obj;
+        return static::_requestPage($url, \Stripe\Collection::class, $params, $opts);
```
pakrym-stripe, line 24: Can we reuse the existing `_request` function? In most languages we try to use a single function for all streaming requests and one function for non-streaming ones.

richardm-stripe: `_requestPage` is `_static_request` plus some of the logic that needs to be included for list requests and search result requests, like calling `.addFilters` and some extra validation.

PHP has a little bit more to do inside the function definition than some of the other clients e.g. stripe-java, because stripe-java is doing type casts and does some of this at serialization time: https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/net/LiveStripeResponseGetter.java#L74-L77

Do you not like the idea of `_requestPage`, and prefer that the additional logic be included inline on each of the search and list functions?

pakrym-stripe: Let me ask it the other way. When API requestor comes to PHP are we going to still have a separate method? 

Why can't we do type-casts in PHP the same way as Java?

richardm-stripe: I think ultimately I would advocate something like `ResponseGetter.requestList` and `ResponseGetter.requestSearch` to `stripe-java` to avoid the type casts. We know statically that `Customer.list` is supposed to return a list, it feels wrong to lose and have to recover that information via type-cast inside the infrastructure.

The problem in stripe-java is that we have *one* interface that serves dual purposes: it's the *internal* interface for method definitions to communicate to the internal infrastructure AND it's the *user-facing* interface that we provide so that users can do mocking or otherwise override aspects of behavior. We would have to decouple these. The type casts are better than polluting the public interface, but they aren't ideal.

richardm-stripe: In any case _requestPage captures the existing approach of stripe-php and IMO makes the intent more clear and makes generation more straightforward. It is an internal method and can ~easily/non-breakingly be inlined if we decide to change approaches.

Do you prefer no helper and that the logic be inlined?

pakrym-stripe: Treat my comment here as non-blocking. Separate helper is fine with me but getting closer to our ideal implementation would've been even better.

richardm-stripe (approved):

PR #1635, opened 2024-01-16
Add deserialized object assertion to webhook test

pakrym-stripe:
Was doing some testing and realized we don't cover this scenario.


tests/Stripe/WebhookTest.php
```diff
@@ -37,6 +38,7 @@ public function testValidJsonAndHeader()
         $sigHeader = $this->generateHeader();
         $event = Webhook::constructEvent(self::EVENT_PAYLOAD, $sigHeader, self::SECRET);
         static::assertSame('evt_test_webhook', $event->id);
+        static::assertInstanceOf(\Stripe\Terminal\Reader::class, $event->data->object);
```
anniel-stripe, line 41: Fix PHPStan error?

pakrym-stripe: Done

anniel-stripe (approved):

PR #1646, opened 2024-02-13
More specific PHPDoc type for RequestOptions arrays on services

richardm-stripe:
## Summary
Defines and uses "array shape" types for "RequestOptions"
```
/**
 * @phpstan-type RequestOptionsArray array{api_key?: string, idempotency_key?: string, stripe_account?: string, stripe_version?: string, api_base?: string }
 * @psalm-type RequestOptionsArray = array{api_key?: string, idempotency_key?: string, stripe_account?: string, stripe_version?: string, api_base?: string }
 */
 ```
that are more specific than `array`.

In psalm, this is sufficient to prevent typos/misuses like `["api_kkey" => "sk_test_xyz"]` https://psalm.dev/r/7a73e0e48f

But in PHPStan this isn't: https://phpstan.org/r/cbdc5206-60c0-4104-8d59-9dd62d9e1bfe because, at least for the time being, "array shapes" are considered "unsealed" (see https://github.com/phpstan/phpstan/issues/8438).

### DIFF
--- a/lib/Service/AccountLinkService.php
+++ b/lib/Service/AccountLinkService.php
@@ -4,8 +4,12 @@
 
 namespace Stripe\Service;
 
-// /** @phpstan-import-type RequestOptionsArray from \Stripe\Util\RequestOptions */
-// /** @psalm-import-type RequestOptionsArray from \Stripe\Util\RequestOptions */
+/**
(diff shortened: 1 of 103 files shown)


richardm-stripe:
Marking this as ready for review. I don't think there's any reason not to merge this?

helenye-stripe (approved):
Thanks for taking this!

PR #1701, opened 2024-05-28
Added PHPDocs for `create`, `update`, `delete`, `all`, `retrieve` methods after moving them out of traits. 

prathmesh-stripe:
Moved `Create`, `Update`, `Delete`, `All`, `Retrieve` traits to resources.

### DIFF
--- a/lib/Account.php
+++ b/lib/Account.php
@@ -45,9 +45,6 @@ class Account extends ApiResource
 {
     const OBJECT_NAME = 'account';
 
-    use ApiOperations\All;
-    use ApiOperations\Create;
-    use ApiOperations\Delete;
     use ApiOperations\NestedResource;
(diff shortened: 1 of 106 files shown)


anniel-stripe (approved):

ramya-stripe:
Are we not planning to delete the files that define the different traits that we are now not using?

ramya-stripe:
And are we not planning to remove the update method from the Update trait?

prathmesh-stripe:
We can't remove the traits as there is a concern with these [overrides](https://github.com/stripe/sdk-codegen/blob/master/src/php/overrides.ts#L822-L824) that still use the traits. 

I'll have to look into these and change some behavior(or the override itself).

anniel-stripe (approved):

PR #1709, opened 2024-06-07
update recommended extensions and settings

xavdid-stripe:
Removing old extensions, adding recommended settings


.vscode/settings.json
```diff
@@ -4,12 +4,11 @@
...
+
   // Formatting
   "editor.formatOnSave": true,
-  "phpformatter.composer": true,
-  "phpformatter.arguments": [""],
-
-  // Linting
-  "phpcs.enable": true,
-  "phpcs.standard": "PSR2"
+  "intelephense.format.braces": "psr12"
```
xavdid-stripe, line 13: I think this is the same, but i'm not positive. I did see some minor bracket motion when saving a file, so i'm not sure.

.vscode/extensions.json
```diff
@@ -1,9 +1,8 @@
 {
   "recommendations": [
-    "felixfbecker.php-intellisense",
-    "ikappas.phpcs",
```
xavdid-stripe, line 4: this is deprecated, so I pulled it

.vscode/extensions.json
```diff
@@ -1,9 +1,8 @@
 {
   "recommendations": [
-    "felixfbecker.php-intellisense",
```
xavdid-stripe, line 3: this has been removed from the marketplace

.vscode/extensions.json
```diff
@@ -1,9 +1,8 @@
...
-    "ikappas.phpcs",
-    "emallin.phpunit",
-    "EditorConfig.editorconfig",
-    "sophisticode.php-formatter"
```
xavdid-stripe, line 7: also deprecated

phpstan.neon.dist
```diff
@@ -11,3 +11,7 @@ parameters:
 		- '#Unsafe usage of new static\(\).#'
 
 	reportUnmatchedIgnoredErrors: false
+
+	paths:
+		- lib
```
xavdid-stripe, line 16: needed for the phpstan extension

xavdid-stripe (commented):
Ok, please double check this one - it's the ecosystem I know least about. Feel free to push back!

.vscode/settings.json
prathmesh-stripe: Can you tell me the files that were changed? I am wondering how this change would affect the formatter in our CI. I doubt they support PSR12.

xavdid-stripe: This PR didn't change any files directly - stuff would get changed as vscode users format/save things. 

That said, this isn't a great solution. It doesn't seem like a 1:1 replacement and not everyone on the team uses vscode. I wonder if we don't use vscode for formatting at all and lean on `make format` instead, which will be consistent across editors and CI.

In JS and other languages, the formatter and the vscode extension use the same program under the hood, so they always agree. I'm not sure PHP can have the same setup.

prathmesh-stripe: Let's see if this would break anything in the future. We can anyways fallback on the `make fmt` for formatting.

prathmesh-stripe (approved):

PR #1713, opened 2024-06-14
Removed the protected method _searchResource as it is no longer used

prathmesh-stripe:
## Changelog
* Remove the protected method `_searchResource` as it is no longer used.


anniel-stripe (approved):

ramya-stripe:
Small tweak to the changelog:

> Removed `_searchResource` method in favor of `search` on resources Charge, Customer, Invoice, PaymentIntent, Price, Product, and Subscription. 

Can we also include code snippets for before and after? Especially because the method signature is different for the two.  We can also call out that you don't need to pass the search url anymore

ramya-stripe:
> Can we also include code snippets for before and after? Especially because the method signature is different for the two. We can also call out that you don't need to pass the search url anymore

I stand corrected. Since this was not a public method meant to be used by users, we don't need the before/after

PR #1732, opened 2024-08-09
Introduce StripeClient.rawRequest as a canonical way to request Stripe API

prathmesh-stripe:
Introducing `StripeClient.rawRequest`, a new way to request Stripe APIs. Previously available only in beta, `StripeClient.rawRequest` now has general availability. 

If you
- would like to send a request to an undocumented API (for example you are in a private beta)
- prefer to bypass the method definitions in the library and specify your request details directly
- used the method `_request` on `\Stripe\ApiOperation\Request` trait to specify your own requests. 

you can use the `rawRequest` method on the StripeClient.

## Changelog
- Adds the ability to make raw requests to the Stripe API, by providing an HTTP method and url. This is an alternative to using the trait `\Stripe\ApiOperation\Request` to make custom requests, which is discouraged and will be broken in a future major version.
```php
$stripe = new \Stripe\StripeClient('sk_test_xyz');
$response = $stripe->rawRequest('post', '/v1/undocumented_endpoint', [
  "caveat": "emptor"
], [
  "stripe_version" => "2024-06-20",
]);
// $response->body is a string, you can call $stripe->deserialize to get a \Stripe\StripeObject.
$obj = $stripe->deserialize($response->body);

// For GET requests, the params argument must be null, and you should write the query string explicitly.
$get_response = $stripe->rawRequest('get', '/v1/beta_endpoint?caveat=emptor', null, [
  "stripe_version" => "2022-11_15",
]);
```

### DIFF
--- a/lib/BaseStripeClient.php
+++ b/lib/BaseStripeClient.php
@@ -164,6 +164,42 @@ class BaseStripeClient implements StripeClientInterface, StripeStreamingClientIn
     }
 
     /**
+     * Sends a raw request to Stripe's API. This is the lowest level method for interacting
+     * with the Stripe API. This method is useful for interacting with endpoints that are not
+     * covered yet in stripe-php.
+     *
+     * @param 'delete'|'get'|'post' $method the HTTP method
+     * @param string $path the path of the request
+     * @param array $params the parameters of the request
+     * @param array $opts the special modifiers of the request
+     *
+     * @return \Stripe\ApiResponse
+     */
+    public function rawRequest($method, $path, $params, $opts)
+    {
+        if ('post' !== $method && null !== $params) {
+            throw new Exception\InvalidArgumentException('Error: rawRequest only supports $params on post requests. Please pass null and add your parameters to $path');
+        }
+        $headers = [];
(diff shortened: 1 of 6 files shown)


lib/BaseStripeClient.php
```diff
@@ -163,6 +163,42 @@ public function request($method, $path, $params, $opts)
...
+     * @param string $path the path of the request
+     * @param array $params the parameters of the request
+     * @param array $opts the special modifiers of the request
+     *
+     * @return \Stripe\ApiResponse
+     */
+    public function rawRequest($method, $path, $params, $opts)
+    {
+        if ('post' !== $method && null !== $params) {
+            throw new Exception\InvalidArgumentException('Error: rawRequest only supports $params on post requests. Please pass null and add your parameters to $path');
```
xavdid-stripe, line 181: I like the direction of this error message, but `Please pass null` could be more descriptive. Maybe something like

> raw $method requests don't support params (expected $params to be `null`). Either make a POST request instead or pass `null` to the params argument

Feel free to tweak that further though!

xavdid-stripe (approved):
looks good! I appreciate the cleanup of the request building logic, too

PR #1734, opened 2024-08-13
Get Tax Id List Supported by Stripe

talha-core:
I think it would be nice if just have the function to get the list of supported tax types as an array


prathmesh-stripe:
This file is autogenerated and we can't add this method to the TaxId resource. However you could use the following snippet in your code to get the list of all TaxID.

```php
function getTaxIds()
{
    $taxIdClass = new ReflectionClass(\Stripe\TaxId::class);

    return array_filter($taxIdClass->getConstants(), function ($key) {
        // For PHP 8+
        return str_starts_with($key, 'TYPE_');
        // For PHP 7 or older
        // return substr($key, 0, 5 ) === "TYPE_";

    }, ARRAY_FILTER_USE_KEY);
}
```

Issue #1739, opened 2024-08-21
PHP 8.0 incompatibilities:


Sangi0:
### Describe the bug

FILE: lib/Stripe/lib/HttpClient/CurlClient.php
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
FOUND 1 ERRORS AND 0 WARNINGS AFFECTING 1 LINES
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 202 | ERROR | Since PHP 7.0, functions inspecting arguments, like func_get_args(), no longer report the original value as passed to a parameter, but will instead provide the current value. The parameter "$method" was changed on line 198.

### To Reproduce

Download MultiVendorX plugin >> Install the plugin PHP incompatibility >> Start scan and you will get the error 

### Expected behavior

There should not be any compatibility errors

### OS

macOS

### PHP version

PHP 8.0

### Library version

15.7.0

### API version

2024-06-20

richardm-stripe:
Hello @Sangi0, thanks for the report. We're exploring a fix for this in https://github.com/stripe/stripe-php/pull/1741

PR #1741, opened 2024-08-26
Avoid using func_get_args

ramya-stripe:
Resolves the incompatibility issues with PHP 8.0 as described in https://github.com/stripe/stripe-php/issues/1739

### DIFF
--- a/lib/HttpClient/CurlClient.php
+++ b/lib/HttpClient/CurlClient.php
@@ -199,7 +199,7 @@ class CurlClient implements ClientInterface, StreamingClientInterface
 
         $opts = [];
         if (\is_callable($this->defaultOptions)) { // call defaultOptions callback, set options to return value
-            $opts = \call_user_func_array($this->defaultOptions, \func_get_args());
+            $opts = \call_user_func_array($this->defaultOptions, [$method, $absUrl, $headers, $params, $hasFile]);
             if (!\is_array($opts)) {
                 throw new Exception\UnexpectedValueException('Non-array value returned by defaultOptions CurlClient callback');
             }


lib/HttpClient/CurlClient.php
```diff
@@ -199,7 +199,7 @@ private function constructRequest($method, $absUrl, $headers, $params, $hasFile)
 
         $opts = [];
         if (\is_callable($this->defaultOptions)) { // call defaultOptions callback, set options to return value
-            $opts = \call_user_func_array($this->defaultOptions, \func_get_args());
+            $opts = \call_user_func_array($this->defaultOptions, [$method, $absUrl, $headers, $params, $hasFile]);
```
richardm-stripe, line 202: There's a really obscure case where this is breaking. For users on PHP < 7, `\call_user_func_array` returns the *original* value of `$method` (i.e. before we do `$method = \strtolower($method)` on line 198), but this change will change it to lower case it. Since `$this->defaultOptions` is a potentially-user-defined callback we can't be 100% certain this won't break somebody, but this would only happen in an instance where somebody was introducing a non-lower-cased verb into our request infrastructure (custom requests?)

I think this case is unlikely but maybe it would be better to postpone this to the major?

ramya-stripe: custom requests are only available on the beta SDKs, not the GA ones - so we should be good?

richardm-stripe: Unfortunately you can initialize `StripeClient` or `ApiRequestor` directly and call `->request` (happens [in the wild](https://github.com/gfkacid/dripit-poc/blob/fc73d1b13830708f0c2945b75aca98c93955d1c9/backend/app/Http/Controllers/StripeController.php#L29)). I don't think we ever documented this though, so it's probably fine?

ramya-stripe: fair enough, will close this out and this can go in the next month's major release

PR #1750, opened 2024-09-24
Adding support for the new Usage Billing APIs

prathmesh-stripe:
Adding support for the new Usage Billing APIs


lib/Stripe.php
```diff
@@ -58,7 +64,7 @@ class Stripe
     /** @var float Initial delay between retries, in seconds */
     private static $initialNetworkRetryDelay = 0.5;
 
-    const VERSION = '15.10.0';
+    const VERSION = '0.8.0';
```
ramya-stripe, line 67:
```suggestion
    const VERSION = '15.10.0';
```

lib/ApiRequestor.php
```diff
@@ -241,6 +245,43 @@ private static function _specificAPIError($rbody, $rcode, $rheaders, $resp, $err
...
+                    $rcode,
+                    $rbody,
+                    $resp,
+                    $rheaders,
+                    $code
+                );
+
+            // The end of the section generated from our OpenAPI spec
+            default:
+                return self::_specificV1APIError($rbody, $rcode, $rheaders, $resp, $errorData);
```
ramya-stripe, line 281: Shouldnt we be checking for $code == "invalid_fields" here before falling back to v1 errors?

lib/Webhook.php
```diff
@@ -22,7 +22,7 @@ abstract class Webhook
      * @throws Exception\UnexpectedValueException if the payload is not valid JSON,
      * @throws Exception\SignatureVerificationException if the verification fails
      *
-     * @return Event the Event instance
+     * @return Event the Events instance
```
ramya-stripe, line 25: The class name is `Event`. Why are we pluralizing here?

lib/ThinEvent.php
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Stripe;
+
+/**
+ * @property string             $id Unique identifier for the event.
+ * @property string             $type
+ * @property string             $created
+ * @property null|string        $context
+ * @property null|RelatedObject $related_object
+ * @property null|Reason        $reason
```
ramya-stripe, line 11: We can add better doc strings here. Refer to https://github.com/stripe/stripe-node/blob/934e0fa98ccdfa7b5fb715f4bc4e9f8e3b76bb4e/types/V2/Events.d.ts#L37

PR #1752, opened 2024-09-25
Fixes requests with no params and responses with no response

jar-stripe:
### Why?
When testing the meter event stream examples, I saw error code 400 return from creating a MeterEventSession (using MeterEventSessionService.create()) the issue was that we were sending an encoded empty array [] instead of no body.  The fix here is to allow null params through, and then test for null where we attempt to iterate thru them.

After fixing this, I saw that creating a MeterEventStream (using MeterEventSessionService) was raising an exception when trying to setLastResponse.  The issue here is that the response from the server is empty, which json_decode decodes as an empty array, which is then interpreted as a list in `convertToStripeObject`.  The fix here (I think) is to not interpret an empty array as a list.

### What?
- changed ApiRequestor.request to not use an array as the default for $params, and to test for null before iterating thru $params
- added a test for empty array to convertToStripeObject (in Util) so that this case falls thru the is_array branch and returns an empty StripeObject object


jar-stripe:
The indented comments were automatic and I couldn't figure out how to get VS code to knock it off.

helenye-stripe:
is the test supposed to be in this PR?

lib/ApiRequestor.php
```diff
@@ -129,7 +129,6 @@ private static function _encodeObjects($d)
      */
     public function request($method, $url, $params = null, $headers = null, $apiMode = 'v1', $usage = [])
     {
-        $params = $params ?: [];
```
helenye-stripe, line 132: re: your comment from the other PR, do we want to preserve sending of empty bodies? I'm confused why we've never seen this shape of failure before, did we always pass `[]`?

prathmesh-stripe: This needs to be "[]" because nulling out is risky.
Screenshot 2024-09-27 at 10 54 41 AM

And sending a curl request with `[]` body seems to work for me. 
Screenshot 2024-09-27 at 10 55 20 AM

prathmesh-stripe: Although I am facing the same issue that you are facing.

jar-stripe:
> > added a test for empty array to convertToStripeObject (in Util) so that this case falls thru the is_array branch and returns an empty StripeObject object
> 
> is the test supposed to be in this PR?

sorry, poor choice of words.  by test i meant a conditional test, as part of an if statement.

jar-stripe:
This was fixed in another way in another branch.  Closing.

PR #1754, opened 2024-09-27
Fixed empty param for v2 requests

prathmesh-stripe:
Creating a meter event session was failing(400) because we were trying to send an invalid json body in the request. 
```php
$stripeClient->v2->billing->meterEventSession->create([]);
```

```php
json_encode([]) // '[]' - string brackets
json_encode(null) // 'null' - String null
```

In order to fix this, I added a check to skip encoding params when the params are empty.

### DIFF
--- a/lib/HttpClient/CurlClient.php
+++ b/lib/HttpClient/CurlClient.php
@@ -209,6 +209,13 @@ class CurlClient implements ClientInterface, StreamingClientInterface
                 return [$absUrl, $params];
             }
             if ('v2' === $apiMode) {
+                if (\is_array($params) && 0 === \count($params)) {
+                    // Send a request with empty body if we have no params set
+                    // Setting the second parameter as null prevents the CURLOPT_POSTFIELDS
+                    // from being set with the '[]', which is result of `json_encode([]).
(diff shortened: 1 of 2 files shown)


lib/HttpClient/CurlClient.php
```diff
@@ -209,6 +209,13 @@ private function constructUrlAndBody($method, $absUrl, $params, $hasFile, $apiMo
                 return [$absUrl, $params];
             }
             if ('v2' === $apiMode) {
+                if (\is_array($params) && 0 === \count($params)) {
+                    // Send a request with empty body if we have no params set
+                    // Setting the second parameter as null prevents the CURLOPT_POSTFIELDS
+                    // from being set with the '[]', which is result of `json_encode([]).
+                    return [$absUrl, null];
+                }
+
```
helenye-stripe, line 218: Is it possible to add a test to verify this?

prathmesh-stripe: I tested this in dogfooding and forgot to add an actual test in this PR.

helenye-stripe (approved):

PR #1757, opened 2024-09-27
Fix Event typing

prathmesh-stripe:
Events have a changed object tag`v2.core.event`. This PR allows the types to be correctly assigned after the change.


jar-stripe (approved):
LGTM!

PR #1758, opened 2024-09-27
Fix type of Billing Meter Event

prathmesh-stripe:
fetchRelatedObject could not correctly return v1 objects. Therefore I had to manually add that type in v2mapping.

```php
$baseEvent = $stripeClient->v2->core->events->retrieve('evt_test_65RD9HQIXX53rfbiu8116QQCIX1TSQbbOu5RHvml0KWLTM');

//print_r($baseEvent);
if ($baseEvent instanceof \Stripe\Events\V1BillingMeterErrorReportTriggeredEvent) {
    print_r($baseEvent->fetchRelatedObject()); // Returns StripeObject. Should return \Stripe\Billing\Meter
}
```

Followup to move this to codegen: https://jira.corp.stripe.com/browse/DEVSDK-2204

### DIFF
--- a/lib/Util/ObjectTypes.php
+++ b/lib/Util/ObjectTypes.php
@@ -161,7 +161,7 @@ class ObjectTypes
      */
     const v2Mapping = [
         // V1 Class needed for fetching the right related object
-        // TODO: Make a more standardized fix in codegen for all languages
+        // TODO: https://go/j/DEVSDK-2204 Make a more standardized fix in codegen for all languages
         \Stripe\Billing\Meter::OBJECT_NAME => \Stripe\Billing\Meter::class,
 
         // v2 object classes: The beginning of the section generated from our OpenAPI spec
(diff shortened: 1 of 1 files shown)


lib/Util/ObjectTypes.php
```diff
@@ -160,6 +160,10 @@ class ObjectTypes
      * @var array Mapping from v2 object types to resource classes
      */
     const v2Mapping = [
+        // V1 Class needed for fetching the right related object
```
jar-stripe, line 163: there's likely a way to fix this in codegen (like what I did for java); probably not a blocker for this but consider logging a jira ticket to revisit later.

jar-stripe:
very small nit but can you remove that commented print_r from the PR description?

xavdid-stripe (approved):
this is fine, but could have been fixed in https://github.com/stripe/sdk-codegen/pull/1777

We should probably generalize that approach

lib/Util/ObjectTypes.php
prathmesh-stripe: Already filed one

jar-stripe (approved):
Looks good, pending a couple of non-blocking comments!

PR #1759, opened 2024-09-28
Fall back to InvalidRequestException when code is `invalid_fields`

helenye-stripe:
For v2 APIs, when an error has code `invalid_fields`, it should fail with `InvalidRequestException`. This addresses that for V2, but V1 behavior remains unchanged.

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -278,6 +278,10 @@ class ApiRequestor
 
             // The end of the section generated from our OpenAPI spec
             default:
+                if ('invalid_fields' === $code) {
+                    $param = isset($errorData['param']) ? $errorData['param'] : null;
+                    return Exception\InvalidRequestException::factory($msg, $rcode, $rbody, $resp, $rheaders, $code, $param);
+                }
                 return self::_specificV1APIError($rbody, $rcode, $rheaders, $resp, $errorData);
         }
     }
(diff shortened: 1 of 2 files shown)


lib/ApiRequestor.php
```diff
@@ -278,6 +278,10 @@ private static function _specificV2APIError($rbody, $rcode, $rheaders, $resp, $e
 
             // The end of the section generated from our OpenAPI spec
             default:
+                if ('invalid_fields' === $code) {
+                    $param = isset($errorData['param']) ? $errorData['param'] : null;
+                    return Exception\InvalidRequestException::factory($msg, $rcode, $rbody, $resp, $rheaders, $code, $param);
+                }
```
prathmesh-stripe, line 284: Wouldn't it better if we moved this to `_specificV1APIError` under 400 status code?

prathmesh-stripe: That would make it consistent with all languages except node.

helenye-stripe: That would be a pretty breaking change for a lot of users - I think this minimizes the scope a lot more. I don't know that it's worth changing how exceptions work for all v1 users here for not much gain (imo). Curious for your opinion, though.

prathmesh-stripe: I'd assume that they would be catching a more general `ApiErrorException` anyways. It would be nicer for PHP to also have more accurate exception type and a more consistent error responses with the rest of the languages. 

This is a major release with probably a migration guide too so I'd say most people wouldn't mind putting in the effort to rename and make the change.

helenye-stripe: I'm going to ask Remi real quick, because my feel is that this is pretty breaking.

helenye-stripe:
Closing because I misread the code here -- it does fall back to `InvalidRequestException`!

PR #1762, opened 2024-10-01
Better property descriptions for thin event class

prathmesh-stripe:
Added descriptions to class properties in new ThinEvent class. Removed old ThinEvent class in PHP.


lib/ThinEvent.php
```diff
@@ -3,12 +3,13 @@
...
- * @property string             $created
- * @property null|string        $context
- * @property null|RelatedObject $related_object
- * @property null|Reason        $reason
+ * @property string             $id       Unique identifier for the event.
+ * @property string             $type     The type of the event.
+ * @property string             $created  Time at which the object was created.
+ * @property null|string        $context  Authentication context needed to fetch the event or related object.
+ * @property null|RelatedObject $related_object // TODO Check if this can be removed and fix the tests
+ * @property null|Reason        $reason   Reason for the event.
```
helenye-stripe, line 11: Are reason and livemode available in ThinEvent? I thought it was only the above five.

prathmesh-stripe: I'll remove those.

helenye-stripe (approved):

Issue #1787, opened 2024-11-18
RFC: Improve PHP types hinting for method parameters and class properties


prathmesh-stripe:
:wave: Hello PHP community, we are looking for community feedback on improvement we are working on. 
We are extending type hints that exist in Stripe-PHP SDK. New type hints will be compatible with PHPStan. Type hint improvements we are planning to implement so far:

## Create/Update/Retrieve/Delete/All/Search parameters

Existing parameters array in Stripe functions do not specify the keys of the array. 
```php
* @param null|array $params
```

After the changes, you will be able to know the keys that you can pass without switching out of your IDE. 
```php
* @param null|array{customer:string, components: array} $params
```

You will see hints in IDE(PHPStorm used in screenshot below) when writing array parameters for the Stripe API methods. 
<img width="417" alt="PHPStorm IDE with array type hints" src="https://github.com/user-attachments/assets/e914dcda-354f-4df2-b82e-217ad931e71d">

## Class properties
We will change the type of class properties from `StripeObject` to something more specific. 

For example: Invoice settings is defined as a StripeObject in Customer resource. 

https://github.com/stripe/stripe-php/blob/bae10cd799404f0f4862ec03810c5ff8ca634b30/lib/Customer.php#L25

After adding phpstan types in PHP docs, you will be able to call `custom_fields` and `rendering_options` on `customer->invoice_settings` without PHPStan complaining. 
```php
// New
/** 
 * @phpstan-type CustomerInvoiceSettingsCustomField object{name: string, value: string}&StripeObject
 * @phpstan-type CustomerInvoiceSettingsRenderingOptions object{amount_tax_display?: string, template?: string}&StripeObject
 * @phpstan-type CustomerInvoiceSettings object{custom_fields: array<CustomerInvoiceSettingsCustomField>, rendering_options: CustomerInvoiceSettingsRenderingOptions}&StripeObject
 * 
 * @property null|CustomerInvoiceSettings $invoice_settings
 */
```

## Conclusion
Existing users shouldn't be affected by our addition of types in PHPDocs. These additions will improve development experience for you.  
If you have more ideas about what you would like to see in our PHP SDK regarding types and type hinting, please add your suggestions in this thread.

prathmesh-stripe:
We released these changes as part of our beta release: https://github.com/stripe/stripe-php/releases/tag/v16.5.0-beta.1

Please continue sharing your feedback so we continue to improve our PHP SDK.

Issue #1740, opened 2024-08-23
Add specific classes for nested objects (instead of `StripeObject`)


momala454:
### Describe the bug

This is what the spec says for `Price->recurring`

` * @property null|\Stripe\StripeObject $recurring The recurring components of a price such as <code>interval</code> and <code>usage_type</code>.`

So the IDE have no idea what's inside

According to the doc, we know what's inside https://docs.stripe.com/api/prices/object
![image](https://github.com/user-attachments/assets/3afa9f2b-aa4d-4854-975b-05a8e43d7029)

### To Reproduce

Just lookup doc

### Expected behavior

Complete doc

### OS

windows

### PHP version

8.3

### Library version

v15.7.0

### API version

[2024-06-20](https://docs.stripe.com/upgrades#2024-06-20)

richardm-stripe:
Hello @momala454, thanks for writing in!

The Openapi spec actually does have a definition for Price.recurring, and we have full types for this in [other libraries](https://github.com/stripe/stripe-node/blob/master/types/Prices.d.ts#L109) but the problem our generator for stripe-php only generate classes for top-level resources and doesn't generate classes yet for the nested data structures inside.

This is definitely on our radar and an improvement we plan to make to the library, but I am reclassifying as a feature request and editing the issue name! We'll keep you posted.

prathmesh-stripe:
@momala454 We released these changes as part of our beta release: https://github.com/stripe/stripe-php/releases/tag/v16.5.0-beta.1

Please continue sharing your feedback so we continue to improve our PHP SDK.

PR #1794, opened 2024-12-11
Improved php type hints

prathmesh-stripe:
Related Issues:
https://github.com/stripe/stripe-php/issues/1787
https://github.com/stripe/stripe-php/issues/1740

### Adds Create/Update/Retrieve/Delete/All/Search parameters

You will now be able to get type hints of the keys that can passed without switching out of your IDE. Eg.
```php
* @param null|array{customer:string, components: array} $params
```

PHPStorm IDE with array type hints

### Updated StripeObject class properties
We changed the type of class properties from `StripeObject` to something more specific. 

For example: Invoice settings was defined as a StripeObject in Customer resource. 

https://github.com/stripe/stripe-php/blob/bae10cd799404f0f4862ec03810c5ff8ca634b30/lib/Customer.php#L25

Now you will be able to reference `custom_fields` and `rendering_options` on `customer->invoice_settings` without PHPStan complaining. 
```php
* @property object{custom_fields: null|object{name: string, value: string}&\Stripe\StripeObject&\stdClass[], default_payment_method: null|string|\Stripe\PaymentMethod, footer: null|string, rendering_options: null|object{amount_tax_display: null|string, template: null|string}&\Stripe\StripeObject&\stdClass}&\Stripe\StripeObject&\stdClass $invoice_settings
 */
```


helenye-stripe:
Can we add a changelog entry?

helenye-stripe (approved):

PR #1801, opened 2025-01-08
Add justfile, remove coveralls, and fix AUTOLOAD in CI

xavdid-stripe:
### Why?

In an effort to modernize and simplify our local tooling, we're moving our dev commands from makefiles to [justfiles](https://github.com/casey/just). This is intended to be mostly a drop-in replacement, but some command names may change per standardization efforts.

Namely, I renamed `fmt` to `format`, added `lint` in place of `phpstan`, and swapped `vendor` to `install`.

One actual change is that dependencies won't auto-install before running tests like they did before. Users will have to `just install` before `just test`.

Also, I didn't add a ton of docstrings, since the recipes are mostly self explanatory. But, we could add some. The recipes are also grouped, so the help output looks like:

```
Available recipes:
    [useful]
    install            # install vendored dependencies; only used locally
    test *options
    format *options

    [CI]
    ci-test
    format-check
    lint *options

    [misc]
    phpstan-baseline
    phpdoc
```

Without groups, output of `just` is:

```
Available recipes:
    install            # install vendored dependencies; only used locally
    test *options
    ci-test
    format *options
    format-check
    lint *options
    phpstan-baseline
    phpdoc
```

Which isn't bad either. I could go either way on the groups. If nothing else, I think it makes the `justfile` itself harder to grok, so i'm fine to pull them too.

### CI

Also, there was an issue where we weren't correctly setting the `AUTOLOAD` env var, so we were never running tests with `autoload=1`. The justfile flagged this issue and I changed the way the matrix was loaded to acommodate it.

### What?

- added `justfile`
- added warning to top of makefile
- tweaked CI to use `just` instead of `make`
- fixed CI so AUTOLOAD was correctly set
- removed PHPstan from a matrix and just hardcoded the version we wanted.

### See Also

- [DEVSDK-2325](https://go/j/DEVSDK-2325)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,8 +128,8 @@ jobs:
       - name: Run test suite
         run: |
           php --version
-          make ci-test
-          just ci-test
+          echo "autoload is: $AUTOLOAD"
+          AUTOLOAD=1 just ci-test
 
   publish:
     # Doesn't actually publish. The publish happens via a packagist webhook configured in the Github UI. But we still display a message here
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,8 +73,6 @@ jobs:
         run: just lint
 
   tests:
-    name: Tests
-
     runs-on: ubuntu-latest
 
     strategy:
@@ -93,6 +91,7 @@ jobs:
           - "8.0"
           - "8.1"
           - "8.2"
+    name: Tests (php@${{ matrix.php-version }}, AUTOLOAD=${{ matrix.autoload }})
 
     steps:
       - uses: extractions/setup-just@v2
@@ -126,14 +125,11 @@ jobs:
       - uses: stripe/openapi/actions/stripe-mock@master
 
       - name: Run test suite
-        env:
-          AUTOLOAD: ${{ matrix.autoload }}
         # --yes skips the confirmation dialogue for the CI test, which modifies files
         run: |
           php --version
           echo "autoload is: $AUTOLOAD"
-          make ci-test
-          just --yes ci-test
+          just --yes ci-test ${{ matrix.autoload }}
 
   publish:
     # Doesn't actually publish. The publish happens via a packagist webhook configured in the Github UI. But we still display a message here
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,7 +128,6 @@ jobs:
         # --yes skips the confirmation dialogue for the CI test, which modifies files
         run: |
           php --version
-          echo "autoload is: $AUTOLOAD"
           just --yes ci-test ${{ matrix.autoload }}
 
   publish:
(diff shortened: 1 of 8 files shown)


composer.json
```diff
@@ -1,46 +1,33 @@
...
-      "Stripe\\": "lib/"
-    }
-  },
-  "autoload-dev": {
-    "psr-4": {
-      "Stripe\\": [
-        "tests/",
-        "tests/Stripe/"
-      ]
-    }
```
prathmesh-stripe, line 39: Removing the autoload is causing the build failures. 
When you run `ci-test` with `AUTOLOAD=0` it [modifies](https://github.com/stripe/stripe-php/blob/master/build.php#L8-L14) the `composer.json` to get rid of these autoloads. Subsequently running `ci-test` without restoring the file locally is what is causing your build failures.

prathmesh-stripe (approved):

.github/workflows/ci.yml
```diff
@@ -37,18 +37,14 @@ jobs:
...
-
     steps:
+      - uses: extractions/setup-just@v2
       - uses: actions/checkout@v3
 
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: ${{ matrix.php-version }}
+          php-version: "8.2"
```
xavdid-stripe, line 47: Since we're not running `phpstan` across multiple PHP versions, I just hardcoded the value we're using

.github/workflows/ci.yml
```diff
@@ -69,19 +65,17 @@ jobs:
...
 
   tests:
-    name: Tests
-
     runs-on: ubuntu-latest
 
     strategy:
       fail-fast: false
       matrix:
-        env:
```
xavdid-stripe, line 82: this never worked as intended: `env` doesn't actually set the environment, it just sets a github variable called `env`

.github/workflows/ci.yml
```diff
@@ -69,19 +65,17 @@ jobs:
...
-          - AUTOLOAD=0
-          - AUTOLOAD=1
+        autoload:
```
xavdid-stripe, line 76: because `just` recipes can take arguments, we can replace the env-looking value with the actual value we want and pass it in directly

.github/workflows/ci.yml
```diff
@@ -92,15 +86,16 @@ jobs:
           - "8.0"
           - "8.1"
           - "8.2"
+    name: Tests (php@${{ matrix.php-version }}, AUTOLOAD=${{ matrix.autoload }})
```
xavdid-stripe, line 89: This will show a nicer name now that we're not using the full `AUTOLOAD` string

xavdid-stripe:
Because the names of the CI jobs changed, there's a mismatch on the required tests. We'll have to bypass some branch protections in PHP during the cutover, but should be smooth sailing after that.

Issue #1815, opened 2025-02-12
`advice_code` not defined in `StripeError` (incorrect release notes for v16.5.0)


LeonMelis:
### Describe the bug

The  v16.5.0 release notes states that support for `advice_code` in the `StripeError` object is added:

> Add support for `advice_code` on `StripeError`

However, it does not seem to be actually added. In fact, searching for `advice_code` in the entire repository only yields one occurrence: the release notes of 16.5.0.

Also note that the release notes mention `StripeError`, though in this PHP lib the class is known as `ErrorObject`.

See: https://github.com/stripe/stripe-php/releases/tag/v16.5.0

This field should be added to the PHPDoc of the ErrorObject class, so IDEs and linters can use this.

While we are at it: it would be really helpful to define the `$last_payment_error` in `PaymentIntent.php` as being an `ErrorObject`, instead of a generic `StripeObject`, again for code completion and linters.

### To Reproduce

- Fetch version 16.5.0 (or newer, at least up to 05c7c3a8a15b1bc396f09d17c88539c0db3d3255)
- Notice how `/lib/ErrorObject.php` does not define the `advice_code` field in the class description (PHPDoc)
- Notice how in the entire project `advice_code` is only mentioned once: in the release notes of v16.5.0

### Expected behavior

The field `advice_code` should be defined in the PHPDoc of class `ErrorObject` in `lib/ErrorObject.php`

And if possible, the field `$last_payment_error` in `PaymentIntent.php` should be defined as an `ErrorObject` instead of a generic `StripeObject`

### Code snippets

```PHP

```

### OS

n/a

### PHP version

n/a

### Library version

stripe-php v16.5.0

### API version

n/a

jar-stripe:
Hi @LeonMelis thanks for the report, and very sorry for the inconvenience here!  Let me look into this and see what is going on.  Will be back in touch soon!

jar-stripe:
Hi @LeonMelis , sorry for the delay in getting back to you here!  It looks like this field was not supposed to be included in our changelog.  We are fixing this (see https://github.com/stripe/stripe-php/pull/1819); if you need access to this field, if it is present in the response it will be available in the error values so you should be able to access it by name or by calling `values()` on the error object.  Please let me know how that goes, and sorry again for the trouble here!

jar-stripe:
[...]

PR #1819, opened 2025-03-02
Fixed changelog

prathmesh-stripe:
### Why?
Since PHP is untyped, the diff generator mistakenly generated changelog entries.

### See Also
https://github.com/stripe/stripe-php/issues/1815


jar-stripe (approved):
LGTM

PR #1822, opened 2025-03-13
Updated PHP CS Fixer

prathmesh-stripe:
### Why?
We are using a version of PHP CS Fixer that is 3 year old. 

### What?
 Updated `composer.json`, `.github/workflows/ci.yaml` and ran `just format`.


.github/workflows/ci.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: PHP-CS-Fixer
-        uses: docker://oskarstark/php-cs-fixer-ga:3.4.0
+        uses: docker://oskarstark/php-cs-fixer-ga:latest
```
jar-stripe, line 31: can we pin to a specific version?  latest makes me nervous, as it can lead to broken builds at the worst possible time

composer.json
```diff
@@ -22,7 +22,7 @@
   },
   "require-dev": {
     "phpunit/phpunit": "^5.7 || ^9.0",
-    "friendsofphp/php-cs-fixer": "3.5.0",
+    "friendsofphp/php-cs-fixer": "^3",
```
jar-stripe, line 25: similar to above; can we pin to within a version range where nonbreaking changes are expected?

jar-stripe:
All of the php file changes are from `just format`, right?

.github/workflows/ci.yml
prathmesh-stripe: Last pinned release was in 2023 and our builds would still fail on that version. I've [asked](https://github.com/OskarStark/php-cs-fixer-ga/issues/114) the author to create a new release.

prathmesh-stripe:
> All of the php file changes are from just format, right?
Yes

composer.json
prathmesh-stripe: Unpinned this to match with `latest` in CI. Diverging could break things.

.github/workflows/ci.yml
prathmesh-stripe: The other alternate is to build our own workflow instead of using this docker image.

jar-stripe (approved):
LGTM

PR #1832, opened 2025-03-27
Added requestCollection and requestSearchResult to StripeClientInterface

prathmesh-stripe:
### Why?
We are adding new helper methods to `StripeClientInterface` to improve the return types in SDKs. 
Stripe's implementation in [BaseStripeClient](https://github.com/stripe/stripe-php/blob/f65c497d0bc175aaa04538602fd49645f44f9384/lib/BaseStripeClient.php#L259-L315).

### See Also
https://jira.corp.stripe.com/browse/DEVSDK-2458#L80-L94

## Changelog
- ⚠️ Added `requestSearchResult`, `requestCollection` to `StripeClientInterface`.

### DIFF
--- a/lib/StripeClientInterface.php
+++ b/lib/StripeClientInterface.php
@@ -18,4 +18,28 @@ interface StripeClientInterface extends BaseStripeClientInterface
      * @return StripeObject the object returned by Stripe's API
      */
     public function request($method, $path, $params, $opts);
+
+    /**
+     * Sends a request to Stripe's API and expects to return a SearchResult.
+     *
+     * @param 'get' $method the HTTP method
+     * @param string $path the path of the request
+     * @param array $params the parameters of the request
+     * @param array|Util\RequestOptions $opts the special modifiers of the request
+     *
+     * @return SearchResult of ApiResources
+     */
+    public function requestSearchResult($method, $path, $params, $opts);
+
+    /**
+     * Sends a request to Stripe's API and expects to return a Collection|V2\Collection.
+     *
+     * @param 'get' $method the HTTP method
+     * @param string $path the path of the request
+     * @param array $params the parameters of the request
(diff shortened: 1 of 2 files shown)


lib/StripeClientInterface.php
```diff
@@ -18,4 +18,28 @@ interface StripeClientInterface extends BaseStripeClientInterface
...
+     * @param array|Util\RequestOptions $opts the special modifiers of the request
+     *
+     * @return SearchResult of ApiResources
+     */
+    public function requestSearchResult($method, $path, $params, $opts);
+
+    /**
+     * Sends a request to Stripe's API and expects to return a Collection|V2\Collection.
+     *
+     * @param 'delete'|'get'|'post' $method the HTTP method
```
jar-stripe, line 37: does 'delete' or 'post' ever return a collection?  can we get rid of `$method`

lib/StripeClientInterface.php
```diff
@@ -18,4 +18,28 @@ interface StripeClientInterface extends BaseStripeClientInterface
      * @return StripeObject the object returned by Stripe's API
      */
     public function request($method, $path, $params, $opts);
+
+    /**
+     * Sends a request to Stripe's API and expects to return a SearchResult.
+     *
+     * @param 'delete'|'get'|'post' $method the HTTP method
```
jar-stripe, line 25: same as below; do we support search across multiple http methods, and if not can we get rid of `$method`?

jar-stripe (commented):
LGTM!

jar-stripe (approved):

lib/StripeClientInterface.php
prathmesh-stripe: We currently do not, but we might in the future with V2 APIs. These methods have existed for a long time on BaseStripeClient. We risk breaking a lot of users that are already using requestCollection and requestSearchObject method on StripeClient by changing the signature.

PR #1836, opened 2025-03-31
Fail Code formatter in CI

prathmesh-stripe:
### Why?
CI PHP Code Formatter was always green and wasn't failing even for incorrect formatting. This PR fixes that and I ran the formatter on the unformatted files. 

### What?
Changed formatting command from `just format` to `just format-check`


jar-stripe (approved):
LGTM

PR #1838, opened 2025-03-31
Prathmesh/merge php beta

prathmesh-stripe:
### Why?
We GAed types. This is generated using the merge script.


prathmesh-stripe:
Don't need this

PR #1842, opened 2025-04-02
Update PaymentMethod.php: Adding `null|` to `$card`

ThomasLandauer:
### Why?
According to https://docs.stripe.com/api/payment_methods/object#payment_method_object-card, `$card` is a "**nullable** associative array"

### What?
Just added `null|` in front.

### See Also
Besides, I think your general approach of `\stdClass&StripeObject` is wrong: https://www.php.net/manual/en/class.stdclass.php says:
> This is not a base class as PHP does not have a concept of a universal base class. However, it is possible to create a custom class that extends from stdClass and as a result inherits the functionality of dynamic properties.

When I'm doing `if ($paymentMethod->card instanceof StripeObject)`, then PHPStan is reporting:

> Instanceof between *NEVER* and Stripe\StripeObject will always evaluate to false.

The `*NEVER*` is probably meant to indicate that `\stdClass&StripeObject` will never be true.
Why do you mention `\stdClass` there at all? Have you thought about changing `StripeObject`?:
```php
class StripeObject extends \stdClass


prathmesh-stripe:
Thank you for identifying an issue with non required objects. We are working on releasing a fix that will address this issue across all the classes. 

> Why do you mention \stdClass there at all? Have you thought about changing StripeObject?

`stdClass` allows developers to access properties on objects that have not been officially released and are still part of beta. I'll follow up on this PR to find a solution for your PHPStan error once we merge in the primary issue.

prathmesh-stripe:
We shipped a fix for this bug in https://github.com/stripe/stripe-php/releases/tag/v17.1.0

prathmesh-stripe:
We addressed the issue that you raised about `stdClass` in our latest release: https://github.com/stripe/stripe-php/releases/tag/v17.1.1

PR #1844, opened 2025-04-02
Fix event names and related objects, add null type to optional objects

ramya-stripe:
### Why?
Manually creating PRs as automation is currently failing
Re-generating after the latest set of fixes in the codegen around fixing event names and related objects as well as adding null type to optional fields

### What?
Re-run code gen


prathmesh-stripe:
No diff after merging https://github.com/stripe/stripe-php/pull/1845

PR #1846, opened 2025-04-04
Fix unresolvable stdClass&StripeObject types in stripe-php 17.x

ostrolucky:
### Why?
`stdClass&StripeObject` doesn't make sense, because they don't extend each other. Phpstan then evaluates this as `never` type.

Plase see phpstan playground link for demonstration.

### What?
Removes &\stdClass from types, since it doesn't make sense and doesn't play nice with SA analyzers

### See Also
https://phpstan.org/r/6b205ece-5db1-4f84-8a16-4da3cc0f99a6
https://github.com/spaze/phpstan-stripe/issues/30)


prathmesh-stripe:
Thanks for raising this issue. We have addressed this in our latest release:  https://github.com/stripe/stripe-php/releases/tag/v17.1.1

PR #1851, opened 2025-04-07
Handle external_account field

mbroshi-stripe:
### Why?

The `external_account` should be a union type, not a `string`.

### What?

- Changes `external_account` field in `externalAccounts.create` from a `string` to a union type.

## Changelog
- Changes `external_account` field in `externalAccounts.create` from a `string` to a union type.


xavdid-stripe (approved):
I think it's fine since it lacks further types, but I expected PHP to have more detail here. Nothing to do with your PR though

PR #1856, opened 2025-04-11
Faster parallel runner for PHP formatter

prathmesh-stripe:
### Why?
Currently PHP CS Fixer takes a long time to run(~50s). With a parallel runner(currently in beta), it only takes ~7s.

### What?
Added ability for PHP CS Fixer to run parallel on multiple cores. This feature is limited to our local machines due to [incompatibility](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/b62538df5cf84c9f1c990e67c04819fbf4c8f285/tests/Runner/RunnerTest.php#L254) of the parallel runner with linux machines.


xavdid-stripe (approved):
awesome!

Issue #1854, opened 2025-04-09
The doctype regarding metadata is wrong


tomasfejfar:
### Describe the bug

**Expected:** 
metadata key of $params is metadata?: array<string, string>

**Actual:**
metadata key of $params is metadata?: null|\Stripe\StripeObject

----

Array works and even [docs write it that way](https://docs.stripe.com/api/metadata?lang=php).

That's a problem for static analysis like PHPStan, where you need to ignore the error, potentially ignoring more you intended and missing an actual bug. 

I noticed it's the case for:
* `Stripe\Service\CustomerService::create()`
*  `Stripe\Service\InvoiceService::create()`

### To Reproduce

Write following code:
```
$stripe = new \Stripe\StripeClient('xxxxx');
$stripe->customers->create(['metadata' => ['order_id' => '6735']]);
```

Run PHPStan analysis

### Expected behavior

**Expected:** 
No errors.

**Actual:**
`💡 Offset 'metadata' (Stripe\StripeObject|null) does not accept type array<string, string>.`

### Code snippets

```PHP
$stripe = new \Stripe\StripeClient('xxxxx');
$stripe->customers->create(['metadata' => ['order_id' => '6735']]);
```

### OS

WSL2

### PHP version

8.4

### Library version

stripe/stripe-php v17.1.1

### API version

N/A

kierwils:
I'm getting the same issue, the request landing in the sandbox is also coming through wrong 🙈 



```
{
...
"metadata": "Stripe\StripeObject JSON: { "user_id": 123 }","name": "No Name",
...
}
```

![](https://i.imgur.com/bjYfOG1.png)

clementbirkle:
Same issue with the `update` method.

prathmesh-stripe:
Thanks for reporting this. We are working on a fix this.

okorneliuk:
Also, Psalm says, that this:
```php
\Stripe\Checkout\Session::create([
  // ...
  'line_items' => [
    [
        'quantity' => 1,
        // ...
    ],
  ]
])
```
is invalid, but this:
```php
\Stripe\Checkout\Session::create([
  // ...
  'line_items' => [
      'quantity' => 1,
      // ...
  ]
])
```
is valid

prathmesh-stripe:
[...]

PR #1869, opened 2025-05-09
Fixed: Change type of metadata and other map parameters to array<K, V>

prathmesh-stripe:
### Why?
Fixed type of `metadata` and other map parameters like `currency_options` in all API functions based on https://docs.stripe.com/api/metadata

### What?
- Changed the type of `metadata` from `StripeObject` to `array<string, string>`

### See Also
https://github.com/stripe/stripe-php/issues/1854

## Changelog
- Changed type of map parameters(eg. metadata, currency_options) from `StripeObject` to `array<KType, VType>` in all methods.


jar-stripe (approved):
LGTM

PR #1879, opened 2025-06-13
Xavdid/merge php beta

xavdid-stripe:


API_VERSION
```diff
@@ -0,0 +1 @@
+2025-05-28.basil
```
xavdid-stripe, line 1: @helenye-stripe is it weird that this doesn't match the generated lines below?

jar-stripe: ahh yea, I needed to add API_VERSION to the automerge configs.  once i did (in https://github.com/stripe/sdk-codegen/pull/2579), the merge conflict went away.  we can close this PR.

PR #1881, opened 2025-06-20
Ensure compatibility with POST on older versions of libcurl

jar-stripe:
### Why?
Related to https://github.com/stripe/stripe-php/issues/1817.  Several users reported a similar issue where running stripe-php released on or after 9/30 using PHP with curl on or around version 7.29 result in 400 errors.

#### Why did this work previously to the 9/30 release?
The libcurl spec states that:
> If CURLOPT_POSTFIELDS is explicitly set to NULL then libcurl gets the POST data from the read callback. To send a zero-length (empty) POST, set CURLOPT_POSTFIELDS to an empty string, or set CURLOPT_POST to 1 and CURLOPT_POSTFIELDSIZE to 0.

Prior to the acacia major release, CurlClient.php would set CURLOPT_POSTFIELDS when the params hash is empty.  This changed in acacia and when params are empty, CURLOPT_POSTFIELDS not set at all.  [Older versions of libcurl have a bug](https://curl.se/mail/lib-2014-01/0103.html) where not setting CURLOPT_POSTFIELDS will result curl sending a `Content-Length: -1` which is likely the cause of the 400 errors.

### What?
- add code to `constructCurlOptions` explicitly set \CURLOPT_POSTFIELDS to empty string if body is null
- updated tests

This also changes a test in CurlClientTest.php which appears to be flakey; it is testing that we raise an exception when we encounter a disconnect after calling `executeStreamingRequestWithRetries`, but as written the request completes successfully before the server can disconnect the client.

### See Also
https://go/j/RUN_DEVSDK-1629

### DIFF
--- a/lib/HttpClient/CurlClient.php
+++ b/lib/HttpClient/CurlClient.php
@@ -267,6 +267,11 @@ class CurlClient implements ClientInterface, StreamingClientInterface
         if ($body) {
             $opts[\CURLOPT_POSTFIELDS] = $body;
         }
+        // inspired by https://github.com/stripe/stripe-php/issues/1817#issuecomment-2670463182
+        else if (isset($opts[\CURLOPT_POST]) && $opts[\CURLOPT_POST] === 1) {
+            $opts[\CURLOPT_POSTFIELDS] = '';
+        }
+
         // this is a little verbose, but makes v1 vs v2 behavior really clear
         if (!$this->hasHeader($headers, 'Idempotency-Key')) {
             // all v2 requests should have an IK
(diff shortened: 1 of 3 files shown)


jar-stripe:
> LGTM but can you validate this before merging.

yep, working on it. thanks!

jar-stripe:
We've not yet been able to verify this specific fix, but given all of the evidence and the fact that the fix is based on the successful workaround in https://github.com/stripe/stripe-php/issues/1817, I'm going to merge this as in and we can revisit if there is still an issue.

Issue #1893, opened 2025-08-18
Add getter for `stripe_account`


andreaskienast:
### Is your feature request related to a problem? Please describe.

In an application that deals with different Stripe accounts I've configured multiple Stripe clients via Symfony's dependency injection, where the Stripe account is passed into the configuration.

To build a mapping between the application user's and their existing Stripe accounts, I need to access the Stripe account of the client, which is not possible in any way.

### Describe the solution you'd like

It should be fine to add a `getStripeAccount()` method to `BaseStripeClient`, which returns `$this->config['stripe_account']`.

I will provide a PR for this.

PR #1894, opened 2025-08-18
Add getter for Stripe Account on BaseStripeClient

andreaskienast:
### Why?
In an application that deals with different Stripe accounts, a mapping is required to keep track of the customer ids to one specific application user.

### What?
The method `BaseStripeClient::getStripeAccount()` is introduced that allows to obtain the configured `stripe_account` from the used client instance again.

Fixes: #1893

## Changelog
- Add `getStripeAccount` method on `BaseStripeClient` to retrieve [Stripe Account ID](https://docs.stripe.com/connect/authentication?lang=php) from a `StripeClient` instance.


andreaskienast:
Thank you very much for taking care of this PR.

PR #1903, opened 2025-09-13
⚠️ Add strongly typed EventNotifications

xavdid-stripe:
### Why?

As part of the Next-Gen Event Handling project, we're adding fully typed classes to represent all the ~~thin events~~ Event Notifications a user could receive. There are also helper methods that help with the event handling experience.

### What?

- (see changelog below for user-facing changes)
- add tests
- update examples

### See Also

[DEVSDK-2662](https://go/j/browse/DEVSDK-2662)

## Changelog

We've overhauled how V2 Events are handled in the SDK! This approach should provide a lot more information at authoring and compile time, leading to more robust integrations. As part of this process, there are a number of changes to be aware of.

- ⚠️ Renamed `\Stripe\ThinEvent` to `\Stripe\V2\EventNotification`
- add the `Stripe\V2\UnknownEventNotification` class
- ⚠️ Removed the `relatedObject` property from `Stripe\V2\EventNotification`. It's on `UnknownEventNotification` instead.
- ⚠️ Renamed `StripeClient->parseThinEvent` to `StripeClient->parseEventNotification` and changed its method signature accordingly
- ⚠️ removed the `Util::json_decode_thin_event_object`. Its functionality was folded into the new `\Stripe\V2\EventNotification::fromJson` method.
- Added matching `EventNotification` classes for every v2 `Event`. For example, there's now a `V1BillingMeterErrorReportTriggeredEventNotification` to match the existing `V1BillingMeterErrorReportTriggeredEvent`. Each of these interfaces defines a `fetchEvent()` method to retrieve its corresponding event. For events with related objects, there's a `fetchRelatedObject()` method that performs the API call and casts the response to the correct type.
- add the `EventNotification` Stripe.V2.EventNotification` interface, which is a union of all possible event notifications (when this SDK was generated)


tests/Stripe/EventTest.php
```diff
@@ -105,4 +110,146 @@ public function testV2EventFetchRelatedObject()
...
+        $event = EventNotification::fromJson($eventData, new StripeClient());
+        self::assertInstanceOf(V1BillingMeterNoMeterFoundEventNotification::class, $event);
+
+        // @var V1BillingMeterNoMeterFoundEventNotification $event
+        self::assertSame('evt_234', $event->id);
+        self::assertSame('v1.billing.meter.no_meter_found', $event->type);
+        self::assertSame('2022-02-15T00:27:45.330Z', $event->created);
+    }
+
+    public function testJsonDecodeThinEventObjectWithNoReasonObject()
```
prathmesh-stripe, line 239: Names of tests should be EventNotifications?

lib/V2/EventNotification.php
```diff
@@ -0,0 +1,122 @@
...
+        return $this->client->deserialize($response->body, 'v2');
+    }
+
+    protected function fetchRelatedObject()
+    {
+        if (null === $this->related_object) {
+            return null;
+        }
+
+        $response = $this->client->rawRequest(
```
prathmesh-stripe, line 111: Do we not want to cache this response?

lib/BaseStripeClient.php
```diff
@@ -280,7 +282,12 @@ public function rawRequest($method, $path, $params = null, $opts = [], $maxNetwo
         $opts->headers = \array_merge($opts->headers, $headers);
         $baseUrl = $opts->apiBase ?: $this->getApiBase();
         $requestor = new ApiRequestor($this->apiKeyForRequest($opts), $baseUrl);
-        list($response) = $requestor->request($method, $path, $params, $opts->headers, $apiMode, ['raw_request'], $maxNetworkRetries);
+
+        if (null === $usage) {
+            $usage = ['raw_request'];
+        }
```
prathmesh-stripe, line 288: Don't we want to append `raw_request` to usage if it is not null but calls stripe API through this function?

lib/BaseStripeClient.php
```diff
@@ -267,7 +269,7 @@ public function rawRequest($method, $path, $params = null, $opts = [], $maxNetwo
             $headers = $opts['headers'] ?: [];
             unset($opts['headers']);
         }
-        if (\is_array($opts) && \array_key_exists('stripe_context', $opts)) {
+        if (\is_array($opts) && \array_key_exists('stripe_context', $opts) && null !== $opts['stripe_context']) {
```
prathmesh-stripe, line 272: Why are we doing a null check? 
Consider this use case:
1. I create a stripe client with Stripe_context value.
2. I want to send a request to the main Stripe Account without stripe_context set. 

As a user I might think I can pass null or an empty string as one of the options to raw request and that should clear out the stripe_context for that request. Are we sure about not giving support for `null` values for these special headers?

tests/Stripe/EventTest.php
```diff
@@ -105,4 +110,146 @@ public function testV2EventFetchRelatedObject()
...
+
+        // @var V1BillingMeterErrorReportTriggeredEventNotification $event
+        self::assertSame('evt_234', $event->id);
+        self::assertSame('v1.billing.meter.error_report_triggered', $event->type);
+        self::assertSame('2022-02-15T00:27:45.330Z', $event->created);
+        self::assertSame('fa_123', $event->related_object->id);
+        self::assertSame('financial_account', $event->related_object->type);
+        self::assertSame('/v2/financial_accounts/fa_123', $event->related_object->url);
+        self::assertInstanceOf(Reason::class, $event->reason);
+        self::assertInstanceOf(ReasonRequest::class, $event->reason->request);
```
prathmesh-stripe, line 216: nit: Can we also assert the instance of $event->related_object here?

prathmesh-stripe (approved):

lib/V2/EventNotification.php
xavdid-stripe: No, the intention is that every call to `fetchRelatedObject` makes an API call. The object may have changed in the meantime and caching would lead to more confusion. The user can store it if they want.

lib/BaseStripeClient.php
xavdid-stripe: Yeah, that's a good point. Maybe i'll be a little more defensive in `fetchRelatedObject` to only set the context if there's one to set.

xavdid-stripe: I'm about to change a bunch of context stuff, so i'll make sure all our bases are covered!

PR #1909, opened 2025-09-24
add support for v2 deleted objects (#1889)

prathmesh-stripe:
### Why?

The typing of delete calls has always been a little confusing in V1: SDKs say they return an entire object (like a `Customer`) but with the `deleted` property set to `True`. That's true for some APIs but not for others. For consistency, we just always say you get a "real" object.

V2 APIs are more strict about the data they return: there's only ever `id` and `object`. To better represent this, we're now generating a `V2.DeletedObject` class that is returned from all v2 delete calls. This PR includes that generated class, but APIRequestor-related code to use it when appropreate.

> Note: this PR is generated using the V2 OpenAPI spec rather than the protos so that the `DeletedObject` class is generated, so there are some other small changes.

### What?

- add param to `Util::convertToStripeObject`
- provide that param from the API requestor

### See Also

[DEVSDK-2588](https://go/j/DEVSDK-2588)


ramya-stripe:
These changes are already in master

Issue #1972, opened 2025-11-25
Deprecated `curl_close` usage in `/lib/HttpClient/CurlClient.php`


CommanderRoot:
### Describe the bug

When running PHP 8.5 this deprecated message is logged / displayed.
```
Function curl_close() is deprecated since 8.5, as it has no effect since PHP 8.0 in /vendor/stripe/stripe-php/lib/HttpClient/CurlClient.php on line 761
```

The best solution would probably be surrounding the above line with a version check like this
```
if (PHP_VERSION_ID < 80000) {
    \curl_close($this->curlHandle);
}
```

### To Reproduce

Run a script which uses the Stripe sdk to contact the Stripe API.
See error on script finish.

### Expected behavior

No deprecated function is used

### Code snippets

```PHP

```

### OS

Debian 13

### PHP version

PHP 8.5

### Library version

stripe-php v19.0.0

### API version

2025-11-17.clover

ramya-stripe:
The fix is out in the latest version of the library - 19.1.0

PR #1973, opened 2025-11-25
Remove deprecated usage of curl_close() for PHP versions > 8.0

CommanderRoot:
### Why?
```
Function curl_close() is deprecated since 8.5, as it has no effect since PHP 8.0 in /vendor/stripe/stripe-php/lib/HttpClient/CurlClient.php on line 761
```

### What?
Only use `curl_close()` on PHP versions before 8.0 when it still did something

### See Also
#1972


ramya-stripe (approved):
Thank you!
This will get shipped in our next release in December

PR #2054, opened 2026-04-04
Add Ref type for typed v2 object references

jar-stripe:
### Why?

Custom objects need typed references to other Stripe objects. A reference has the wire shape `{type, id, url}` and in the SDK becomes `Ref` — a container that can fetch the referenced object.

### What?

- `\Stripe\V2\Ref` class with `@template T of \Stripe\StripeObject`
- Public `$type`, `$id`, `$url` properties
- Stores `$client` (a `BaseStripeClient`) for making requests
- `fetch()` calls `$this->client->rawRequest()` + `$this->client->deserialize()` — same pattern as `EventNotification.fetchRelatedObject()`
- Client threading: `$client` parameter added to `Util::convertToStripeObject()`, passed from `BaseStripeClient.request()` and `BaseStripeClient.deserialize()`. Stored on `StripeObject._client` in `constructFrom()`, threaded through `updateAttributes()` for nested field deserialization.
- Ref shape detection in `convertToStripeObject`: v2 arrays with `type`, `id`, `url` keys and no `object` key are instantiated as `Ref($resp, $client)`

### See Also

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/Util/Util.php
+++ b/lib/Util/Util.php
@@ -44,14 +44,14 @@ abstract class Util
      *
      * @return array|StripeObject
      */
-    public static function convertToStripeObject($resp, $opts, $apiMode = 'v1', $isV2DeletedObject = false)
+    public static function convertToStripeObject($resp, $opts, $apiMode = 'v1', $isV2DeletedObject = false, $client = null)
     {
         $types = 'v1' === $apiMode ? ObjectTypes::mapping
             : ObjectTypes::v2Mapping;
         if (self::isList($resp)) {
             $mapped = [];
             foreach ($resp as $i) {
-                $mapped[] = self::convertToStripeObject($i, $opts, $apiMode);
+                $mapped[] = self::convertToStripeObject($i, $opts, $apiMode, false, $client);
             }
 
             return $mapped;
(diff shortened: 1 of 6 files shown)


tests/Stripe/V2/RefTest.php
```diff
@@ -0,0 +1,173 @@
...
+        $json = [
+            'type' => 'v2.core.account',
+            'id' => 'acct_123',
+            'url' => '/v2/core/accounts/acct_123',
+        ];
+
+        $ref = new Ref($json);
+
+        $this->expectException(\Stripe\Exception\UnexpectedValueException::class);
+        $this->expectExceptionMessage('Ref has no client.');
```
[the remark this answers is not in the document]
jar-stripe: Good catch — the test asserts the wrong message string. Will fix to match the actual exception message.

lib/Util/Util.php
```diff
@@ -76,11 +76,19 @@ public static function convertToStripeObject($resp, $opts, $apiMode = 'v1', $isV
                 // TODO: this is a horrible hack. The API needs
                 // to return something for `object` here.
                 $class = \Stripe\V2\Collection::class;
+            } elseif (
+                'v2' === $apiMode
+                && !\array_key_exists('object', $resp)
+                && \array_key_exists('type', $resp)
+                && \array_key_exists('id', $resp)
+                && \array_key_exists('url', $resp)
```
mbroshi-stripe, line 84: Nit: Can you leave a comment here explaining why these particular conditionals imply it's a `Ref`? Otherwise, it looks a little mysterious and scary to update.

mbroshi-stripe (approved):

lib/Util/Util.php
mbroshi-stripe: I'm also wondering if there's some edge case where another V2 object has `type`, `id`, and `url` fields. Is it worthwhile tightening `!exists('object')` to `!exists(any other field)`?

jar-stripe:
Changing direction for now.

Issue #2091, opened 2026-07-03
PHPDoc on Collection::getIterator()/getReverseIterator()/autoPagingIterator() drops the TStripeObject generic, breaking type inference in PHPStan


darkdarin:
### Is your feature request related to a problem? Please describe.

`Stripe\Collection` is correctly declared as a generic, iterable type:

```php
/**
 * @template TStripeObject of StripeObject
 *
 * @template-implements \IteratorAggregate<TStripeObject>
 * ...
 */
class Collection extends StripeObject implements \Countable, \IteratorAggregate
```

However, the iterator-returning methods narrow the return type in PHPDoc **without** carrying the `TStripeObject` generic parameter:

```php
/**
 * @return \ArrayIterator an iterator ...
 */
public function getIterator() { ... }

/**
 * @return \ArrayIterator an iterator ...
 */
public function getReverseIterator() { ... }

/**
 * @return \Generator|TStripeObject[] A generator ...
 */
public function autoPagingIterator() { ... }
```

Because these PHPDoc annotations are more specific than the `@template-implements \IteratorAggregate<TStripeObject>` on the class, PHPStan uses them and resolves the iterated value type to `mixed`:

- `\ArrayIterator` without parameters is treated as `\ArrayIterator<mixed, mixed>`.
- `\Generator|TStripeObject[]` is a union that collapses to `\Generator<mixed, mixed>|array<int, TStripeObject>`; when iterated, the `\Generator` branch yields `mixed`, so the whole element type widens to `mixed`.

The net effect: iterating a collection directly (which goes through `getIterator()` via `IteratorAggregate`) loses the element type entirely.

Example (illustrative, not real code) — assuming a collection of `Stripe\Customer`:

```php
/** @var \Stripe\Collection<\Stripe\Customer> $customers */
$customers = $client->customers->all();

foreach ($customers as $customer) {
    // Expected: $customer is \Stripe\Customer
    // Actual (PHPStan): $customer is mixed
    echo $customer->email; // PHPStan: "Cannot access property $email on mixed"
}

foreach ($customers->autoPagingIterator() as $customer) {
    // Same problem: $customer is inferred as mixed
    echo $customer->id;
}
```

The generic on the class strongly implies these should stay typed as `TStripeObject`, but the method-level PHPDoc overrides that and drops it.

### Describe the solution you'd like

Add the generic value type to the iterator PHPDocs so they stay consistent with the class-level `@template-implements`:

```php
/**
[...]

PR #2093, opened 2026-07-07
Add TStripeObject to iterator PHPDoc comments

zacchua-stripe:
### Why?

Collection implements `IteratorAggregate<TStripeObject>`, but `getIterator()` and `getReverseIterator()` were annotated to return a bare `\ArrayIterator` with no type parameters. PHPStan uses the method's own declared return type when inferring foreach loop variable types. This caused loop variables to be inferred as `mixed` instead of the concrete type (e.g. `Customer`), producing errors like `"Cannot access property $email on mixed"` for users running PHPStan at the strictest level 9.

### What?

- `Collection::getIterator()` and `getReverseIterator()`: `@return \ArrayIterator` → `@return \ArrayIterator<int, TStripeObject>`
- Add `examples/IteratorExample.php`: a minimal reproduction case that iterates a `Collection<Customer>` via both `foreach` and `autoPagingIterator()`, serving as the PHPStan lint target
- Add `lint-test` recipe: runs PHPStan at level 9 against the example to verify the fix and prevent regressions
- Wire just lint-test into CI (phpstan job)

### See Also

- Issue: https://github.com/stripe/stripe-php/issues/2091

## Changelog

- Fixed: PHPStan no longer infers iterated Collection values as `mixed`; loop variables are now correctly typed as the collection's generic type parameter

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,7 +96,7 @@ jobs:
         run: composer install --prefer-dist --no-progress
 
       - name: Run phpstan
-        run: just lint && just lint-test
+        run: just lint-test lint
 
   tests:
     runs-on: "ubuntu-24.04"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,7 +96,7 @@ jobs:
         run: composer install --prefer-dist --no-progress
 
       - name: Run phpstan
-        run: just lint
+        run: just lint && just lint-test
 
   tests:
     runs-on: "ubuntu-24.04"
--- a/justfile
+++ b/justfile
@@ -36,8 +36,9 @@ format-check: (format "--dry-run")
 lint *args:
     php -d memory_limit=512M vendor/bin/phpstan analyse lib tests {{args}}
 
+# statically analyze code (strict)
 lint-test:
-    phpstan analyse examples/IteratorExample.php --level=9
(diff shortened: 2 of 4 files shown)


justfile
```diff
@@ -36,6 +36,10 @@ format-check: (format "--dry-run")
 lint *args:
     php -d memory_limit=512M vendor/bin/phpstan analyse lib tests {{args}}
 
+# statically analyze code (strict)
+lint-test:
+    vendor/bin/phpstan analyse examples/IteratorExample.php --level=9
```
xavdid: nit: you should be able to remove `vendor/bin/`

.github/workflows/ci.yml
```diff
@@ -96,7 +96,7 @@ jobs:
         run: composer install --prefer-dist --no-progress
 
       - name: Run phpstan
-        run: just lint
+        run: just lint && just lint-test
```
xavdid: `just` lets you specify multiple recipes and it will run them serially (and deduplicate any shared dependencies). So this can be `just lint-test lint`

(you'd normally be able to do `just lint lint-test` too, but the `*args` on `lint` mean that `just` interprets any further arguments as input to the recipe, not recipes in and of themselves. So if you put non-arg recipes first, everything works)

xavdid (approved):
nice! had a couple of small comments. once you fix, you can merge! 🎉

PR #2095, opened 2026-07-09
Remove unused Retry-After header support

xavdid:
### Why?

We confirmed in [RUN_API-26216](https://go/j/RUN_API-26216) that the Stripe API doesn't (and has not ever, as far as we can tell) sent `Retry-After` to external consumers.

This was one of the SDKs that included support for reading that header. Given that it's always a no-op, we can remove the dead code.

### What?

- remove the code that reads the `Retry-After` header and associated calculations
- remove related tests

### See Also

[RUN_DEVSDK-2569](https://go/j/RUN_DEVSDK-2569)


lib/HttpClient/CurlClient.php
```diff
@@ -738,15 +737,7 @@ private function sleepTime($numRetries, $rheaders)
...
-        $sleepSeconds = \max(Stripe::getInitialNetworkRetryDelay(), $sleepSeconds);
-
-        // And never sleep less than the time the API asks us to wait, assuming it's a reasonable ask.
-        $retryAfter = isset($rheaders['retry-after']) ? (float) ($rheaders['retry-after']) : 0.0;
-        if (\floor($retryAfter) === $retryAfter && $retryAfter <= Stripe::getMaxRetryAfter()) {
-            $sleepSeconds = \max($sleepSeconds, $retryAfter);
-        }
-
-        return $sleepSeconds;
+        return \max(Stripe::getInitialNetworkRetryDelay(), $sleepSeconds);
```
jar-stripe: i was a bit confused by this.  is this to correct the value if e.g. applying the jitter causes it to go below the base?  if so, do we also need to bound it at the top by getMaxNetworkRetryDelay?

jar-stripe (commented):
if my question/interpretation is correct, we probably need a change to that code before shipping; if I was not correct, lmk and i can approve!

lib/HttpClient/CurlClient.php
xavdid: > is this to correct the value if e.g. applying the jitter causes it to go below the base?

correct- it's unchanged from before (now just `return`ed instead of used in a calculation

> if so, do we also need to bound it at the top by getMaxNetworkRetryDelay?

yes, but it's already accounted for above- when `$sleepSeconds`, it's run through `min` with `getMaxNetworkRetryDelay`. The jitter might take it above that value, but not by much.

jar-stripe: Oh I see, the jitter never goes above 1 * $sleepSeconds.  A very small nit: I might move the \max call to that line so its more connected with applying the jitter and correcting for going below the lower bound.

jar-stripe (approved):

lib/HttpClient/CurlClient.php
xavdid: sure!

PR #2098, opened 2026-07-10
Replace source hash with Telemetry UUID

xavdid:
### Why?

Recently, we introduced the `source` hash in the user-agent header to help our support team track down exactly where incoming requests originated from.

They've reported back to us that this hasn't been as useful as we wanted, so we're going a different direction.

Instead, we're adopting an industry-standard method of persisting a small UUID on disk that we include with subsequent requests. It's easy for users to read this file and it contains no PII. It also honors our existing telemetry opt outs.

It's meant to be lightweight and failure tolerant.

### What?

- remove `source` key from user-agent header
- add `telemetry_id` to the user-agent header
- adjust tests accordingly

### See Also

- original request: [DEVSDK-3131](https://go/j/DEVSDK-3131)
- update request: [RUN_DEVSDK-2563](https://go/j/RUN_DEVSDK-2563)

### DIFF
--- a/lib/TelemetryId.php
+++ b/lib/TelemetryId.php
@@ -1,0 +1,95 @@
+<?php
+
+namespace Stripe;
+
+class TelemetryId
+{
+    /** @var null|string */
+    private static $cachedId;
+
+    /** @var bool */
+    private static $loaded = false;
+
+    /**
+     * @return null|string
+     */
+    public static function get()
+    {
+        if (self::$loaded) {
+            return self::$cachedId;
+        }
+        self::$loaded = true;
+
+        $configDir = self::getConfigDir();
+        if (null === $configDir) {
+            return null;
+        }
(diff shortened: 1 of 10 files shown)


lib/TelemetryId.php
```diff
@@ -0,0 +1,95 @@
...
+     */
+    public static function getConfigDir()
+    {
+        if ('\\' === \DIRECTORY_SEPARATOR) {
+            $appData = \getenv('APPDATA');
+            if (false === $appData || '' === $appData) {
+                return null;
+            }
+
+            return $appData . \DIRECTORY_SEPARATOR . 'Stripe';
```
jar-stripe: small nit, but this pattern is repeated enough that it might be nice to pull out a path_join function

jar-stripe (approved):

PR #2105, opened 2026-07-24
add/adjust event parsing helpers

xavdid:
### Why?

Historically, our webhook parsing logic has been closely tied to signature validation (because the expectation is that users verify webhooks' authenticity before use). But, as the webhook handling experience has grown in complexity over time, our helper methods haven't kept up.

In addition to the classic parse+verify & handle, we also support / encourage :

- using cloud providers (like [AWS EventBridge](https://docs.stripe.com/event-destinations/eventbridge) and [Azure Event Grid](https://docs.stripe.com/event-destinations/eventgrid)), which don't require validation at all
- validating & ACKing an event quickly, then queuing it up for later processing
- unit tests w/ signed test payloads

This gives us 4 use cases that each need a subset of helper methods:

| case                                  | `verify` | `parse` | `parseAndVerify` | `signTestPayload` |
| ------------------------------------- | -------- | ------- | ---------------- | ----------------- |
| process immediately                   |          |         | ✅               | ✅                |
| verify, queue, process later          | ✅       | ✅      |                  | ✅                |
| process immediately (cloud provider)  |          | ✅      |                  |                   |
| queue, process later (cloud provider) |          | ✅      |                  |                   |

To support these workflows, we need all of these methods publicly available in every SDK:

1. `constructEvent(payload, signature, ...)` (parse + verify)
2. `parseEventNotification(payload, signature, ...)` (parse + verify)
3. `constructEventWithoutVerification(payload)` (parse, supports both direct webhooks and cloud provider envelopes)
4. `parseEventNotificationWithoutVerification(payload)` (parse, supports both direct webhooks and cloud provider envelopes)
5. `verify(payload, signature, secret, ...)`
6. `signTestPayload(payload, secret)`

The exact naming & arguments vary by language, but each SDK should provide roughly the same capabilities.

> [!IMPORTANT]
> For reviewers: the exact error messages are a little inconsistent between SDKs right now. Everything throws in the right place, but we're still iterating on some of the naming. We're trying out "thin event notification" as a descriptor on some of these surfaces. We're planning to have this all cleaned up by the GA launch of interop events; ignore exact messages for now.

### What?

> [!NOTE]
> these notes generic across all the SDKs; see `Changelog` below for the SDK-specific list of changes

- add missing event handling methods to `StripeClient` & `Webhook` classes, where relevant.
- Add an internal-facing `maybe_extract_from_cloud_provider_envelope` for handling wrapped events without the user having to think about it
- centralize some event parsing logic; adjusting accepted types for some methods
- adjust some error messages
- add tests
- update docstrings

### See Also

- [stripe/stripe-java#2201](https://github.com/stripe/stripe-java/issues/2201)
- [DEVSDK-3085](https://go/j/DEVSDK-3085)
- [DEVSDK-3210](https://go/j/DEVSDK-3210)

## Changelog

- Added methods that return their respective `Event`/`EventNotification` class instances without verifying authenticity. Use them when you've previously verified an event (e.g. you verified, put the event in a queue, and are now processing). Supports events from [AWS EventBridge](https://docs.stripe.com/event-destinations/eventbridge) and [Azure Event Grid](https://docs.stripe.com/event-destinations/eventgrid) natively.
  - `Webhook::constructEventWithoutVerification($payload)`
  - `BaseStripeClient::constructEventWithoutVerification($payload)`
  - `BaseStripeClient::parseEventNotificationWithoutVerification($payload)`
- Added `WebhookSignature::generateSignatureHeader($payload, $secret, $timestamp = null)`, which computes a full `Stripe-Signature` header for the given payload. Useful for unit tests!


lib/Webhook.php
```diff
@@ -28,21 +30,102 @@ public static function constructEvent($payload, $sigHeader, $secret, $tolerance
...
+    {
+        $data = self::parsePayload($payload);
+
+        // Could add as many checks as we want here, but we'll start simple
+        if (\array_key_exists('detail', $data)) {
+            // AWS
+            // https://docs.stripe.com/event-destinations/eventbridge#event-structure
+            return $data['detail'];
+        }
+        if (\array_key_exists('specversion', $data) && \array_key_exists('data', $data)) {
```
jar-stripe, line 117: small nit: maybe we should throw if data isnt there, with an error indicating the we expect data to be there?

jar-stripe (approved):
LG!  one small nit to consider

lib/Webhook.php
xavdid: If data is missing, we'll fall down to the `Exception\UnexpectedValueException` below! Assuming there's not an `object` field set to one of the strings we expect

PR #2111, opened 2026-07-30
Add discriminated union serialization tests

jar-stripe:
### Why?

Validates that the discriminated union serialization design works correctly in stripe-php before codegen produces these patterns. Covers both request-side (array params with literal discriminator) and response-side (StripeObject construction).

### What?

- Adds `tests/Stripe/DiscriminatedUnionTest.php` with 10 tests covering:
  - Standalone union: array params structure with discriminator key (request side)
  - Inline union: discriminator at parent level (request side)
  - Form encoding: verifies discriminator survives `Util::encodeParameters`
  - Response deserialization via `Util::convertToStripeObject` for both patterns

### DIFF
--- a/tests/Stripe/DiscriminatedUnionTest.php
+++ b/tests/Stripe/DiscriminatedUnionTest.php
@@ -1,0 +1,308 @@
+<?php
+
+namespace Stripe;
+
+/**
+ * Tests for discriminated union serialization.
+ *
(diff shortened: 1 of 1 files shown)


jar-stripe:
Collapsed into #2113 (coercion PR now includes serialization tests)

Issue #2114, opened 2026-08-04
v2 timestamp properties are annotated int but the API returns RFC 3339 strings


billypoke:
# v2 timestamp properties are annotated `int` but the API returns RFC 3339 strings

## Summary

Every timestamp property on the V2 resources is annotated `@property int`, but the v2 API
returns RFC 3339 strings. Passing one of these values to a function expecting an `int`
throws at runtime, and static analysis cannot catch it because the annotation is the thing
that is wrong.

Several of the annotations contradict themselves in the same line — the type says `int`
and the description gives an RFC 3339 example. From `lib/V2/Core/Account.php:18` on
current `master`:

```php
 * @property int $created Time at which the object was created. Represented as a RFC 3339 date &amp; time UTC value in millisecond precision, for example: 2022-09-18T13:22:18.123Z.
```

## Versions

- `stripe/stripe-php` v20.3.1 (also present on `master` as of v21.1.1 — see file list below)
- PHP 8.5.9

## Reproduction

```php
$link = $stripe->v2->core->accountLinks->create([
    'account' => 'acct_...',
    'use_case' => [
        'type' => 'account_onboarding',
        'account_onboarding' => [
            'configurations' => ['recipient'],
            'refresh_url' => 'https://example.test/refresh',
            'return_url' => 'https://example.test/return',
        ],
    ],
]);

var_dump($link->expires_at);            // string(24) "2026-08-04T20:06:11.000Z"
date('Y-m-d H:i:s T', $link->expires_at);
// PHP Fatal error: date(): Argument #2 ($timestamp) must be of type ?int, string given
```

The annotation says `int`, so this code type-checks cleanly under PHPStan at max level and
then fails in production.

A direct API call shows the same shape on `v2.core.account`:

```
$ curl -G https://api.stripe.com/v2/core/accounts/acct_... \
    -u "$KEY:" -H "Stripe-Version: 2026-06-24.dahlia"

{
  "id": "acct_...",
  "object": "v2.core.account",
  "created": "2026-07-13T17:33:11.000Z",
  ...
}
```

Reproduces identically on `2026-06-24.dahlia` and `2026-07-29.preview`.

## Affected properties

All on `master`:

| File | Line | Property |
| --- | --- | --- |
| `lib/V2/Core/Account.php` | 18 | `created` |
| `lib/V2/Core/AccountLink.php` | 12, 13 | `created`, `expires_at` |
[...]

PR #2115, opened 2026-08-04
Fix v2 timestamp phpdoc type annotations

zacchua-stripe:
### Why?

PHPDoc was hardcoding 'v1' semantics when generating `@property` annotations, causing v2 DateTime fields to be typed as int (Unix timestamp) instead of string (RFC 3339).

### What?

- Changed Event created phpdoc type annotation to string

### See Also

http://go/j/RUN_DEVSDK-2671
https://github.com/stripe/stripe-php/issues/2114


xavdid (commented):
nice! 

A couple of housekeeping items and then I can ✅ this.

In general, even when we're making changes that are mostly or entirely in codegen, we want a PR to `master` that present that change to users so they can reason about it just like any other change. 

In your case, that means you should:

- run `just gen php` in your codegen branch targeting this branch
- as a result, this PR should have all of those generated changes, plus this manual one, and nothing else (we don't want any API changes sneaking through). That's the user-facing change: all the annotations are updated
- Update the PR title to better describe the holistic change (v2 timestamp annotations are fixed) and remove the jira ticket tag (isn't useful in our changelog)
- either remove the `## changelog` (the PR title should be sufficient) or edit to describe the user impact. What was broken before and was was fixed? is there runtime impact for this change (no)

zacchua-stripe:
Thanks for the comments @xavdid, fixed!

xavdid (approved):
Nice! your PR description has a rogue github link (probably the result of typing `@property` without backticks) but otherwise good.

Also by convention, we don't link to codegen from open source PRs (we go the other way) but it's not a big deal.

zacchua-stripe:
Updated the description, thanks!

PR #2117, opened 2026-08-06
Emit Claude Code plugin hint at module load time

johno-stripe:
## Why?

Claude Code supports a [plugin hint](https://code.claude.com/docs/en/plugin-hints) feature, where if we detect the `CLAUDECODE` or `CLAUDE_CODE_CHILD_SESSION` environment variables, emitting the following string will prompt the user to install our [official plugin](https://claude.com/plugins/stripe):

```
<claude-code-hint v="1" type="plugin" value="stripe@claude-plugins-official" />
```

Our plugin's source code can be found [here](https://github.com/stripe/ai/tree/main/providers/claude/plugin).

## What?

- Adds `lib/agent_check.php`, which checks for either env variable at file scope and writes the hint to stderr if found. Errors
are swallowed so hint emission can never crash the SDK. Named generically to provide a clear target for future agent-specific affordances.
- Registers `lib/agent_check.php` in the files autoload array in composer.json, so it fires on every Composer autoloader
bootstrap.
- Also adds it as an explicit require in `init.php` for non-Composer users.

Note that while this will fire on every PHP invocation that loads the SDK, Claude Code automatically deduplicates by plugin so the user won't be re-prompted:

> Claude Code deduplicates by plugin, so emitting on every invocation has no downside ([source](https://code.claude.com/docs/en/plugin-hints#:~:text=Claude%20Code%20deduplicates%20by%20plugin%2C%20so%20emitting%20on%20every%20invocation%20has%20no%20downside))

The logic lives in its own file rather than version_check.php (the only other file-scope code in the library) because the
concerns are unrelated. PHP has no class-level initializer equivalent to Node's module load, so a files-autoloaded standalone
file is the idiomatic hook for this pattern.

See Also

- https://code.claude.com/docs/en/plugin-hints
- https://github.com/stripe/stripe-node/pull/2805
- https://github.com/stripe/stripe-python/pull/1867

Changelog

- Emits new Claude Code plugin hint when `CLAUDECODE` or `CLAUDE_CODE_CHILD_SESSION` environment variables are detected.


lib/agent_plugin_hint.php
```diff
@@ -0,0 +1,9 @@
+<?php
```
xavdid, line 1: minor, but i'd name this something else. We already have agent checking code and it's not here. Maybe `claude_hint.php` or something?

xavdid (approved):

lib/agent_plugin_hint.php
johno-stripe: I want to leave this file as a clear home for similar suggestions when other agent harnesses get around to supporting them, but I'm with you on `_check.php` not being well-scoped.  Updated to `agent_plugin_hint.php`, keep it very clear.

Issue #2125, opened 2026-08-20
agent_plugin_hint.php writes to stderr on every load, corrupting PHPUnit process-isolated output


karl-bullock:
### Summary

`lib/agent_plugin_hint.php` is loaded through composer's `files` autoload, so it runs on every load of the SDK. When it detects a coding-agent environment it writes a line to stderr:

```php
if (false !== getenv('CLAUDECODE') || false !== getenv('CLAUDE_CODE_CHILD_SESSION')) {
    fwrite($stderr, '<claude-code-hint v="1" ... />' . "\n");
}
```

Under PHPUnit's process isolation (`processIsolation="true"`), that line lands where the child process's serialised result is expected, and **every isolated test errors**, with the hint as the exception message:

```
PHPUnit\Framework\Exception: <claude-code-hint v="1" type="plugin" value="stripe@claude-plugins-official" />
```

### Reproducing

1. A project with `processIsolation="true"` and `stripe/stripe-php` in the autoload tree.
2. Run the suite with `CLAUDECODE` set to anything.
3. Every test errors. Unsetting the variable makes them pass; setting it empty does not, since the check is `false !== getenv(...)`.

Observed with 21.2.0 and PHPUnit 12.5.

### Why this is worth changing regardless of the feature

Writing to stderr unconditionally at autoload time is not safe for a library. Test harnesses, CLI tools and anything that parses a child process's output can be broken by it, and the failure is very hard to trace: the message does not look like it came from a dependency, and nothing in the stack points at the SDK.

The environment the check keys on is also exactly the environment where the output is most likely to be parsed by something.

### Suggestions, in the order I would prefer them

1. Do not emit anything from an autoloaded file. If the hint is worth keeping, move it behind an explicit call the host opts into.
2. Failing that, gate it on an interactive TTY (`stream_isatty(STDERR)`), which excludes captured pipes and test children.
3. Failing that, make it opt-out with a documented environment variable, and document the interaction with process isolation.

Happy to send a PR for whichever of those you would take.

xavdid:
This is fixed in `v21.2.1`, which just released. Thanks for reporting!

In the future, we've also fixed the bug where unsetting `CLAUDECODE` doesn't prevent the message.

PR #2126, opened 2026-08-20
don't emit agent hint during tests, which crashes isolated tests

xavdid:
### Why?

PHP unit tests can run in isolated mode which, among other things, crashes if there's anything in stderr. Our new claude hint printed to stdout, so a claude running isolated tests would see them all fail.

Instead of autoloading the hint and running it, we're moving it to a regular util and printing it at-most-once on client creation or before an API call is made (since there are non-client users that don't otherwise have a good startup hook)

Also, we're not checking the `CLAUDE_CODE_CHILD_SESSION` env var anymore. If it's set, `CLAUDECODE` is also set. In order to make it easier to disable the hint, you only have to zero out one env var now.

### What?

- Move the claude hint to a new, non-autoloaded file
- call it in 2 places
- add tests

### See Also

- https://github.com/stripe/stripe-php/issues/2125
- [RUN_DEVSDK-2741](https://go/j/RUN_DEVSDK-2741)


xavdid:
A note about release: once approved, i'm going to run the release script _from this branch_ so we can release the fix without releasing the new event handler stuff before next week. So please **don't** "update branch"

lib/Util/AgentPluginHint.php
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * @return bool
+     */
+    private static function shouldEmit()
+    {
+        // Claude Code sets this to "1" in every subprocess it spawns; treat an empty value as
+        // unset, so `CLAUDECODE= <command>` suppresses the hint.
+        if ('' === (string) \getenv('CLAUDECODE')) {
```
zacchua-stripe, line 49: Do we also need to check for `CLAUDE_CODE_CHILD_SESSION`? It was in the deleted `lib/agent_plugin_hint.php` file below.

zacchua-stripe (approved):
Looks good! Just had one clarifying question

lib/Util/AgentPluginHint.php
xavdid: Ah I should have mentioned in the description - its removal was intentional. It's a subset of `CLAUDECODE` (i.e. if it's set, `CLAUDECODE` is too). To make it easier to disable in case we missed something, you just need to blank out one variable, not both.

PR #2131, opened 2026-08-21
Add `.preHandle` method to `EventNotificationHandler`

xavdid:
### Why?

In final testing, users have expressed interest in being able to centrally run code before any other handler executes. It can be used for centralized logging, event deduplication, and more. I don't want to go full middleware, but this felt like a reasonable way to dip our toe into the waters.

Because users are responsible for [deduplicating events](https://docs.stripe.com/webhooks#handle-duplicate-events), this pre-handle hook (if present) should return a `bool`, signifying whether handling should continue.

If the function returns `true` (or is missing entirely), the handler proceeds like normal (running at-most-one other callback). If it returns `false`, no further callback is run and the original `.handle()` call finishes cleanly.

### What?

- add the `.pre_handle` method
- & associated tests
- update some error messages with correct verbs & consistency
- refactored some internal error handling
- updated example

### See Also

- http://go/j/DEVSDK-3249

### DIFF
--- a/lib/AbstractEventNotificationHandler.php
+++ b/lib/AbstractEventNotificationHandler.php
@@ -111,7 +111,7 @@ abstract class AbstractEventNotificationHandler
     private function assertCanRegister()
     {
         if ($this->hasHandledEvents) {
-            throw new Exception\BadMethodCallException('Cannot register new event handlers after .handle() has been called. This is indicative of a bug.');
+            throw new Exception\BadMethodCallException('Cannot register new callbacks after an event has been handled. This is indicative of a bug.');
         }
     }
 
@@ -128,7 +128,7 @@ abstract class AbstractEventNotificationHandler
     {
         $this->assertCanRegister();
         if (isset($this->registeredHandlers[$eventType])) {
-            throw new Exception\InvalidArgumentException("Handler for event type \"{$eventType}\" is already registered");
+            throw new Exception\InvalidArgumentException("Callback for event type \"{$eventType}\" is already registered");
         }
 
         $this->registeredHandlers[$eventType] = $handler;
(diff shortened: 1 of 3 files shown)


lib/AbstractEventNotificationHandler.php
```diff
@@ -107,16 +126,37 @@ protected function createClientWithContext($context)
      */
     protected function register($eventType, $handler)
     {
-        if ($this->hasHandledEvents) {
-            throw new Exception\BadMethodCallException('Cannot register new event handlers after .handle() has been called. This is indicative of a bug.');
-        }
+        $this->assertCanRegister();
         if (isset($this->registeredHandlers[$eventType])) {
```
jar-stripe, line 130: does this check belong in assertCanRegister?  the comment on that method says "registered once", so it may be responsible for both checks.

jar-stripe (approved):
Approving to unblock, but please consider that comment about assertCanRegister

lib/AbstractEventNotificationHandler.php
xavdid: yeah that would make sense. `assertCanRegister` is mostly ensuring we haven't handled any events yet (so that ruby rename to `assert_hasnt_handled_yet` was good), but we could probably do it as `assertCanRegister($eventName)` and do both checks (make sure it's valid to register _any_ handlers and that that event is available). 

I'll fix it in any open PRs and make a ticket to circle back to merged ones. Since it's an internal organization thing (and doesn't change the user-facing behavior) IMO it's not worth opening new PRs before the release for at this point.

xavdid: ah wait, now I remember why I didn't do that. `preHandle` needs to check that no event has been handled, but doesn't have an eventName to check against. So they're slightly different functions.

I'll just do the same rename here- `assertHasntHandledYet`. The "once" meant "register everything once, up front, before handling". I've clarified

PR #2138, opened 2026-08-28
Harden API requestor code against malicious URLs

xavdid:
### Why?

If an attacker got access to a user's webhook secret (or the user wasn't validating incoming webhook signatures) _and_ a user was calling the built-in `event_notification.fetch_related_object()` method, a malicious webhook payload could cause a user integration to make an authenticated request to an attacker-controlled domain (leaking the secret key)

To solve, we're being more careful during url construction to ensure user input can't be used to send requests to non-`stripe.com` domains (on a per-request basis).

### What?

- add path validation method and call it before the final URL is built
- Encode the `event.id` path param in `fetch_event()` to match the sanitation in the rest of the generated endpoints
- fix a bug where malicious newlines in headers could lead to injection attacks
- add tests

### See Also

- [RUN_DEVSDK-2808](https://go/j/RUN_DEVSDK-2808)

### DIFF
--- a/lib/ApiRequestor.php
+++ b/lib/ApiRequestor.php
@@ -480,14 +480,14 @@ class ApiRequestor
      *
      * The absolute URL is built by concatenating the API base onto this path,
      * and no base URL ends in a slash. A path like "@evil.example/v1/x" or
-     * ".evil.example/v1/x" would therefore land inside the authority component
-     * and send the request — Authorization header included — to a host of the
-     * path's choosing. Some request paths originate in remote data (a webhook
-     * body's related_object.url, a response's next_page_url), so this cannot be
-     * assumed to be well-formed.
+     * ".evil.example/v1/x" would modify the resulting host and direct the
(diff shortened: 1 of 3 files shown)


lib/ApiRequestor.php
```diff
@@ -474,6 +474,71 @@ private static function _defaultHeaders($apiKey, $clientInfo = null, $appInfo =
...
     }
 
+    /**
+     * Asserts that a request path is origin-relative: that it begins with a
+     * single "/" and carries no scheme, authority or userinfo.
+     *
+     * The absolute URL is built by concatenating the API base onto this path,
+     * and no base URL ends in a slash. A path like "@evil.example/v1/x" or
+     * ".evil.example/v1/x" would modify the resulting host and direct the
+     * request (including the API key) to a non-Stripe host.
```
jar-stripe, line 484: this feels like a lot of exposition for validatePath.  is there a better place to put this?

jar-stripe (approved):

PR #2142, opened 2026-09-08
Validate that webhook secrets are non-empty

xavdid:
### Why?

Fixes a potential security issue where, if the user passes an empty string into the event validation functions, the SDK would still generate a signature. If an attacker knew that a victim was inadvertently using an empty webhook secret, the attacker could send signed webhooks (using the empty secret) that a victim's integration would treat as valid.

The solution is to throw an error if the webhook secret is empty. This ensures all payloads are validated against an actual secret.

We also updated our examples to have better defaults, helping users to avoid this issue.

### What?

- verify that a webhook secret is non-empty & non-null when verifying a webhook
- update examples
- add tests

### See Also

- [RUN_DEVSDK-2953](https://go/j/RUN_DEVSDK-2953)

### DIFF
--- a/examples/EventNotificationHandlerEndpoint.php
+++ b/examples/EventNotificationHandlerEndpoint.php
@@ -15,6 +15,12 @@ require 'vendor/autoload.php';
 $api_key = getenv('STRIPE_API_KEY');
 $webhook_secret = getenv('WEBHOOK_SECRET');
 
+if (empty($webhook_secret)) {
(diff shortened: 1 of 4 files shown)


zacchua-stripe (approved):

nocolandiaflclxte7-ship-it (commented):
nocolandia.flcl.xte.7@gmail.com

PR #2148, opened 2026-09-15
Expand PHPDoc union parameter shapes

jar-stripe:
### Why?

Request-parameter PHPDoc currently reduces object- and list-shaped union variants to a generic `array`, which loses useful static-analysis and autocomplete information.

### What?

- Preserve complete object shapes and typed lists in generated request-parameter unions.
- Add a major-version changelog entry because stricter annotations may surface new PHPStan or IDE diagnostics.
- Keep runtime behavior and the minimum supported PHP version unchanged.

### Verification

- `just lint`
- `just test` (1044 tests, 2340 assertions)
- Regenerated twice from the same generator revision and confirmed identical output.

### See Also

This change is intended for the next major release and must not ship in a minor release.

## Configuration

- [ ] `skip-changefile`: This PR is not a user-facing change, so there's no changefile.


.hark/changes/2026-09-15_unknown_expand-phpdoc-union-shapes.change.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Expand PHPDoc union parameter shapes
+pr_url: https://github.com/stripe/stripe-php/pull/2148
+semver_level: major
+---
+
+- ⚠️ Request-parameter PHPDoc now describes the complete object and list shapes in unions instead of using a generic `array` type. This change affects documentation and static analysis only, but may produce new PHPStan or IDE diagnostics for arrays that do not match the documented types.
```
xavdid: you don't need the `⚠️` here- that'll get added to the top-level bullet (because it's marked as major). As written, this will produce:

```md
- ⚠️ [#2148](https://github.com/stripe/stripe-php/pull/2148) Expand PHPDoc union parameter shapes
  - ⚠️ Request-parameter PHPDoc now describes the complete object and list shapes in unions instead of using a generic `array` type. This change affects documentation and static analysis only, but may produce new PHPStan or IDE diagnostics for arrays that do not match the documented types.
```

so I'd drop the manual warning

.hark/changes/2026-09-15_unknown_expand-phpdoc-union-shapes.change.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Expand PHPDoc union parameter shapes
+pr_url: https://github.com/stripe/stripe-php/pull/2148
+semver_level: major
```
xavdid, line 4: This is marked breaking, but is merging into `master`. should this be going into the `next-major` branch (or whatever we're using)?

.hark/migration-guides/expand-phpdoc-union-shapes.md
```diff
@@ -0,0 +1,19 @@
+# Expanded PHPDoc union parameter shapes
```
xavdid: This isn't clear from CI errors (since we decided the migration guide stuff should be low-touch) but the intention is to have a single `.hark/migration-guides/<VERSION>.md` file that we build up as we need. So i'd rename this file `22.0.0.md`, change this header to an h2, and add a `Migration guide for v22` h1 at the top.

Eventually, this becomes (at least the draft of what we dump into) the wiki migration guide.

xavdid (approved):
approving to unblock, but double check the merge target please!

.hark/migration-guides/expand-phpdoc-union-shapes.md
jar-stripe: We should encode this guidance somewhere obvious to both humans and agents who will probably scaffold out these files.

.hark/changes/2026-09-15_unknown_expand-phpdoc-union-shapes.change.md
jar-stripe: good catch!  I changed the base to sdk-release/next-major

.hark/migration-guides/expand-phpdoc-union-shapes.md
xavdid: yep! i'll start with a better error message in CI and then maybe pre-seed the next few release files so it's not a blank canvas

.hark/changes/2026-09-15_unknown_expand-phpdoc-union-shapes.change.md
jar-stripe: Fixed in [dd5509f](https://github.com/stripe/stripe-php/pull/2148/commits/dd5509fbeee73099e823a768ab5c2b3b95b6b29b)

.hark/migration-guides/expand-phpdoc-union-shapes.md
jar-stripe: Fixed in [5bee8f0](https://github.com/stripe/stripe-php/pull/2148/commits/5bee8f0b54e8366f02a9220a279bc627b621af26)

